PR #3, opened 2018-04-16
AEIM-1076 - Fix up service require/notify

botimer:
The base config file was not required by the service, so the first
install on a server was breaking. Also, the filebeat and metricbeat
files were not notifying the services, so changes wouldn't cause a
restart.


botimer:
Fixed up in a marathon pairing session. Thanks, @daaang!

PR #6, opened 2018-06-15
AEIM-1244: Add functions for getting facts for other nodes

daaang:
They mock puppetdb so that you don't have to. You can just mock these
functions instead. I tested that puppetdb_query works the way I'm
assuming it does by checking by hand in production, adding lines like
this one to site.pp:

notify { to_json(puppetdb_query([etc])): }

Then I ran puppet agent on one of the other nodes to see if it returned
the output I expected. I am confident that these functions can be
trusted to work right when in production.

### DIFF
--- a/lib/puppet/functions/fact_for.rb
+++ b/lib/puppet/functions/fact_for.rb
@@ -10,19 +10,24 @@ Puppet::Functions.create_function(:fact_for) do
 
   def fact_for(node_id, fact_name)
     fact_keys = fact_name.split('.')
-    fact_base = fact_keys.shift
 
-    value = call_function('puppetdb_query',
-                          ['from', 'facts',
-                           ['extract', ['value'],
-                            ['and',
-                             ['=', 'certname', node_id],
-                             ['=', 'name', fact_base]]]])[0]['value']
+    value = run_query(node_id, fact_keys.shift)
 
-    fact_keys.each do |key|
-      value = value[key]
+    if fact_keys.empty?
+      value
+    else
+      value.dig(*fact_keys)
     end
+  end
 
-    value
+  private
+
+  def run_query(node_id, fact_name)
+    call_function('puppetdb_query',
+                  ['from', 'facts',
+                   ['extract', ['value'],
+                    ['and',
+                     ['=', 'certname', node_id],
+                     ['=', 'name', fact_name]]]])[0]['value']
   end
 end
--- a/lib/puppet/functions/fact_for.rb
+++ b/lib/puppet/functions/fact_for.rb
@@ -1,0 +1,28 @@
+# Copyright (c) 2018 The Regents of the University of Michigan.
+# All Rights Reserved. Licensed according to the terms of the Revised
+# BSD License. See LICENSE.txt for details.
+
+Puppet::Functions.create_function(:fact_for) do
+  dispatch :fact_for do
+    required_param 'String', :node_id
+    required_param 'String', :fact_name
+  end
+
+  def fact_for(node_id, fact_name)
+    fact_keys = fact_name.split('.')
+    fact_base = fact_keys.shift
+
+    value = call_function('puppetdb_query',
+                          ['from', 'facts',
+                           ['extract', ['value'],
+                            ['and',
+                             ['=', 'certname', node_id],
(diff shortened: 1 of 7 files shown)


lib/puppet/functions/fact_for.rb
```diff
@@ -0,0 +1,28 @@
...
+# BSD License. See LICENSE.txt for details.
+
+Puppet::Functions.create_function(:fact_for) do
+  dispatch :fact_for do
+    required_param 'String', :node_id
+    required_param 'String', :fact_name
+  end
+
+  def fact_for(node_id, fact_name)
+    fact_keys = fact_name.split('.')
```
malakai97: Lines 12-13 look like one responsibility, 15-20 another, and 22-24 a third. 

22-24 also confuses me: the first item in fact_keys is going to result in overwriting value.  Is there a way to use a non-destructive enumerable operation like `inject` to accomplish this that might be more clear?  (It could also be made clear by having its own method with an appropriate name).

lib/puppet/functions/nodes_for_role.rb
```diff
@@ -0,0 +1,19 @@
...
+# BSD License. See LICENSE.txt for details.
+
+Puppet::Functions.create_function(:nodes_for_role) do
+  dispatch :nodes_for_role do
+    required_param 'String', :role
+    return_type 'Array[String]'
+  end
+
+  def nodes_for_role(role)
+    db_role = role.split('::').map { |x| x.capitalize }.join('::')
```
malakai97: `def capitalize_role(role)`

lib/puppet/functions/nodes_for_role.rb
```diff
@@ -0,0 +1,19 @@
...
+
+    call_function('puppetdb_query',
+                  ['from', 'resources',
+                   ['extract', ['certname'],
+                    ['=', 'title', db_role]]]).map { |x| x['certname'] }
```
malakai97: Could be useful for understanding the result here by naming the `x` parameter to what we're actually iterating over.  I suspect it's a node?  Having that name would give me an idea what to go look for in puppet docs.

spec/functions/fact_for_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+
+describe 'fact_for' do
+  let(:node_id) { '' }
+  let(:fact_name) { '' }
+  let(:fact_value) { '' }
+
+  let!(:puppetdb_query) do
+    MockFunction.new('puppetdb_query') do |f|
+      f.stubbed
+       .with(['from', 'facts', ['extract', ['value'], ['and', ['=', 'certname', node_id], ['=', 'name', fact_name]]]])
```
malakai97: This seems fragile. I also suspect that there's not really anything you can do about it.

lib/puppet/functions/fact_for.rb
daaang: There's a lot of ruby features I don't know about, so it wouldn't surprise me to learn that ruby had a shortcut for this. The `fact_keys` array will be empty if there aren't any dots in the fact name, so the loop at the end won't execute at all in that case.

What I'm trying to do is use the first element as the parameter I hand to puppetdb (which doesn't support facter's dot syntax), and then use the rest to select keys; e.g. `one.two.three` becomes `value_for('one')['two']['three']`. Is there a ruby-ism for doing that?

malakai97:
```
[:a,:b:c].inject(start){|memo, key| memo.fetch(key) }

# results in:
start.fetch(:a)
start.fetch(:a).fetch(:b)
start.fetch(:a).fetch(:b).fetch(:c)
```
And returns the last one.

daaang: I ended up using dig instead; looks a bit cleaner to me. Apparently it does pretty much the exact thing I was trying to do.

spec/functions/fact_for_spec.rb
daaang: Yeah these are the things I tested in production. We don't have to structure the queries exactly like this, but if we do, the queries will work. I can't think of a better way, and I do think this is still better than no tests.

malakai97:
Feel free to just merge this, or cleanup the history first then merge.

PR #15, opened 2018-06-20
keepalived profile

aelkiss:
This adds a profile for keepalived.

Currently, the floating_ip parameter is intentionally duplicated between haproxy and keepalived to eliminate dependencies. When we complete AEIM-1248 we will likely rethink how this data is represented in hiera.

This moves the machinery for mocking puppet functions to spec_helper for now; eventually it will likely be extracted to a separate gem.

The actual mocked functions are moved to a shared example group, since the data used is very similar between the haproxy and keepalived profiles.

Comments welcome on the helper functions and their interface (as demonstrated by with_mocked_nodes.rb), especially if the comments come with a concrete working example.


spec/spec_helper.rb
```diff
@@ -2,6 +2,9 @@
 
 RSpec.configure do |c|
   c.mock_with :rspec
+  c.before(:each) do
```
botimer: I'm not sure that we want to do this out here. I don't think it will necessarily cause any trouble -- just a question of doing something so global.

manifests/profile/keepalived.pp
```diff
@@ -0,0 +1,42 @@
+# Copyright (c) 2018 The Regents of the University of Michigan.
+# All Rights Reserved. Licensed according to the terms of the Revised
+# BSD License. See LICENSE.txt for details.
+
+# keepalived
+#
+# @example
+#   include nebula::profile::keepalived
+class nebula::profile::keepalived(String $floating_ip) {
+#  require nebula::profile::haproxy
```
daaang: kill this comment

manifests/profile/keepalived.pp
```diff
@@ -0,0 +1,42 @@
...
+  package { 'keepalived': }
+  package { 'ipset': }
+
+  service { 'keepalived':
+    ensure => 'running',
+    enable => true,
+    hasrestart => true,
+#    require => Package['keepalived','haproxyctl'],
```
daaang: kill this comment

spec/classes/profile/keepalived_spec.rb
```diff
@@ -0,0 +1,96 @@
...
+        let(:file) { '/etc/sysctl.d/keepalived.conf' }
+
+        it { is_expected.to contain_file(file).with(ensure: 'present') }
+        it { is_expected.to contain_file(file).with(mode: '0644') }
+
+        it 'says it is managed by puppet' do
+          is_expected.to contain_file(file).with_content(
+            %r{\A# Managed by puppet \(nebula\/profile\/keepalived\/sysctl\.conf\.erb\)\n},
+          )
+        end
```
daaang: I haven't been testing this, but I'm open to the idea. My gut reaction is that this test is pretty brittle, though. I guess I'd like it better if it didn't specifically test the template path.

spec/spec_helper.rb
```diff
@@ -29,3 +32,34 @@
...
+def stub(name)
+  dbl = double(name)
+
+  yield dbl
+  stub_function(name, dbl)
+end
+
+def allow_call(dbl)
+  allow(dbl).to receive(:call)
+end
```
botimer: With the invocations wrapped up into the shared context, I'm not sure that it's worth indirecting this. If we added it as an object method, I'd like it more. For example: 
```ruby
def stub_with_call(name)
  double(name).tap do |dbl|
    dbl.define_singleton_method(:allow_call) do
      allow(dbl).to receive(:call)
    end
    stub_function(name, dbl)
  end
end
```

I think that form is a little tidier because of the return rather than the block. Then downstream can do whatever with it as an imbued double. But it's pretty even either way.

spec/classes/profile/keepalived_spec.rb
```diff
@@ -0,0 +1,96 @@
...
+
+        it { is_expected.to contain_file(file).with_content(%r{interface #{facts[:networking][:primary]}}) }
+
+        it { is_expected.to contain_file(file).with_content(%r{notification_email {\n\s.*root@default.invalid\n\s.*}}m) }
+        it { is_expected.to contain_file(file).with_content(%r{notification_email_from root@default.invalid}) }
+      end
+
+      describe 'service' do
+        it { is_expected.to contain_service(service).that_requires('Package[keepalived]') }
+        xit { is_expected.to contain_service(service).that_requires('Package[haproxyctl]') }
```
daaang: Why the skipped test?

manifests/profile/keepalived.pp
botimer: Extraneous?

manifests/profile/keepalived.pp
```diff
@@ -0,0 +1,42 @@
...
+# keepalived
+#
+# @example
+#   include nebula::profile::keepalived
+class nebula::profile::keepalived(String $floating_ip) {
+#  require nebula::profile::haproxy
+  package { 'keepalived': }
+  package { 'ipset': }
+
+  service { 'keepalived':
```
daaang: The puppet linter complains when these `=>`s aren't aligned ¯\\_(ツ)_/¯

The main puppet vim syntax settings do this automatically, for what it's worth.

manifests/profile/keepalived.pp
```diff
@@ -0,0 +1,42 @@
...
+#    require => Package['keepalived','haproxyctl'],
+    require => Package['keepalived'],
+  }
+
+  $nodes_for_role = nodes_for_role($title)
+  $nodes_for_datacenter = nodes_for_datacenter($::datacenter)
+  $email = lookup('nebula::root_email')
+  $datacenter = $::datacenter
+
+  file { '/etc/keepalived/keepalived.conf':
```
daaang: More `=>`s to align.

spec/spec_helper.rb
aelkiss: I meant to have this only in the shared context - thanks

manifests/profile/keepalived.pp
aelkiss: Gone.

manifests/profile/keepalived.pp
aelkiss: Gone

manifests/profile/keepalived.pp
aelkiss: Yes.

spec/spec_helper.rb
aelkiss: With this example, `allow` isn't within the right scope, and it fails. I had tried a variation on this (defining the method on the class) but it isn't obvious how to get the correct thing in there to call `allow` on.

spec/classes/profile/keepalived_spec.rb
aelkiss: I removed the template path.

spec/classes/profile/keepalived_spec.rb
aelkiss: Relates to the commented out dependency in keepalived.pp earlier - we had wanted to remove the dependency on the haproxy profile but I don't remember what the specific issue was. I'll remove this test for now. We may or may not want to put this dependency back in some form later.

aelkiss: The worst case if keepalived gets installed without haproxyctl is that all nodes will fail their check, which is probably fine, b/c if haproxyctl isn't installed that probably means haproxy isn't either...

aelkiss: This kind of reflects that the profile is misnamed - it isn't really keepalived, it's keepalived specifically for haproxy.

PR #18, opened 2018-06-29
provision user for monitoring haproxy

aelkiss:
Things I'd like feedback on:

- Whether the method of passing the user to the haproxy profile looks reasonable
- The nebula::authzd_user define in general - the general existence of it, but also in particular if there's a way of passing excess parameters to the user resource, that'd be nice (i.e. basically have nebula::authzd_user act as a decorator that ensures the home directory exists and can provision an ssh key in addition to actually creating the user)

- If there's a cleaner way of passing the parameters from `nebula::profile::haproxy` to `nebula::authzd_user` (a la https://github.com/mlibrary/nebula/blob/master/manifests/profile/vmhost/host.pp#L30 ) (I might experiment a little bit here if I have time)

- An easier-to-read way than https://github.com/mlibrary/nebula/compare/aeim-1296?expand=1#diff-889480bf39ad9af01eb92e9ac1a43da0L3 for conditionally putting the command in

### DIFF
--- a/data/common.yaml
+++ b/data/common.yaml
@@ -24,6 +24,13 @@ nebula::profile::elastic::filebeat::prospectors::mgetit::log_path: /var/log/mget
 nebula::profile::elastic::period: 90
 nebula::profile::haproxy::keepalived::floating_ip: '12.23.32.22'
 nebula::profile::haproxy::cert_source: ''
+nebula::profile::haproxy::monitoring_user:
+  name: haproxyctl
+  home: /var/haproxyctl
+  key:
+    type: ecdsa-sha2-nistp256
+    data: CCCCCCCCCCCC
+    comment: 'haproxyctl@default.invalid'
 nebula::profile::ruby::install_dir: '/opt/rbenv'
 nebula::profile::ruby::plugins:
 - rbenv/rbenv-vars
--- a/manifests/authzd_user.pp
+++ b/manifests/authzd_user.pp
@@ -11,20 +11,13 @@ define nebula::authzd_user(
   Hash $key,
   String $gid) {
 
-  $key_file = "${home}/.ssh/authorized_keys"
-
-  file { $home:
-    ensure => 'directory',
-    mode   => '0755'
-  }
-
   user { $title:
-    gid     => $gid,
-    home    => $home,
-    require => File[$home]
+    gid        => $gid,
+    home       => $home,
+    managehome => true
   }
 
-  nebula::file::ssh_keys { $key_file:
+  nebula::file::ssh_keys { "${home}/.ssh/authorized_keys":
     keys   => [ $key],
     secret => true,
   }
--- a/manifests/authzd_user.pp
+++ b/manifests/authzd_user.pp
@@ -1,0 +1,31 @@
+# Copyright (c) 2018 The Regents of the University of Michigan.
+# All Rights Reserved. Licensed according to the terms of the Revised
+# BSD License. See LICENSE.txt for details.
+
+# keepalived
+#
+# @example
+#   include nebula::profile::authzd_user()
+define nebula::authzd_user(
+  String $home,
+  Hash $key,
+  String $gid) {
+
+  $key_file = "${home}/.ssh/authorized_keys"
+
+  file { $home:
+    ensure => 'directory',
+    mode   => '0755'
+  }
+
+  user { $title:
+    gid     => $gid,
+    home    => $home,
+    require => File[$home]
+  }
+
+  nebula::file::ssh_keys { $key_file:
+    keys   => [ $key],
+    secret => true,
+  }
+}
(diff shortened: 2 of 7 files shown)


manifests/profile/haproxy.pp
```diff
@@ -52,4 +52,11 @@
     notify  => Service['haproxy'],
   }
 
+  nebula::authzd_user { $monitoring_user['name']:
+    gid  => 'haproxy',
+    home => $monitoring_user['home'],
+    key  => merge($monitoring_user['key'],
+{ command => '/usr/sbin/haproxyctl' } )
```
daaang: Hm yeah I see what you mean. It might be clearer if it's on one line, and maybe using `+` instead of `merge`, so it can't be confused with the parameters?

```
key => $monitoring_user['key'] + {'command' => '/usr/sbin/haproxyctl'},
```

manifests/authzd_user.pp
```diff
@@ -0,0 +1,31 @@
...
+    mode   => '0755'
+  }
+
+  user { $title:
+    gid     => $gid,
+    home    => $home,
+    require => File[$home]
+  }
+
+  nebula::file::ssh_keys { $key_file:
```
daaang: Since this is the only place `$key_file` is used, I feel like it'd be fine to skip it and just make this

```
nebula::file::ssh_keys { "${home}/.ssh/authorized_keys":
```

spec/defines/authzd_user_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+    }
+  end
+
+  on_supported_os.each do |os, os_facts|
+    context "on #{os}" do
+      let(:facts) { os_facts }
+
+      describe 'users' do
+        it { is_expected.to contain_user(title).with(name: title, gid: params[:gid], home: home).that_requires("File[#{home}]") }
+        it { is_expected.to contain_file(home).with(ensure: 'directory', mode: '0755') }
```
daaang: I'm curious about whether this test will continue passing if we use `managehome` instead of an explicit `file` resource. My bet is that it will, and you won't have to change the tests for that.

manifests/authzd_user.pp
```diff
@@ -0,0 +1,31 @@
...
+define nebula::authzd_user(
+  String $home,
+  Hash $key,
+  String $gid) {
+
+  $key_file = "${home}/.ssh/authorized_keys"
+
+  file { $home:
+    ensure => 'directory',
+    mode   => '0755'
```
daaang: I must have accidentally deleted this comment somehow. This resource could be omitted by adding `managehome => true` to the user resource. It'd accomplish the same thing, but it'd be more idiomatic puppet.

manifests/authzd_user.pp
```diff
@@ -0,0 +1,31 @@
...
+  }
+
+  user { $title:
+    gid     => $gid,
+    home    => $home,
```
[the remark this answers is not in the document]
aelkiss: OK, leaving this one as is for now until we need something different.

templates/file/ssh_keys.erb
```diff
@@ -1,4 +1,4 @@
 # Managed by puppet (nebula/file/ssh_keys)
 <% @keys.each do |key| -%>
-<%= key['type'] %> <%= key['data'] %> <%= key['comment'] %>
+<%= key.has_key?('command') ? "command=\"#{key['command']}\" " : '' %><%= key['type'] %> <%= key['data'] %> <%= key['comment'] %>
 <% end -%>
```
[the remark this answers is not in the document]
aelkiss: yeah, it's a little easier. Don't suppose there's such a thing as presenters or view helpers in puppet...?

data/common.yaml
```diff
@@ -23,6 +23,13 @@ nebula::profile::elastic::logstash_hosts:
 nebula::profile::elastic::filebeat::prospectors::mgetit::log_path: /var/log/mgetit.default.invalid
 nebula::profile::elastic::period: 90
 nebula::profile::haproxy::keepalived::floating_ip: '12.23.32.22'
+nebula::profile::haproxy::monitoring_user:
+  name: haproxyctl
+  home: /var/haproxyctl
+  key:
+    type: ecdsa-sha2-nistp256
+    data: CCCCCCCCCCCC
+    comment: 'haproxyctl@default.invalid'
```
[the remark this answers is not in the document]
aelkiss: OK. I'll leave that as-is too!

manifests/profile/haproxy.pp
aelkiss: Done.

manifests/authzd_user.pp
aelkiss: I think in a previous iteration it might have been there multiple times. Fixed now.

spec/defines/authzd_user_spec.rb
aelkiss: It doesn't then `require('File[/var/haproxyctl]')`, but I'm OK with updating the test a little bit.

manifests/authzd_user.pp
aelkiss: Done.

PR #20, opened 2018-07-03
AEIM-1249: copy directory of ssl certs

aelkiss:
With AEIM-1248 this should be genericized per front-end.

### DIFF
--- a/manifests/profile/haproxy.pp
+++ b/manifests/profile/haproxy.pp
@@ -6,7 +6,7 @@
 #
 # @example
 #   include nebula::profile::haproxy
-class nebula::profile::haproxy(String $floating_ip) {
+class nebula::profile::haproxy(String $floating_ip, String $cert_source) {
   service { 'haproxy':
     ensure     => 'running',
     enable     => true,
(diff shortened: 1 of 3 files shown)


aelkiss:
This is cribbed from `manifests/profile/elastic.pp`

The expectation is that the certs for haproxy are in the appropriate place on the puppet master.

This doesn't do anything about deploying certs for the web servers, somewhat intentionally since the certs are in a different form for haproxy (PEM) vs. apache (separate crt / key /chain files)

n.b. `puppet-lint` complains about both `elastic.pp` and `haproxy.pp`, claims it fixes the problem, but really doesn't do anything:

```
aelkiss@claret:~/nebula$ puppet-lint -f manifests/profile/haproxy.pp manifests/profile/elastic.pp
manifests/profile/haproxy.pp - FIXED: puppet:// URL without modules/ found on line 72
manifests/profile/elastic.pp - FIXED: puppet:// URL without modules/ found on line 40
```

manifests/profile/haproxy.pp
```diff
@@ -52,4 +52,26 @@
...
+    file { '/etc/ssl/private/www-lib':
+      ensure  => 'directory',
+      mode    => '0700',
+      owner   => 'haproxy',
+      group   => 'haproxy',
+      recurse => true,
+      purge   => true,
+      links   => 'follow',
+      notify  => Service['haproxy'],
+      source  => "puppet://${cert_source}/www-lib"
```
daaang: Very weird that puppet-lint complains about this. Our Rakefile explicitly says to skip that check.

aelkiss: Probably because I ran `puppet-lint` directly rather than using whatever was specified in the `Rakefile`.

PR #21, opened 2018-07-05
Aeim 1248 balanced_frontend

daaang:


spec/classes/profile/haproxy_spec.rb
```diff
@@ -16,6 +16,7 @@
 
       let(:scotch) { { 'ip' => '111.111.111.123', 'hostname' => 'scotch' } }
       let(:soda)   { { 'ip' => '222.222.222.234', 'hostname' => 'soda' } }
+      let(:third_server) { { 'ip' => '333.333.333.345', 'hostname' => 'third_server' } }
```
daaang: I tried to use Faker::Internet here, like in the keepalive spec, but it kept running twice for reasons I couldn't figure out, so the tests always failed. Eventually, I gave up and gave it static values.

templates/profile/haproxy/frontends.cfg.erb
```diff
@@ -1,16 +1,19 @@
...
   bind <%= @floating_ip -%>:80,<%= @networking['ip'] -%>:80
   stats uri /haproxy?stats
-  default_backend www-lib-<%= @datacenter -%>-http-back
+  default_backend <%= frontend %>-<%= @datacenter -%>-http-back
   http-request set-header X-Client-IP %ci
   http-request set-header X-Forwarded-Proto http
 
-frontend www-lib-<%= @datacenter -%>-https-front
+frontend <%= frontend %>-<%= @datacenter -%>-https-front
   bind <%= @floating_ip -%>:443,<%= @networking['ip'] -%>:443 ssl crt /etc/ssl/private/www-lib
```
aelkiss: We'll need a separate floating IP for each frontend, and that will have to be configured in keepalived too. We could get around that if we track all the virtual hosts for a particular backend, but I'm not sure I want to do that yet.

PR #24, opened 2018-07-10
only use other keepalive nodes as keepalived "unicast peers"

aelkiss:
partially reverts https://github.com/mlibrary/nebula/commit/53e3bd8b#diff-6587d3550c955340798dee887ef692b5R36

### DIFF
--- a/manifests/profile/haproxy/keepalived.pp
+++ b/manifests/profile/haproxy/keepalived.pp
@@ -22,7 +22,8 @@ class nebula::profile::haproxy::keepalived(Hash $floating_ips,
     require    => Package['keepalived'],
   }
 
-  $frontends = balanced_frontends()
+  $nodes_for_class = nodes_for_class($title)
+  $nodes_for_datacenter = nodes_for_datacenter($::datacenter)
   $email = lookup('nebula::root_email')
 
   file { '/etc/keepalived/keepalived.conf':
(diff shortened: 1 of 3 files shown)


manifests/profile/haproxy/keepalived.pp
```diff
@@ -22,7 +22,8 @@
     require    => Package['keepalived'],
   }
 
-  $frontends = balanced_frontends()
+  $nodes_for_class = nodes_for_class($title)
```
daaang: I'm not sure this will work, but I'm definitely interested. I'd thought that classes don't have the `$title` parameter, but I never actually checked. If it don't work in production, it still passes the tests because our `nodes_for_class` mock doesn't pay attention to its input.

spec/classes/profile/haproxy/keepalived_spec.rb
```diff
@@ -22,19 +22,14 @@
...
-      include_context 'with mocked puppetdb functions', 'somedc', %w[thisnode scotch soda coffee]
+      include_context 'with mocked puppetdb functions', 'somedc', %w[thisnode haproxy2]
 
-      before(:each) do
-        stub('balanced_frontends') do |d|
-          allow_call(d).and_return('www-lib': %w[scotch soda], 'svc2': %w[scotch coffee])
-        end
-      end
+      # required for included haproxy role, but isn't tested here
+      before(:each) { stub('balanced_frontends') { |d| allow_call(d).and_return({}) } }
```
daaang: I've created https://tools.lib.umich.edu/jira/browse/AEIM-1328, so this is fine as it is, in this scope.

manifests/profile/haproxy/keepalived.pp
aelkiss: It seems to work as expected in production

PR #25, opened 2018-07-16
Support throttling parameters for haproxy configuration

aelkiss:
- Factors out the frontend/backend config to a separate define (a la vmhost)

- Changes the floating_ips hash to a hash for all per-service config that can include throttling parameters as well as the floating IP

- Adds a bunch of helpers directly in the haproxy service.cfg.erb template (only for lack of a better place to put them, if anybody has a better sense than me of how this works)


aelkiss:
There are some wacky and inconsistent things in terms of how whitespace gets added largely to prevent having to add extraneous whitespace in the tests - that is, the code is a little bit uglier for some of the optional parameters to prevent there being extra trailing whitespace if the optional parameter isn't used. If there's a better way of dealing with that in general I'd be happy to change anything regarding whitespace.

aelkiss:
The last commit https://github.com/mlibrary/nebula/pull/25/commits/34021a1e09efe464c9ee4e10e93d248a638f36fd makes things substantially uglier in terms of how we expect info to be represented in hiera and how it gets passed through from the keepalived role to haproxy. I'd like to reconsider where the floating_ips info gets stored in hiera and the general flow of info between the keepalived/haproxy roles - if you have any thoughts I'm happy to incorporate them. I'll think some about it tomorrow.

aelkiss:
We should put the service configs in a subdirectory and include that subdirectory so that we don't need to itemize the service configs in defaults.

spec/defines/haproxy_service_spec.rb
```diff
@@ -49,13 +49,13 @@
...
           "frontend svc1-hatcher-http-front\n" \
-          "  bind 1.2.3.4:80,40.41.42.43:80\n" \
+          "  bind 1.2.3.4:80\n" \
           "  stats uri \/haproxy?stats\n" \
           "  default_backend svc1-hatcher-http-back\n" \
           "  http-request set-header X-Client-IP %ci\n" \
           "  http-request set-header X-Forwarded-Proto http\n",
           "frontend svc1-hatcher-https-front\n" \
-          "  bind 1.2.3.4:443,40.41.42.43:443 ssl crt /etc/ssl/private/svc1\n" \
+          "  bind 1.2.3.4:443 ssl crt /etc/ssl/private/svc1\n" \
```
daaang: I'm not sure if I was right to change the tests like this—but it made them pass. It looked to me like you intentionally removed the local ips from the template, so I removed them from the tests as well.

aelkiss: Yes, that was exactly the right thing to do - thanks

PR #32, opened 2018-07-25
AEIM-1293: rewrite throttling & whitelisting

aelkiss:
This addresses a couple major issues with the behavior of whitelisted
paths in the previous design that are counter to how throttling
works in the NetScaler implementation:

- If you requested only whitelisted paths, they still incremented the
requst rate. So, if you limited most requests to for example 10 requests
per second, and you did 20 requests to whitelisted paths in one second,
the next request to a non-whitelisted path would get throttled.

- With the temporary ban list in the front end but the whitelists in the
back end, once you had gone over the request rate ALL requests to the
front-end would be denied, even those for whitelisted paths.

Now:

- Throttling only checks the average request rate in the backend -
there's no longer any blocking in the front end (we could bring that
back if desired later while still maintaining the whitelist)

- The whitelist is checked in the front end; there are separate
back-ends for exempted requests. Whitelisted requests don't count
against the limit and will be satisfied even if you've gone over the
request rate

- There is a combined request rate check for http + https requests -
there are separate tracking tables, but the ACL adds the rates and
compares to the limit.

### DIFF
--- a/manifests/haproxy_service.pp
+++ b/manifests/haproxy_service.pp
@@ -19,6 +19,15 @@ define nebula::haproxy_service(
 
   $service = $title
 
+  if $max_requests_per_sec > 0 {
+    file { "/etc/haproxy/errors/${service}509.http":
+      ensure => 'present',
+      mode   => '0644',
+      notify => Service['haproxy'],
+      source => "puppet://errorfiles/${service}509.http"
+    }
+  }
+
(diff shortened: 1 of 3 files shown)


aelkiss:
There's also a couple other small changes:

- puppet now deploys an custom error file for when the bandwidth limit is exceeded
- each server only has one monitoring check - the other backends dependent on the server now use the "track" functionality to rely on the check done in the https backend.

PR #33, opened 2018-07-27
Create named_instance defined resource

malakai97:
Resolves AEIM-1128


manifests/named_instance.pp
```diff
@@ -0,0 +1,129 @@
...
+    ensure => 'present',
+    gid    => $gid,
+  }
+
+  # Add sudoers and passed users to the group
+  (lookup('nebula::users::sudoers').map |$user, $_data| { $user } + $users).each |$user| {
+    exec { "${user} ${title} membership":
+      unless  => "/bin/grep -q ${title}\\S*${user} /etc/group",
+      onlyif  => "/usr/bin/id ${user}",
+      command => "/sbin/usermod -aG ${title} ${user}",
```
daaang: Our users really should be virtual resources, but I think that's out of scope for this ticket.

manifests/named_instance.pp
```diff
@@ -0,0 +1,129 @@
...
+    mode   => '0700',
+    owner  => $uid,
+    group  => $gid,
+  }
+
+  # Install the authorized ssh key for fauxpaas
+  ssh_authorized_key { 'fauxpaas':
+    ensure => 'present',
+    user   => $title,
+    type   => 'ssh-rsa',
```
aelkiss: We probably don't want to assume the key is of type ssh-rsa (e.g. it may be ssh-ecdsa)

manifests/profile/named_instances.pp
```diff
@@ -0,0 +1,29 @@
...
+  String      $fauxpaas_pubkey,
+  String      $fauxpaas_puma_config,
+  Array[Hash] $instances = [],
+) {
+  $instances.each |$instance| {
+    nebula::named_instance { $instance['name']:
+      path        => $instance['path'],
+      uid         => $instance['uid'],
+      gid         => $instance['gid'],
+      users       => $instance['users'],
```
aelkiss: I think this would be more idiomatically written as:

`Hash $instances = {}` in the params

```$instances.each |name,params| { $name:
    * => $params,
    pubkey => $fauxpaas_pubkey,
    puma_config => $fauxpaas_puma_config
}```

I'd be curious what @daaang  thinks about that.

malakai97:
I changed it to a hash, but I did not use the syntactic sugar to map the values, as I felt the clarity was valuable here.

PR #34, opened 2018-08-01
Add puma wrapper to profile::named_instances

malakai97:
Resolves AEIM-1156

This adds a wrapper over puma that can handle applications with legacy
or modern deployments. This should aid in migrating those deployments.

The possible scenarios are (in order of how common they are):

- [x] bundled puma + no fauxpaas config (previous standard)
- [x] bundled puma + fauxpaas config (typical transition pattern)
- [x] no bundled puma + fauxpaas config (current standard)
- [x] no bundled puma + no fauxpaas config (doesn't happen)

This solution addresses all of these scenarios. Those that have been tested on live servers are checked off.

### DIFF
--- a/manifests/profile/named_instances/puma_wrapper.pp
+++ b/manifests/profile/named_instances/puma_wrapper.pp
@@ -1,0 +1,20 @@
+# Copyright (c) 2018 The Regents of the University of Michigan.
+# All Rights Reserved. Licensed according to the terms of the Revised
+# BSD License. See LICENSE.txt for details.
+
+# Wrapper for puma
+# It uses bundled puma if present, system puma otherwise
+# It prefers the fauxpaas puma config over the default location
+#
+# @example
+class nebula::profile::named_instances::puma_wrapper(
+  String  $path
+){
+  file { $path:
+    ensure  => 'present',
+    mode    => '0755',
+    owner   => 'root',
+    group   => 'root',
+    content => template('nebula/profile/named_instances/puma_wrapper/script.erb'),
+  }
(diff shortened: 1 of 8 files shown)


templates/named_instance/puma_wrapper/script.erb
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
+# Managed by puppet (nebula/named_instance/fauxpuma.erb)
+
+RBENV="<%= @rbenv_root -%>/bin/rbenv"
+FAUXPAAS_PUMA_CONFIG="<%= @puma_config -%>"
+DEFAULT_PUMA_CONFIG="config/puma.rb"
+
+$RBENV exec bundle exec puma --version > /dev/null 2>&1
+PUMA_STATUS=$?
```
botimer: This is just clever enough. I like it!

manifests/profile/named_instances/puma_wrapper.pp
```diff
@@ -0,0 +1,20 @@
+# Copyright (c) 2018 The Regents of the University of Michigan.
+# All Rights Reserved. Licensed according to the terms of the Revised
+# BSD License. See LICENSE.txt for details.
+
+# Wrapper for puma
+# It prefers system puma over bundled puma
```
botimer: This comment is a little misleading. In the script, a bundled puma will be used if present, and fall back to the system puma -- and we know that system breaks if there is a different version bundled...

So I guess _we_ prefer system, but defer to any bundled version... right?

PR #35, opened 2018-08-10
AEIM-1410: Allow adding an "only if" condition for throttling

aelkiss:
To minimize disruption to the existing setup with the default backend
throttled and a separate backend exempt from throttling, this is
expressed as a negated condition for the exemption back end - that is,
use the backend exempt from throttling only if the incoming request does
NOT match a particular pattern.

This also changes the optional string parameters to Optional[String]
with a default of undef rather than '' - this seems to be the more
idiomatic way to handle it in puppet given that we still needed
conditionality to handle ''.

### DIFF
--- a/templates/profile/haproxy/service.cfg.erb
+++ b/templates/profile/haproxy/service.cfg.erb
@@ -46,26 +46,46 @@ EOT
     "  server #{hostname} #{ip}:#{port} #{check} cookie s#{ip.split('.').last}"
   end
 
-  def any_whitelists?
-    @whitelists.find { |_,exemptions| exemptions.size() > 0 }
+  def nonempty_whitelists
+    @whitelists.select { |_,exemptions| exemptions.size() > 0 }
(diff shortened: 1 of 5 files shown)


templates/profile/haproxy/service.cfg.erb
```diff
@@ -46,26 +46,46 @@ EOT
...
+    acls = acls + ["!throttle_condition"] if @throttle_condition
+    acls.join(" OR ")
+  end
+
+  def exemption(service_prefix)
+    if any_exemptions?
+      "  use_backend #{service_prefix}-back-exempt if #{exemption_condition}\n"
+    else
+      ""
+    end
```
daaang: Part of me always feels like this is exactly how not to use erb files, but I honestly kind of love this. Everything you need is just right there, but there's still a clear separation between logic and format.

Anyway "Let's Get The Manual"

PR #37, opened 2018-08-30
AEIM-1502: install i40e-dkms if intel x710 network card is installed

aelkiss:
The default driver that comes with Debian Stretch is not stable. See JIRA issue and commit messages for more details.


aelkiss:
done

PR #39, opened 2018-10-23
AEIM-1074 - Ensure that state.yaml never gets bloated

daaang:


daaang:
Since when has Travis been broken? The tests pass on my computer. Also the tests I wrote as part of this PR are passing, even in Travis. I'm torn between the dual philosophies of "never commit if Travis says no" and "fixing Travis is out of scope for this PR."

botimer:
I took a look at this... I'm not really sure _why_ travis failing, but it looks like it started a couple of commits ago. I'm happy to move ahead with merging this, provided that we plan to fix the failures next.

PR #44, opened 2018-11-20
AEIM-1428 - Add CPAN modules to HTVM

botimer:
This is not a beautiful implementation, but it should do the trick.
There is a reasonable check in the unless that just loads the module.
The defined type sits at the nebula level, which may not be right for
the long term.


aelkiss:
Looks great to me. I'll merge it and give it a shot on ht-web-001.

PR #47, opened 2018-11-29
Add cron jobs

malakai97:
Resolves AEIM-1426

This adds the existing cron jobs from our Hathitrust web servers, both common and not-so-common.

Adding the cron tasks themselves was straightforward. The complexity arose from choosing where they should go. In the end, I chose to focus on putting the tasks in the profiles to which they belong. This resulted in a couple new sub-profiles, and some new sub-roles.

It is the topography of the roles-profiles-crontasks that is most in need of review.

See [here](https://tools.lib.umich.edu/jira/secure/attachment/57434/tabs.txt) for a summary of the cron jobs as they existed on our servers, the profile in which they ended up, and the most general role that includes said profile.

### DIFF
--- a/manifests/profile/hathitrust/apache.pp
+++ b/manifests/profile/hathitrust/apache.pp
@@ -178,7 +178,7 @@ class nebula::profile::hathitrust::apache (
   }
 
   cron { 'apache restart':
-    command => '( /etc/init.d/apache2 stop; /bin/sleep 10; /etc/init.d/apache2 start ) > /dev/null',
+    command => '( /bin/systemctl stop apache2; /bin/sleep 10; /bin/systemctl start apache2 ) > /dev/null',
     user    => 'root',
     minute  => '1',
     hour    => '0',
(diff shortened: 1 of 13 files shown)


aelkiss:
We should also disable the cron jobs that require stuff in /l/local/bin until we've deployed all the things these scripts refer to - that would be part of AEIM-1633

botimer:
I'm not sure why the rebase to current master looks like a bunch of new commits... Anyway, this picks up the fix in master for the SSL tests so we can go green before merge (without a reintegration commit).

manifests/profile/hathitrust/apache.pp
```diff
@@ -164,4 +164,24 @@
...
+
+  $http_files = lookup('nebula::http_files')
+  file { '/usr/local/bin/ckapacheconn':
+    ensure => 'present',
+    mode   => '0755',
+    source => "https://${http_files}/ae-utils/bins/ckapacheconn"
+  }
+
+  cron { 'apache restart':
+    command => '( /etc/init.d/apache2 stop; /bin/sleep 10; /etc/init.d/apache2 start ) > /dev/null',
```
aelkiss: This will work, but it would be better to say `/bin/systemctl stop apache2; /bin/sleep 10; /bin/systemctl start apache2`

PR #48, opened 2018-11-30
AEIM-1628 - Look up staff networks for mod_status

botimer:
The approach used here does not add another parameter to the Apache
profile, but I would not be opposed to doing so. This commit does an
explicit lookup in hiera. The fixture data includes a model for mapping
networks in multiple places without repeating the actual data, using
alias interpolation.

### DIFF
--- a/manifests/profile/hathitrust/apache.pp
+++ b/manifests/profile/hathitrust/apache.pp
@@ -27,6 +27,9 @@ class nebula::profile::hathitrust::apache (
     fact_for($nodename, 'networking')['ip']
   }
 
+  $staff_networks = lookup('hathitrust::networks::staff', default_value => []).flatten.unique.map |$network| {
+    "ip ${network['block']}"
+  }.sort
 
   class { 'apache':
     default_vhost          => false,
@@ -83,9 +86,9 @@ class nebula::profile::hathitrust::apache (
   }
 
   class { 'apache::mod::status':
-    requires        =>  {
+    requires => {
       enforce  => 'any',
-      requires => [ 'local' ]
+      requires => [ 'local' ] + $staff_networks
     }
   }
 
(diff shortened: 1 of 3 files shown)


manifests/profile/hathitrust/apache.pp
```diff
@@ -89,10 +92,9 @@
   }
 
   class { 'apache::mod::status':
-    requires        =>  {
+    requires => {
       enforce  => 'any',
-      # TODO look up staff IPs from hiera
-      requires => [ 'local' ]
+      requires => [ 'localhost' ] + $staff_networks
```
aelkiss: Oh, this is actually not correct -- `Require local` is a special apache thing

manifests/profile/hathitrust/apache.pp
```diff
@@ -27,6 +27,9 @@
     fact_for($nodename, 'networking')['ip']
   }
 
+  $staff_networks = lookup('hathitrust::networks::staff').flatten.unique.map |$network| {
+    $network['block']
```
aelkiss: This should be `ip ${network['block']}` so we end up with e.g. `Require ip 10.1.2.0/24` in the Apache config

spec/classes/profile/hathitrust/apache_spec.rb
```diff
@@ -109,6 +109,18 @@ def multiline2re(string)
...
           }
+
+          # Note that we test mod_status in this context rather than above
+          # because the profile relies on a lookup of hiera data. It may make
+          # sense to expose the parameter on the profile so params can be stubbed.
+          it {
+            is_expected.to contain_class('apache::mod::status').with(
+              requires: {
+                'enforce'   => 'any',
+                'requires' => ['localhost', '10.0.1.0/24', '10.0.2.0/24', '10.0.3.0/24']
```
aelkiss:
```ruby
'requires' => ['local','ip 10.0.1.0/24','ip 10.0.2.0/24']
```

manifests/profile/hathitrust/apache.pp
aelkiss: https://httpd.apache.org/docs/trunk/mod/mod_authz_host.html

botimer:
All above are addressed in the squashed/pushed commit.

PR #49, opened 2018-11-30
AEIM-1658 - Retrieve errorfiles over https

daaang:
I haven't tested this yet—I'll comment here once I have.


daaang:
Confirmed: this works as expected, at least on chardonnay.

PR #50, opened 2018-11-30
Add bins from /l/local/bin to /usr/local/bin

malakai97:
Resolves AEIM-1633

This PR is based on aeim-1426/common_crons, as it must be complete in order for that PR to be viable.

This adds the five files described in the commit subjects to their respective profiles. The files are downloaded from ae-files, where they have been uploaded via ae-utils--see that git project for an exact list.

The files were added to /usr/local/bin as we intend /l/local/bin to be a symlink to that directory. Leaving /l/local/bin out of the picture for this issue prevents any race conditions that would need to be resolved via dependency.

See the JIRA for why some files were excluded from this PR.

### DIFF
--- a/manifests/profile/geoip.pp
+++ b/manifests/profile/geoip.pp
@@ -13,11 +13,18 @@ class nebula::profile::geoip () {
   package { 'geoip-bin': }
 
   cron { 'update GeoIP database':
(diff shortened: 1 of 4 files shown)


aelkiss:
The title of the PR says installing them in `/usr/local/bin`, which is where I think they ought to go, but the actual content indicates `/usr/bin`. I'll update that and merge.

PR #52, opened 2018-11-30
AEIM-1661 - Set strict cipher suite for HT web nodes

botimer:
This sets all of the SSL vhosts to use only ECDHE-RSA-AES256-GCM-SHA384
and explicitly sets only TLSv1, 1.1, 1.2 for the protocol. The HAProxy
config already supports this cipher, so no changes are needed there.


aelkiss:
Since this is only for the backchannel between HAproxy and the web host, we should also disable TLS 1.0 and TLS 1.1. I will update that.

PR #57, opened 2018-12-05
AEIM-1449: HTTP fileserver & Debian repository

aelkiss:


aelkiss:
The SSL cert extraction isn't quite all the way there; see AEIM-1236 for additional work. Would be nice to reduce the duplicate knowledge about `/etc/ssl/certs`, `/etc/ssl/private`, etc

aelkiss:
Things that would be nice to get some feedback on:

- ssl cert extraction - OK for now, or is there something quick that we could do that would be better?
- additional `minimum_new_firewall` additional root of the role hierarchy - any ideas on something quick we could do to remove this bifurcation and still minimize duplication between roles?
- puppetdb query for nodes - does this seem OK?

PR #59, opened 2018-12-06
Manage ht /etc/hosts

malakai97:
Resolves AEIM-1648, AEIM-1647

I believe I've captured everything that we care about, but there's enough legacy stuff here that it's worth checking.

I created a test for profile::hathitrust::hosts, but skipped this step for profile::hathitrust::ingest_hosts as I felt the strategy was proven.  Fixture data was updated. 

I have a commit pushed in hiera on the `etc_hosts` branch.


aelkiss:
Merged etc_hosts to production on lensoftruth as well. We could, if we wanted, make this a little less 'magical' with some aliases and/or comments in hiera, but that's out of scope here. Hopefully in the not too distant future we'll be able to do this with exported resources from the solr hosts, so I'm happy to leave it as-is in the hopes it will encourage us to actually do that.

PR #61, opened 2018-12-07
Usergroup defined type

daaang:


daaang:
obviously i'll resolve the common.yaml issue before doing a real pull request, in part since i actually want all this to fail if it isn't set in lensoftruth.

spec/fixtures/hiera/default.yaml
```diff
@@ -98,37 +98,46 @@ nebula::profile::puppet::master::fileservers:
...
+      AAAA
+      AAAA
+  invalid_special_admin:
+    type: ssh-dsa
+    host: invalid.special
+    key: >-
+      BBBB
+      BBBB
+      BBBB
+
```
aelkiss: Do we need this built out this much in `default.yaml`? Seems like it might be preferable just to provide empty lists

spec/fixtures/hiera/usergroups.yaml
```diff
@@ -0,0 +1,29 @@
...
+# BSD License. See LICENSE.txt for details.
+nebula::usergroup::membership:
+  examplegroup1:
+  - exampleuser1
+  - exampleuser2
+  examplegroup2:
+  - exampleuser2
+  - exampleuser3
+
+nebula::profile::groups::all_groups:
```
aelkiss: What about something like:

```yaml
nebula::groups:
  examplegroup1:
    gid: 1234
    members:
    - exampleuser1
    - exampleuser2
  examplegroup2:
    gid: 2486
    members:
    - exampleuser2
    - exampleuser3
```
That would keep all the information about groups a little more closely together at the expense of making using the data in the profiles/functions/etc slightly more complicated.

lib/puppet/functions/nebula/user_group_membership.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+# Copyright (c) 2018 The Regents of the University of Michigan.
+# All Rights Reserved. Licensed according to the terms of the Revised
+# BSD License. See LICENSE.txt for details.
+
+Puppet::Functions.create_function(:'nebula::user_group_membership') do
```
aelkiss: It might be helpful to have a more explicit comment/example that it returns a hash where the keys are users and the values are the list of all groups for the user.

It might be nice to write this in a more functional style, but if this works I'm inclined to leave it alone (especially given the vagaries of the interaction of puppet data types & ruby)

aelkiss: It also would be nice to have tests specifically for this function, if for no other reason than it would aid later refactoring.

manifests/usergroup.pp
```diff
@@ -0,0 +1,23 @@
+# Copyright (c) 2018 The Regents of the University of Michigan.
+# All Rights Reserved. Licensed according to the terms of the Revised
+# BSD License. See LICENSE.txt for details.
+
+# nebula::usergroup
+#
+# A description of what this defined type does
```
aelkiss: This should be corrected

spec/fixtures/hiera/usergroups.yaml
daaang: I went back and forth on this and eventually chose separate hashes because:

1.  It makes the profiles/functions/etc slightly less complicated (particularly in cases like sudo, where we don't want puppet to manage the group itself, but we do want it to manage membership—as it stands, we can just leave it out of the gid group, and we don't need to write any puppet code checking for a gid key-value in each group).
2.  The two lists have different purposes from our perspective. One should only be messed with when we create a new group, but the other is subject to change all the time.

But yeah, disadvantages include the complexity of having two parameters instead of one and a repetition of every group name. I'm not convinced that either is the "right" way to do it. Really, if anything, the right way is to figure out how to handle users and groups outside of hieradata, so I guess I'm inclined to want to think about that instead of this and to stick with the easy solution in the meantime.

spec/fixtures/hiera/default.yaml
daaang: I'm sure I can update the tests and put that in a separate yaml file. Good eye.

manifests/usergroup.pp
daaang: oooooooops will do

spec/fixtures/hiera/usergroups.yaml
aelkiss: OK, sounds good.

lib/puppet/functions/nebula/user_group_membership.rb
daaang: At first I didn't want to do this because I was thinking of it as a private method, but I realized it could pretty easily be generalized into `inverted_hashlist()`, so I went ahead and did that. Also, that way, the particular hiera lookup is named in a `.pp` file instead of a `.rb` file, which feels better to me. Also also, Bryan taught me about the `tap` method, so it looks a lot more functional now.

lib/puppet/functions/inverted_hashlist.rb
```diff
@@ -0,0 +1,32 @@
...
+  def from_lookup(lookup_string)
+    from_hash(call_function('lookup', lookup_string))
+  end
+
+  def from_hash(input_hash)
+    Hash.new { |h, k| h[k] = [] }.tap do |result|
+      input_hash.each do |key, values|
+        values.each do |value|
+          result[value] << key
+          result[value].sort!
```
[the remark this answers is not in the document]
daaang: I'm not sure how better to test that `{a: [1], b: [1]}` returns `{1: [a, b]}`, since it might also return `{1: [b, a]}` and still be what we want but fail the test. I couldn't find an easy way to assert, say, that `result['1']` has length 2 and contains both `a` and `b` or that the resulting array is either `[a, b]` or `[b, a]`.

All I have to work with without doing anything even hackier than sorting each array is `.and_returns()`, which doesn't give me a lot of room. Any ideas?

PR #67, opened 2018-12-13
AWS: Set search domain via dhclient, don't set dns servers

rrotter:
For AWS hosts, EC2 service sets correct DNS via DHCP. Keep this, but override the search domain in dhclient.conf. Restarts networking, it's coarse, but I couldn't find anything cleaner (dhclient -r only looks for a new lease, doesn't appear to reload the config).

### DIFF
--- a/manifests/profile/dns/aws.pp
+++ b/manifests/profile/dns/aws.pp
@@ -1,0 +1,33 @@
+# Copyright (c) 2018 The Regents of the University of Michigan.
+# All Rights Reserved. Licensed according to the terms of the Revised
+# BSD License. See LICENSE.txt for details.
+
+# nebula::profile::dns::aws
+#
+# Configure dhclient for AWS hosts to add our search domain to resolv.conf,
+# also suppresses AWS/DHCP supplied search domain.
+#
+# @example
+#   include nebula::profile::dns::aws
+class nebula::profile::dns::aws {
+
+  $searchpaths = lookup('nebula::resolv_conf::searchpath')
+  $searchpath_str = join($searchpaths, ' ')
+
+  exec { 'restart_networking':
+    command     => '/bin/systemctl restart networking',
+    refreshonly => true,
+  }
+
+  file_line {
+    default:
+      path   => '/etc/dhcp/dhclient.conf',
+      notify => Exec['restart_networking'],
+    ;
+    'search_domain':
+      line => "supersede domain-search \"${searchpath_str}\";",
+    ;
+    'domain_name':
+      line => 'supersede domain-name "";',
+  }
+}
--- a/spec/classes/profile/dns/aws_spec.rb
+++ b/spec/classes/profile/dns/aws_spec.rb
@@ -1,0 +1,24 @@
+# frozen_string_literal: true
+
+# Copyright (c) 2018 The Regents of the University of Michigan.
+# All Rights Reserved. Licensed according to the terms of the Revised
+# BSD License. See LICENSE.txt for details.
+require 'spec_helper'
+
+describe 'nebula::profile::dns::aws' do
+  on_supported_os.each do |os, os_facts|
+    context "on #{os}" do
+      let(:facts) { os_facts }
+
+      it { is_expected.not_to contain_class('resolv_conf')   }
+      it { is_expected.to contain_exec('restart_networking') }
+      it { is_expected.to contain_file_line('domain_name') }
+      # search_domain should match content of nebula::resolv_conf::searchpath
+      it do
+        is_expected.to contain_file_line('search_domain').with_line(
+          'supersede domain-search "searchpath.default.invalid";',
+        )
+      end
+    end
+  end
+end
--- a/manifests/role/aws.pp
+++ b/manifests/role/aws.pp
@@ -21,7 +21,7 @@ class nebula::role::aws {
     }
   }
 
-  include nebula::profile::dns::standard
+  include nebula::profile::dns::aws
   include nebula::profile::elastic::metricbeat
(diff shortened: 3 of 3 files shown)


aelkiss:
`systemctl restart networking` has a nasty habit of breaking everything under most conditions. Have you tested that it works OK for AWS machines?

rrotter:
Yes, I've tested it successfully, and just tested once again on a fresh EC2 host to be sure. Couldn't find any other way to get the change picked up besides a reboot.

manifests/profile/dns/standard.pp
```diff
@@ -9,8 +9,29 @@
 # @example
 #   include nebula::profile::dns::standard
 class nebula::profile::dns::standard {
-  class { 'resolv_conf':
-    nameservers => lookup('nebula::resolv_conf::nameservers'),
-    searchpath  => lookup('nebula::resolv_conf::searchpath'),
+  exec { 'restart_networking':
+    command     => '/bin/systemctl restart networking',
+    refreshonly => true,
```
daaang: Since this is only ever called from inside the `if` condition, would it make sense to put this in there as well? That would also keep it off our own boxes, so that it doesn't somehow get called on any of them.

rrotter: The location was intentional, but you make a good point about polluting the environment. I'll change it.

rrotter:
Per my discussion w/ @aelkiss I moved profile::dns::standard changes into a new profile::dns::aws. Spec also included in this revision.

manifests/profile/dns/aws.pp
```diff
@@ -0,0 +1,32 @@
+# Copyright (c) 2018 The Regents of the University of Michigan.
+# All Rights Reserved. Licensed according to the terms of the Revised
+# BSD License. See LICENSE.txt for details.
+
+# nebula::profile::dns::standard
```
botimer: Mismatched -- should be ::aws.

manifests/profile/dns/aws.pp
```diff
@@ -0,0 +1,32 @@
...
+  exec { 'restart_networking':
+    command     => '/bin/systemctl restart networking',
+    refreshonly => true,
+  }
+
+  file_line {
+    default:
+      path   => '/etc/dhcp/dhclient.conf',
+      notify => Exec['restart_networking'],
+    ;
```
botimer: I don't recognize this syntax. What's going on with the `default:` and the dangling `;`?

manifests/profile/dns/aws.pp
```diff
@@ -0,0 +1,32 @@
...
+
+  $searchpaths = lookup('nebula::resolv_conf::searchpath')
+  $searchpath_str = join($searchpaths, ' ')
+
+  exec { 'restart_networking':
+    command     => '/bin/systemctl restart networking',
+    refreshonly => true,
+  }
+
+  file_line {
```
botimer: This probably seems obvious to those who do DNS/DHCP more often, but maybe a comment either here or at the class level would be good to say what we're expecting to get from hiera and the effect of this addition or replacement.

As with most comments, it's a balance -- I don't think we have to explain everything done by the tools we're configuring with puppet, but this change seems to encode some rationale/practice of ours I don't totally understand by just looking at this code. I think probably half of that is not knowing dhclient well and half of it is that we're doing something with .lib vs .umdl and bare hostnames.

spec/classes/profile/dns/aws_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+  on_supported_os.each do |os, os_facts|
+    context "on #{os}" do
+      let(:facts) { os_facts }
+
+      it { is_expected.not_to contain_class('resolv_conf')   }
+      it { is_expected.to contain_exec('restart_networking') }
+      it { is_expected.to contain_file_line('domain_name') }
+      it do
+        is_expected.to contain_file_line('search_domain').with_line(
+          'supersede domain-search "searchpath.default.invalid";',
```
botimer: Maybe a one-line comment here that this literal is coming from the hiera fixture under the `nebula::resolv_conf::searchpath` key.

manifests/profile/dns/aws.pp
rrotter: Good catch on my copy-pasta. Fixed.

manifests/profile/dns/aws.pp
rrotter: Added a class level comment when fixing the aforementioned copy-pasta. Clearer?

spec/classes/profile/dns/aws_spec.rb
rrotter: Done.

manifests/profile/dns/aws.pp
botimer: Yup, looks great.

manifests/profile/dns/aws.pp
botimer: Ahh, thanks. "Full syntax" looks a little wacky to me. I can go either way, but here's an alternative for consideration (containing a more explicit resource title and using two resources since the default bits are simple and locality is very high):

```
file_line { 'DHCP local search domain':
  path   => '/etc/dhcp/dhclient.conf',
  notify => Exec['restart_networking'],
  line => "supersede domain-search \"${searchpath_str}\";",
}

file_line { 'DHCP local domain name':
  path   => '/etc/dhcp/dhclient.conf',
  notify => Exec['restart_networking'],
  line => 'supersede domain-name "";',
}
```

aelkiss: I think it only looks wacky because we haven't used it much before. I think iit makes a lot of sense to use this syntax here.

PR #68, opened 2018-12-13
AEIM-1687: remove nebula::profile::base::firewall::ipv4

aelkiss:
- All stretch nodes are using nebula::proflie::networking::firewall and
managing firewall resources now

- The "minimum" role no longer has a flag for whether or not to manage
firewall resources

- jessie machines continue to have an unmanaged firewall


aelkiss:
This also removes a bunch of the duplicative tests in the roles for the firewall status, since firewall status no longer differs by role.

When we have common tests for all roles, the tests that are currently in the "minimum" role can/should move there. I don't know that they need to be duplicated in the base profile test as well but I suppose it doesn't hurt.

PR #77, opened 2019-01-17
AEIM-1434: profiles and types for unison

aelkiss:
Roles should include `nebula::profile::unison`. Individual nodes should
indicate in their specific hiera which unison instances they are clients
or servers for. Individual unison instances are configured in hiera via
`nebula::unison::${INSTANCE}`. See `spec/fixtures/hiera/unison.yaml` for an
example.

This works via the unison 'server' type exporting a 'client' resource
which is then realized by the client. There isn't a way to test via
puppet rspec if exported resources are properly realized on the
importing node, but I have tested it on actual nodes and it appears to
work as expected.


aelkiss:
That change should be pretty easy to make here. We might then be able to drop the `nebula::profile::unison` entirely and just use `ensure_resource` for any path prerequisites.

PR #78, opened 2019-01-28
Break up haproxy config; use exported resources (rather than puppetdb query)

aelkiss:
See comments in commit. Ultimately, the motivation here is to make it easier for individual nodes to vary how they bind to haproxy by adding or parameters to haproxy::binding.

We need to think about how we would migrate to this in production. I think we would want to test a single throwaway web host & haproxy node on this branch first to make sure that everything with exported resources worked as expected

Then for production:
1) disable puppet on all webserver & haproxy nodes
2) take a datacenter out of service
3) run haproxy on web server & haproxy nodes there until everything converged

For the MACC where www.lib fronts some nonredundant services, we might want to do something like leaving one web server bound to a haproxy node with the old config, then doing it after the other two are bound to the haproxy node with the new setup. Not sure. We will need to think carefully about it.


aelkiss:
Still to-do: actually support the https binding for the apache nodes.

spec/defines/haproxy_service_spec.rb
```diff
@@ -125,19 +114,25 @@
...
-              "backend svc1-hatcher-https-back-exempt\n" \
-              "  server scotch 111.111.111.123:443 track svc1-hatcher-https-back/scotch cookie s123\n" \
-              "  server soda 222.222.222.234:443 track svc1-hatcher-https-back/soda cookie s234\n",
-            ].each do |stanza|
-              it "contains the stanza #{stanza.split("\n").first}" do
-                is_expected.to contain_file(service_config).with_content(%r{#{stanza}}m)
+              "server scotch 111.111.111.123:443 track svc1-dc1-https-back/scotch cookie s123\n" \
+              "server soda 222.222.222.234:443 track svc1-dc1-https-back/soda cookie s234\n",
+            ].each do |fragment|
+              xit do
```
aelkiss: more superceded tests - move to binding

manifests/haproxy/binding.pp
```diff
@@ -0,0 +1,50 @@
...
+
+  $last_octet = $::ipaddress.split('\.')[-1]
+  $cookie = "cookie s${last_octet}"
+  $track = "track ${service}-${::datacenter}-https-back/${::hostname}"
+
+  # TODO - handle case where https_offload = false - ie proxy https to server
+
+  concat_fragment { "${service}-${::datacenter}-http ${::hostname} binding":
+    target  => "/etc/haproxy/services.d/${service}-http.cfg",
+    order   => '04',
```
aelkiss: should look to see how order is used in "real" puppet modules

manifests/haproxy/binding.pp
```diff
@@ -0,0 +1,50 @@
...
+  }
+
+  concat_fragment { "${service}-${::datacenter}-https ${::hostname} binding":
+    target  => "/etc/haproxy/services.d/${service}-https.cfg",
+    order   => '04',
+    content => "server ${::hostname} ${::ipaddress}:443 check ${cookie}\n",
+    tag     => "${service}-${::datacenter}-https_binding"
+  }
+
+  concat_fragment { "${service}-${::datacenter}-http ${::hostname} exempt binding":
```
aelkiss: need to be virtual resources

manifests/haproxy_service.pp
```diff
@@ -83,17 +80,18 @@
       }
     }
 
-    # <%= nodes.map { |hostname,ip| backend_server(hostname,ip,options[:port],check_or_track(protocol,service_loc,hostname)) }.join("\n") %>
+    Concat::Fragment <<| tag == "${service_prefix}_binding" |>>
```
aelkiss: superceded, doesn't need to be there

manifests/haproxy_service.pp
```diff
@@ -83,17 +80,18 @@
...
       concat_fragment { "${service_prefix} back-exempt":
         target  => $service_cfg,
         content => "backend ${service_prefix}-back-exempt\n",
-        order   => '04'
+        order   => '05'
       }
-      # <%= nodes.map { |hostname,ip| backend_server(hostname,ip,options[:port],track(service_loc,hostname)) }.join("\n") %>
     }
 
+    Concat::Fragment <<| tag == "${service_prefix}-exempt_binding" |>>
```
aelkiss: should make exemption bindings virtual & realize here

manifests/profile/haproxy.pp
```diff
@@ -66,16 +63,35 @@
     require    => Package['keepalived'],
   }
 
-  $nodes_for_class = nodes_for_class($title)
-  $nodes_for_datacenter = nodes_for_datacenter($::datacenter)
   $email = lookup('nebula::root_email')
 
-  file { '/etc/keepalived/keepalived.conf':
-    ensure  => 'present',
+  concat { '/etc/keepalived/keepalived.conf':
```
aelkiss: see if we can put this in a different commit

manifests/profile/haproxy.pp
```diff
@@ -86,4 +102,13 @@
     content => template('nebula/profile/haproxy/keepalived/sysctl.conf.erb'),
   }
 
+  @@firewall { "200 HTTP: HAProxy ${::hostname}":
```
aelkiss: should be a separate commit

manifests/haproxy/binding.pp
aelkiss: looks fine (looking at e.g. the puppetlabs apache module); also, we're likely to migrate longer term to the puppetlabs haproxy module, so I think this is OK as is

aelkiss:
I added some documentation with an overview of the exported & virtual resources used by haproxy here: https://tools.lib.umich.edu/confluence/display/LAS/haproxy

I'm not sure if this is clear enough or if a diagram would be more helpful.

PR #79, opened 2019-01-29
[WIP] AEIM-1760 - Export instances for Apache configuration

botimer:
This introduces a new resource type to be exported by the named instance
and collected by the appropriate web servers. There are definitely some
rough edges, like the number of values that are shuttled along between
the components, but it's a good start. The template is ported almost
verbatim from ansible-predeploy, with the exception being that the
"extra config" is pruned in favor of "sendfile_path", the one feature of
the inlined config that has been used so far. If we need to model
instances more richly, we can do that, but I hesitate to preserve a
literal expansion from hiera.

The instances in lens will need a number of new keys for this to work.

### DIFF
--- a/manifests/named_instance.pp
+++ b/manifests/named_instance.pp
@@ -1,4 +1,4 @@
-# Copyright (c) 2018 The Regents of the University of Michigan.
+# Copyright (c) 2019 The Regents of the University of Michigan.
 # All Rights Reserved. Licensed according to the terms of the Revised
 # BSD License. See LICENSE.txt for details.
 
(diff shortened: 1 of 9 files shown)


manifests/profile/named_instances.pp
```diff
@@ -25,14 +25,20 @@
 
   $instances.each |$instance| {
```
aelkiss:
```
$defaults = {
  puma_wrapper => $puma_wrapper,
  pubkey => $fauxpaas_pubkey,
  puma_config => $fauxpaas_puma_config
}
create_resources(nebula::named_instance, $instances, $defaults)

File: manifests/named_instance.pp [hunk 2]
```diff
@@ -157,6 +164,14 @@
     ]
   }
 
+  @@nebula::profile::proxied_app { $title:
```
aelkiss: I think you'll want to keep this as-is, since some of these parameters are also used elsewhere.

manifests/profile/proxied_app.pp
```diff
@@ -0,0 +1,57 @@
...
+  # These are straightforward translations to maintain parity with the
+  # ansible template. All of these names and semantics should be revisited
+  # and likely decomposed into resources/fragments that can be assembled
+  # by the web host after being exported or virtualized wherever needed.
+  $apache_app_hostname  = $hostname
+  $apache_app_name      = $title
+  $apache_cosign_factor = "UMICH.EDU"
+  $apache_domain        = $public_hostname
+  $apache_port          = $port
+  $apache_ssl_crt       = "${public_hostname}.crt"
```
aelkiss: The ansible setup takes the name of the ssl cert as a parameter, and I think we will probably want to do that too.

botimer: Thanks. These suggestions are on the branch now.

PR #82, opened 2019-02-06
AEIM-1760 - Add tests for vhost template

botimer:


botimer:
A couple of notes on this... There are some iffy parts in the template that are not addressed here, but identified with TODOs. I've left them in the source mostly because I want to discuss direction before filing a bunch of issues. There is also incomplete coverage of the conditionality for much the same reason: the conditional behavior might not be quite right now and modeling as resources would make testing it easier. So, again, I'd like to discuss direction and timeline before going too much further.

botimer:
Thanks. Good point on the match vs. include -- I think I expected to be a bit fuzzier, and resorted to more literal blocks. I think those can go to include. It also relates to the indentation. It seemed that all whitespace to the first indentation after the squiggle was stripped, making it hard to combine squiggles and the various two-space sections we want to match. I used the dash heredocs to counteract that, but maybe it's easier to fix than I thought. We'll see if RSpec and Rubocop agree.

I'll try switching to all includes and a couple of indentation alternatives. More very soon.

PR #83, opened 2019-02-06
AEIM-1057 - Add logrotate to unison

daaang:
Since this is using a new module, I'd like to test it in an environment before merging into production, to make sure the module does what we want.

### DIFF
--- a/manifests/profile/unison.pp
+++ b/manifests/profile/unison.pp
@@ -12,6 +12,15 @@ class nebula::profile::unison (
   Array $servers = [],
   Array $clients = []
 ) {
+  logrotate::rule { 'unison':
+    path          => '/var/log/unison.log',
+    rotate        => 7,
+    rotate_every  => 'day',
+    missingok     => true,
+    ifempty       => false,
+    delaycompress => true,
+    compress      => true,
+  }
 
   $servers.each |String $instance| {
     nebula::unison::server { $instance:
--- a/manifests/profile/unison.pp
+++ b/manifests/profile/unison.pp
@@ -15,7 +15,7 @@ class nebula::profile::unison (
   include nebula::profile::logrotate
 
   logrotate::rule { 'unison':
-    path          => '/var/log/unison.log',
+    path          => '/var/log/unison*.log',
     rotate        => 7,
     rotate_every  => 'day',
     missingok     => true,
--- a/manifests/profile/unison.pp
+++ b/manifests/profile/unison.pp
@@ -12,6 +12,8 @@ class nebula::profile::unison (
   Array $servers = [],
   Array $clients = []
 ) {
+  include nebula::profile::logrotate
+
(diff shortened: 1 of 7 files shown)


daaang:
Debian's default cron.daily/logrotate looks like this:

```bash
test -x /usr/sbin/logrotate || exit 0
/usr/sbin/logrotate /etc/logrotate.conf
```

Puppet replaces it with this one:

```bash
OUTPUT=$(/usr/sbin/logrotate /etc/logrotate.conf 2>&1)
EXITVALUE=$?
if [ $EXITVALUE != 0 ]; then
    /usr/bin/logger -t logrotate "ALERT exited abnormally with [$EXITVALUE]"
    echo "${OUTPUT}"
fi
exit $EXITVALUE
```

Which, while different, doesn't offend me. So unless anyone objects, I'll just let puppet do its thing in this case.

daaang:
Precisely, these are the four files that get created or modified:

/etc/logrotate.conf:

```
# THIS FILE IS AUTOMATICALLY DISTRIBUTED BY PUPPET.
# ANY CHANGES WILL BE OVERWRITTEN.

create
rotate 4
weekly

# configurable file rotations
include /etc/logrotate.d
```

/etc/logrotate.d/debian_btmp:

```
# THIS FILE IS AUTOMATICALLY DISTRIBUTED BY PUPPET.  ANY CHANGES WILL BE
# OVERWRITTEN.

/var/log/btmp {
  create 0660 root utmp
  missingok
  monthly
  rotate 1
}
```

/etc/logrotate.d/debian_wtmp:

```
# THIS FILE IS AUTOMATICALLY DISTRIBUTED BY PUPPET.  ANY CHANGES WILL BE
# OVERWRITTEN.

/var/log/wtmp {
  create 0664 root utmp
  missingok
  monthly
  rotate 1
}
```

/etc/logrotate.d/unison:

```
# THIS FILE IS AUTOMATICALLY DISTRIBUTED BY PUPPET.  ANY CHANGES WILL BE
# OVERWRITTEN.

/var/log/unison.log {
  compress
  daily
  delaycompress
  missingok
  notifempty
  rotate 7
}
```

manifests/profile/unison.pp
```diff
@@ -12,6 +12,17 @@
   Array $servers = [],
   Array $clients = []
 ) {
+  include nebula::profile::logrotate
+
+  logrotate::rule { 'unison':
+    path          => '/var/log/unison.log',
```
aelkiss: `/var/log/unison.log` won't be sufficient; the client logs to `/var/log/unison_<%= @title %>.log`. I believe wildcards can be specified in the logrotate config (that's what we're currently doing for apache), so I think just changing it to `/var/log/unison*.log` should be fine.

daaang: I've done exactly that, so this should be ready to merge.

PR #85, opened 2019-02-07
AEIM-1758: configure databases for moku-deployed apps via puppet

aelkiss:
- Creates the user and the database with the puppetlabs mysql module, running on (at least one of) the hosts that the app is deployed on

- Assumes that credentials that can connect to the database and create users are in `/root/.my.cnf` on the app server (puppetlabs-mysql assumption)

- Database is only created if user/password parameters are given and create_database is true.

### DIFF
--- a/.fixtures.yml
+++ b/.fixtures.yml
@@ -24,6 +24,9 @@ fixtures:
     lvm:
       repo: "puppetlabs/lvm"
       ref: "1.0.1"
+    mysql:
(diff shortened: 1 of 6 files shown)


aelkiss:
The provisioned user defaults to e.g. `'someuser'@'localhost'`, but typically we want to default to the mysql default which is any host. I will check how to do that with `puppetlabs-mysql`

PR #87, opened 2019-02-07
AEIM-1809: move monitoring to babel vhost

aelkiss:
monitoring via haproxy wasn't working because it happened over ssl and the monitor wasn't listening over ssl. This moves the monitoring config to the babel vhost and removes the (now-useless) default vhost.


aelkiss:
Also moves the logging cron jobs & binding & adds specs for the prod & test roles

manifests/role/webhost/htvm/test.pp
```diff
@@ -7,15 +7,25 @@
...
 #   include nebula::role::webhost::htvm::test
 class nebula::role::webhost::htvm::test {
-  @@nebula::haproxy::binding { "${::hostname} test-hathitrust":
-    service       => 'test-hathitrust',
-    https_offload => true,
-    datacenter    => $::datacenter,
-    hostname      => $::hostname,
-    ipaddress     => $::ipaddress
-  }
+  # @@nebula::haproxy::binding { "${::hostname} test-hathitrust":
```
aelkiss: note these are temporary

PR #90, opened 2019-02-09
Initialize solr core for new named instance

rrotter:
Creare dummy "release 0" for each named instance
If there isn't a released version yet, link this one and init the solr core

### DIFF
--- a/manifests/named_instance.pp
+++ b/manifests/named_instance.pp
@@ -21,6 +21,11 @@ define nebula::named_instance(
   String        $url_root = '/',
   String        $protocol = 'http',         # proxy protocol, not user to front-end
   String        $hostname = "app-${title}", # app host
+  String        $solr_host = 'localhost',
+  Integer       $solr_port = 8081,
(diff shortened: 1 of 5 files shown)


aelkiss:
So, I think this will work, but TBH I'm not totally happy with it, and I'm thinking it would be better for it to be moku's job to start the solr core the first time the app is deployed. In particular, I'm concerned that there are cases in which the core would need to be unloaded and then re-created (using the CoreAdmin API) to get it to use a new config, and if moku is going to do that anyway it might as well just start the core on the first deploy.

Moving the responsibility to start the core to moku would also let us avoid the 'default config' stuff here, substantially simplify the puppet side of things, and generally stick to the principle that puppet should only need to do things the deployment user can't do.

We would still create the solr core directory here and set up the symlinks into the app.

I'd also like some feedback on the naming of parameters in the `solr_core` defined type, particularly `config_dir` (the subdirectory under `$instance_home/solr` where the solr core should be)

PR #91, opened 2019-02-11
AEIM-1771: Remove unison drop-ins

aelkiss:
Puts everything in the main systemd service config.

Doesn't quite eliminate the client profile as there are some things
shared across unison instances (e.g. the unison home directory, the
unisonsync script)


manifests/unison/server.pp
```diff
@@ -6,20 +6,17 @@
   String  $root,
   Array[String] $paths,
   Array[String] $filesystems,
+  String $home = '/root',
   Integer $port = 2647
 ) {
-  include nebula::profile::unison::server
+  ensure_packages(['unison'])
```
malakai97: How does this relate to the deleted include? I'm also not sure what `ensure_packages` does and can't find it in the puppet docs. Why this over a `package{}` stanza?

spec/defines/unison_client_spec.rb
```diff
@@ -21,8 +21,25 @@
       end
 
       it do
-        is_expected.to contain_file('/etc/systemd/system/unison-client@myinstance.service.d/drop-in.conf')
-          .with_content(%r{.*Requires=fs1.mount.*ExecStart.*/usr/local/bin/unisonsync myinstance.*}m)
+        is_expected.to contain_file('/etc/systemd/system/unison-client-myinstance.service')
```
malakai97: Encoding the entire file here is going to make this test break if we ever change the template at all. Instead, I'd prefer to match on the smallest bit of pertinent information, perhaps the ExecStart.

spec/defines/unison_server_spec.rb
```diff
@@ -20,15 +20,29 @@
       end
 
       it { is_expected.to contain_package('unison') }
-      it { is_expected.to contain_file('/etc/systemd/system/unison@.service') }
 
       it do
-        is_expected.to contain_file('/etc/systemd/system/unison@myinstance.service.d/drop-in.conf')
-          .with_content(%r{.*Requires=fs1.mount.*ExecStart.*-socket 12345.*}m)
+        is_expected.to contain_file('/etc/systemd/system/unison-myinstance.service')
```
malakai97: I prefer the regex matcher here as well.

manifests/unison/server.pp
aelkiss: `ensure_packages` will create the package resource if it doesn't already exist; unlike a straight package resource it doesn't cause an error if there are multiple `ensure_packages` with the same package: https://forge.puppet.com/puppetlabs/stdlib#ensure_packages

botimer: I had the same question, but I found `ensure_packages` in the docs and it is apparently preferred when you have multiple resources that depend on a package and don't want a dependency between them (neither is really primary). It's for mitigation of potential duplicate resources, which would tank the catalog.

aelkiss: Yes. There are options we could use to avoid `ensure_packages` if we needed to (i.e. put the package resource in the main unison profile) rather than doing this in a defined type, but this doesn't hurt anything.

PR #92, opened 2019-02-13
[WIP] AEIM-1762 moku init

daaang:
Stealing the opening post here.

Open questions:
- Is it worth trying to test the mysql variables? I thought not really, since it's the same logic as everything else.
- How do we get the solr core url from named_instances/solr_core.pp?


malakai97:
Stole the opening post and asked a question @daaang

manifests/profile/moku.pp
```diff
@@ -0,0 +1,130 @@
...
+
+    Concat_fragment <<| target == "${name} deploy init" |>>
+
+    concat_fragment {
+      default:
+        target  => "${name} deploy init",
+        ;
+
+      "${name} deploy init instance.source.url":
+        content => "{\"instance\": {\"source\": {\"url\": \"${instance['source_url']}\"}}}",
```
aelkiss: It might be easier to read/maintain (as would the others below) if this was specified as:

```
{instance => {source => {url => $instance['source_url] }}}.to_json
```

daaang:
I don't really like `ensure_resource` in general, and I like it even less used in conjunction with exported resources, but (a) `ensure_resource` is already all over moku stuff, and (b) this is pretty behind schedule, and we can always refactor later, once it's actually working.

aelkiss:
OK -- let's make a ticket for that then, and we should plan on talking to try to figure out a plan that both allows the named instance define to be responsible for what parameters go to moku and avoids `ensure_resource`. It would be helpful if you could be more specific about your objections to `ensure_resource` -- what problems could it cause down the road, etc, -- since I have been using it more and more.

aelkiss:
I guess one potential objection that I see to `ensure_resource` is that it isn't super clear what would happen if you do it twice with two different sets of parameters - that seems like it probably wouldn't work super well.

One quick solution I thought of would be to declare a particular node the 'primary' one and have it be responsible for exporting the concat fragments for moku - that logic is already there for e.g. the mysql database.

aelkiss:
superceded by #98

PR #93, opened 2019-02-15
use puppetlabs-ntp

aelkiss:
changing ntp servers with file_line or using multiple ntp servers didn't
work very well

this should generate an ntp config that is similar to the debian defaults + using our servers except for collection of statistics that we don't use anyway

### DIFF
--- a/.fixtures.yml
+++ b/.fixtures.yml
@@ -27,6 +27,9 @@ fixtures:
     mysql:
       repo: "puppetlabs/mysql"
       ref: "8.0.0"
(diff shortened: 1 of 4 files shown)


aelkiss:
note this removes some of the tests about requires/ordering, since that is really now within the scope of `puppetlabs-ntp` rather than our code. if we note any ordering problems we can either make a PR to `puppetlabs-ntp`, add a failing test in our code and then add whatever ordering requirements are needed, or a combination of both.

PR #94, opened 2019-02-21
AEIM-1850: add remaining libadm & roger cron jobs to nebula

aelkiss:
Makes home directory for the crons, user the jobs run as, and the mail
recipient configurable via hiera. Adds remaining cron jobs from libadm &
roger crontab.


aelkiss:
Please review but coordinate before merging as lensoftruth changes are also required.

PR #97, opened 2019-02-25
moku init + solr core

aelkiss:
supercedes #92

PR #98, opened 2019-02-27
moku init with solr params, take 2

aelkiss:
Supercedes #97, but I left that branch / PR as is for comparison purposes.

The thing of most interest is the description of the last commit.

This changes the flow of information:

- moku node has named_instance resources that:
  - creates concat fragment resources for the moku config on the moku node
  - creates solr_param resources for the moku config on the moku node
  - export 'named_instance::app', 'named_instance::proxy', and
'named_instance::solr_core' resources

This ensures there is a clear responsibility for the flow of
information and that each piece of information has a single resource
responsible for it.

At the named_instance level, some of the parameters used only by the app
or proxy are bundled into app and proxy hashes, which helps reduce the
number of parameters in named_instance and clarify the purpose of some
of the parameters for named instances in hiera.

Also, this changes all json content to use to_json. I think without all
the escaped quotes this makes it both more consistent and more readable.


aelkiss:
This fixes most of the issues @botimer  and @daaang  reported on #97.

Regarding the index offset for the solr core -- the reason for this is that we expect output like this in `infrastructure.yml`:

```yaml
solr: 
    1: http://localhost:8081/solr/one_core,
    2: http://localhost:8082/solr/second_core
```
However, they are specified in hiera as:

```
   solr_cores:
     one_core:
       host: localhost
       port: 8081
     second_core:
       host: localhost
       port: 8081
```

which was primarily so that we could use `create_resources` with the hash.

On revisiting this, I think just including the `index` parameter explicitly in the hash rather than computing it is more transparent, so I'll make that fix here as well.

aelkiss:
Potential changes for the future:

1) minimize repetition for the JSON stuff
2) factor out mysql stuff

aelkiss:
Need to document how we generate passwords

manifests/profile/moku.pp
```diff
@@ -0,0 +1,19 @@
...
+# BSD License. See LICENSE.txt for details.
+
+# nebula::profile::moku
+#
+# @example
+#   include nebula::profile::moku
+class nebula::profile::moku (
+  String $init_directory = '/etc/moku/init'
+) {
+
```
aelkiss: should be responsible for making sure init_directory exists

manifests/profile/named_instances.pp
```diff
@@ -11,25 +11,21 @@
...
   }
 
-  $defaults = {
-    puma_wrapper    => $puma_wrapper,
-    pubkey          => $fauxpaas_pubkey,
-    puma_config     => $fauxpaas_puma_config,
-    create_database => $create_databases
+  $instances.each |$instance| {
+    Nebula::Named_instance::App <<| title == $instance |>>
+    Nebula::Named_instance::Solr_core <<| instance_title == $instance |>>
```
aelkiss: now just `instance`

spec/defines/named_instance/solr_params_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+# All Rights Reserved. Licensed according to the terms of the Revised
+# BSD License. See LICENSE.txt for details.
+require 'spec_helper'
+
+describe 'nebula::named_instance::solr_params' do
+  on_supported_os.each do |os, os_facts|
+    context "on #{os}" do
+      let(:facts) { os_facts }
+
+      xdescribe 'exported resources'
```
aelkiss: need to fill this in - check that solr core is exported

aelkiss:
> Potential changes for the future:
> 
> 1. minimize repetition for the JSON stuff
> 2. factor out mysql stuff

Created AEIM-1870/AEIM-1871

manifests/profile/moku.pp
aelkiss: I've updated it to ensure that's a directory, but it won't create the parent directory if it doesn't exist. We can probably wait to handle that until we puppetize installing moku itself.

PR #101, opened 2019-02-27
Use a more powerful vim configuration by default

malakai97:
Are you tired of vim having poor tab behavior? Do you hate it when someone commits empty whitespace? Are you confused about where your search results even are? Are you tired of deleting four year old vim swap files? Do you despise the system bell so much that even the idea of it ringing is offputting? Do you hate how when you try to use your own vimrc it doesn't work because the vim features aren't enabled?

If you answered yes to these questions, then do I have the PR for you.

### DIFF
--- a/manifests/profile/vim.pp
+++ b/manifests/profile/vim.pp
@@ -9,7 +9,7 @@
 # @example
 #   include nebula::profile::vim
 class nebula::profile::vim {
-  package { 'vim': }
+  package { ['vim', 'vim-gtk', 'vim-nox': }
 
   file { '/etc/vim/vimrc':
     content => template('nebula/profile/vim/vimrc.vim.erb'),
--- a/templates/profile/vim/vimrc.vim.erb
+++ b/templates/profile/vim/vimrc.vim.erb
@@ -12,42 +12,67 @@
 " properly set to work with the Vim-related packages available in Debian.
 runtime! debian.vim
 
-" Uncomment the next line to make Vim more Vi-compatible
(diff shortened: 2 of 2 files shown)


manifests/profile/vim.pp
```diff
@@ -9,7 +9,7 @@
 # @example
 #   include nebula::profile::vim
 class nebula::profile::vim {
-  package { 'vim': }
+  package { ['vim', 'vim-gtk', 'vim-nox': }
```
aelkiss: vim-gtk is going to bring in X dependencies everywhere, which we probably don't want.

templates/profile/vim/vimrc.vim.erb
```diff
@@ -12,43 +12,68 @@
...
+set shiftwidth=2                " 2 spaces
+set tabstop=2                   " 2 spaces!
+set softtabstop=2               " 2 spaces!!!!
+set autoindent                  " enable automatic indentation
+set smartindent                 " does the right thing
+
+" Disable backups that just clutter git
+set nobackup
+set nowb
+set noswapfile
```
aelkiss: these options might be controversial... I would check in #dev before setting these.

malakai97: I totally forgot that this applied to devs as well.

aelkiss:
This is very stale -- closing for now. If we want to have a discussion at some future point about sane vim defaults with developers, we can do that, but I don't think it serves much point to have this PR open right now.

PR #102, opened 2019-02-27
Include system packages commonly required by ruby

malakai97:
This adds a number of system packages that are commonly required by ruby gems that build native extensions.

These are all system packages I've required at some point or another. Unfortunately, I don't have the mappings of what required them.  At the very least, we should include libsqlite3-dev.


aelkiss:
Given we have to install a full development toolchain everywhere for ruby anyway, I'm fine with including these packages as well.

PR #105, opened 2019-03-05
AEIM-1865: role to run cron jobs

aelkiss:
See top-level comment in manifests/role/cron_runner.pp


manifests/profile/nodejs.pp
```diff
@@ -14,7 +14,7 @@
     apt::source { 'nodesource.com':
       comment       => 'Nodesource apt source for recent nodejs',
       location      => 'https://deb.nodesource.com/node_10.x',
-      release       => $facts['os']['distro']['codename'],
+      release       => $::lsbdistcodename,
```
aelkiss: the `$facts` version does not work in the specs; `$::lsbdistcodename` matches how we're doing this elsewhere in nebula.

PR #111, opened 2019-03-18
Fix travis

daaang:


daaang:
Aaron and I fixed this at the same time without telling each other (although Aaron did at least mention it in our channel where I would have seen it, had I looked). Closing this PR because it's redundant.

PR #112, opened 2019-03-22
Make only one partition on new VMs

rrotter:
Remove our complex partition scheme for VMs and just make one partition. Also eliminates swap.

### DIFF
--- a/templates/virtual_machine/stretch.cfg.erb
+++ b/templates/virtual_machine/stretch.cfg.erb
@@ -165,7 +165,7 @@ d-i clock-setup/ntp boolean true
 # - regular: use the usual partition types for your architecture
 # - lvm:     use LVM to partition the disk
 # - crypto:  use LVM within an encrypted partition
-d-i partman-auto/method string lvm
+d-i partman-auto/method string regular
 
 # If one of the disks that are going to be automatically partitioned
 # contains an old LVM configuration, the user will normally receive a
@@ -188,6 +188,8 @@ d-i partman-lvm/confirm_nooverwrite boolean true
 # just point at it.
 #d-i partman-auto/expert_recipe_file string /hd-media/recipe
 
+# partman fields are: Min Priority Max
+
 # If not, you can put an entire recipe into the preconfiguration file in one
 # (logical) line. This example creates a small /boot partition, suitable
 # swap, and uses the rest of the space for the root partition:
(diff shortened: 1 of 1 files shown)


templates/virtual_machine/stretch.cfg.erb
```diff
@@ -209,40 +209,11 @@ d-i partman-lvm/confirm_nooverwrite boolean true
 #              .
 d-i partman-auto/expert_recipe string                 \
     partman-auto/text/pixie ::                        \
-        2048 100 2048 ext4                            \
+        2048 100 1000000000 ext4                      \
```
aelkiss: What are the magic numbers here? They differ from the previous magic numbers from the largest "partition":

```
1 1000000000 1000000000
```

I'm happy to at least try running entirely without swap - realistically if anything starts swapping we're sunk anyway; we should allocate enough physical RAM to cover normal needs and if there are oom_killer invocations or crashes because of memory leaks or other issues we deserve what we get.

rrotter:
That's what I was thinking on swap. Googling the reason you need swap is "what if ram fills up?", but if we fill up ram the only reason we won't fill swap too is that the system would grind to a crawl, might as well fail faster.

As for the magic numbers:
min priority max
1000000000 == infinity

2048 100 2048 == make it exactly 2GB at the default priority
1 1000000000 1000000000 == make it between infinitely small and infinitely large, but give it no priority. This was junk to fill up the partition table so the installer didn't complain.
2048 100 1000000000 == fail if we didn't allocate at least 2GB (our base Debian install is 1.2GB, so this is close to a bare minimum), fill up the disk, and do this at the default priority.

templates/virtual_machine/stretch.cfg.erb
rrotter: That's what I was thinking on swap. Googling the reason you need swap is "what if ram fills up?", but if we fill up ram the only reason we won't fill swap too is that the system would grind to a crawl, might as well fail faster.

As for the magic numbers:
min priority max
1000000000 == infinity

2048 100 2048 == make it exactly 2GB at the default priority
1 1000000000 1000000000 == make it between infinitely small and infinitely large, but give it no priority. This was junk to fill up the partition table so the installer didn't complain.
2048 100 1000000000 == fail if we didn't allocate at least 2GB (our base Debian install is 1.2GB, so this is close to a bare minimum), fill up the disk, and do this at the default priority.

aelkiss: OK, cool. Probably worth a comment in the actual template.

rrotter: Fair enough, added comments clarifying magic numbers and rationale for swap removal.

PR #113, opened 2019-03-25
HTREPO-211 - Raise the resource limits for ImageMagick

daaang:


daaang:
I checked the values on macc-ht-ingest-001, and ImageMagick stopped throwing errors on one of the problem images, so I'm pretty sure this should be sufficient.

PR #115, opened 2019-04-01
AEIM-1932: symlink physical core location to solr home

aelkiss:
In production, the cores are at e.g. /somewhere/solr/cores, but they
need to be symlinked to the solr home under /var/lib/solr-whatever/home
in order to be picked up at restart.

This adds a separate parameter for the "core home" vs. the "solr home"
and creates that symlink. It also adds some documentation covering how
the solr core defined type works.


aelkiss:
The host/port/solr home is now kind of a data clump that we should extract as a hash that we can default per host and/or pass as an alias from hiera rather than needing to specify it for every core, but I think that's out of scope for this PR.

aelkiss:
This also removes the unused 'index' parameter from the solr core defined type.

There is a corresponding lensoftruth branch.

aelkiss:
Please wait to merge this until I have tested on nectar.

aelkiss:
Need to revert the removal of the index parameter - it wasn't used here but it is used in the infrastructure config for the solr core.

aelkiss:
Tested on nectar without issue; this may be merged at any time.

skorner:
The functional changes look right to me.

PR #117, opened 2019-04-10
AEIM-1277 - Add two_syncs_ago_summary custom fact

daaang:


malakai97:
* Add a top level comment briefly saying what it does (less important) and why. 
* Place the filepath in a constant.
* Optionally, there's a method YAML.load_path that you can use if that's easier.

daaang:
The failing test is failing in master, but I'm not sure why—either way it's not related to this pull request.

PR #119, opened 2019-04-12
AEIM-2028 - Add architecture to the local apt source

daaang:
It looks like the issue with our local apt repo is that it doesn't have any i386 packages, and for some reason apt expects those to exist even when run on an amd64 computer? Adding `[arch=amd64]` to the sources.list does fix the problem, and this is the code that should add that to the sources.list, although I had to look at the apt module source to be sure, and it's possible we'll need to upgrade to a newer apt module version, but this should prevent apt from trying to grab i386 binaries.

### DIFF
--- a/manifests/profile/apt.pp
+++ b/manifests/profile/apt.pp
@@ -60,9 +60,10 @@ class nebula::profile::apt (
 
     if $local_repo {
       apt::source { 'local':
-        *       => $local_repo,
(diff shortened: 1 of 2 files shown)


daaang:
I just looked through their version history, and we shouldn't need an upgrade. The architecture parameter existed in version 4.5.1, which is the one we name as our minimum.

PR #121, opened 2019-04-22
AEIM-2055 - Don't create separate directory for app

botimer:
When a named_instance::app should be deployed to the app user's home
directory, do not create a separate file resource.

### DIFF
--- a/manifests/named_instance/app.pp
+++ b/manifests/named_instance/app.pp
@@ -73,7 +73,7 @@ define nebula::named_instance::app (
   # Create the application user's home directory
   file { "/var/local/${title}":
     ensure => 'directory',
-    mode   => '0700',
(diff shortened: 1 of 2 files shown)


botimer:
That's a good point. Maybe we should just decide here and now whether app dirs should be readable outside the group. I think our purpose in the 2775 was to account for other users in the group and and Apache reading from the app dir. With the assets split off, maybe 0770 is right for both.

aelkiss:
0750 sounds good. I don't think we should need the setgid stuff, since nobody but deployment should be writing there.

botimer:
Because there is a close relationship between the mode and serving the static assets, we have decided to make any mode changes with AEIM-2044, rather than here. This change is just to avoid the duplicate resource problem.

PR #122, opened 2019-04-23
AEIM-2055 - Second pass to fix directory modes

botimer:
This addresses a problem with the home directory being group writeable.
It also goes back to the notion that we can remove setgid and group
write for the rest of the dirs. It did not end up simplifying to
preserve the 2775.


malakai97:
We should use 0750 rather than 750, as puppet will set the setgid bit to 0 if it is omitted. This makes that more clear. See https://puppet.com/docs/puppet/5.3/types/file.html#file-attribute-mode

PR #123, opened 2019-04-26
AEIM-1928: cron jobs for atlassian backups

aelkiss:
Adds some documentation to the tools_lib profiles as well.

Moves the includes for postgres & jdk to the confluence & jira profiles,
so that we can test those profiles without the need for a preamble.

### DIFF
--- a/files/pg_backup
+++ b/files/pg_backup
@@ -1,0 +1,27 @@
+#! /bin/sh
+
+# backup postgres database using the 'custom' archive format
+# which has default compression and allows for the most flexible recovery
+# using pg_restore
+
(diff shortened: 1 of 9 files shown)


aelkiss:
Needs a `MAILTO` envvar set for the cron jobs

aelkiss:
Added the `MAILTO` env var. We might want to consider at some point having a global nebula setting in hiera for cron job recipients (that can be overridden on a profile by profile basis). We could consider adding that now if this seems too ugly.

skorner:
The functional additions look good to me.

PR #124, opened 2019-04-29
HT solr LSS configuration

aelkiss:
This could probably stand to be squashed.

One question - should the entire core dir be symlinked to the provided path (but still have puppet add the conf & lib dirs) as currently, or should only the data dir be symlinked? My feeling is that if puppet is managing conf & lib probably only the data dir should be symlinked.


skorner:
I agree with symlinking at the data directory, given puppet is managing conf and lib.

The functional implementation looks good to me.

aelkiss:
I'll push a fix for that.

aelkiss:
I believe this is all working. This should be merged to master and then we can test it on a clean VM.

PR #125, opened 2019-04-30
Add infrastructure.instance_name to named_instance json

malakai97:
Resolves AEIM-2059


malakai97:
Rubocop forced me to misalign that text.

PR #126, opened 2019-05-08
Allow hiera to control puppetdb threads

daaang:


daaang:
This won't solve the reporting problem, so never mind.

PR #128, opened 2019-05-10
kubernetes

daaang:
Not even a little working yet, but I want Travis to keep an eye on this.

### DIFF
--- a/doc/adr/0004-roll-our-own-kubernetes-profile.md
+++ b/doc/adr/0004-roll-our-own-kubernetes-profile.md
@@ -1,0 +1,53 @@
+# 4. Roll our own kubernetes profiles
+
+Date: 2019-07-25
+
+Status
+------
+
+Accepted
+
+Context
+-------
+
+Puppetlabs maintains [a kubernetes module][1] in the forge, and Rancher
+maintains software for managing multiple clusters running anywhere.
+
+The puppetlabs module claims to support Debian, but, in practice, it
+appears to only support Ubuntu. We tried forking it and making changes
+to get it working, but it was a rabbithole of error after error. Also,
+their solution to managing SSL keys was to generate them locally and
+store them all in hiera, which was cumbersome. They also didn't support
+managing the CIDRs for the internal network.
+
+Rancher seemed very nice at first, but it was unstable in practice.
+Despite running a highly available set of control nodes, bringing one
+down did in fact break the cluster. Also, if /var/lib/docker filled up
+on any machine, the only solution was to destroy and recreate the entire
+cluster.
+
+My assessment of the puppetlabs module is that it tries to do too much
+on its own. My assessment of rancher is that it's not yet stable enough
+to rely on.
+
+[1]: https://forge.puppet.com/puppetlabs/kubernetes
+
+Decision
+--------
+
+We will roll our own kubernetes profiles instead of relying on someone
+else's solution.
+
+Consequences
+------------
+
+While our profiles are sufficient for creating an environment where
+kubernetes can flourish, bootstrapping must be done outside of puppet,
+by hand. It's not ideal, but it also shouldn't come up very often.
+
+We haven't yet tried upgrading docker or kubernetes on an existing
+cluster, but it's easy to imagine that being tricky if we have to do it
+by hand.
+
+It's worth checking back on rancher in a year or so to see if they've
+improved.
--- a/doc/adr/0007-only-manage-certain-netfilter-chains-in-kubernetes-machines.md
+++ b/doc/adr/0007-only-manage-certain-netfilter-chains-in-kubernetes-machines.md
@@ -1,0 +1,67 @@
+# 7. Only manage certain netfilter chains in kubernetes machines
+
+Date: 2019-07-25
+
+Status
+------
+
+Accepted
+
+Context
+-------
+
+When using the puppetlabs firewall module, one of the recommended
+settings is to purge all rules not managed by puppet:
+
+```puppet
+resources { 'firewall':
+  purge => true,
+}
+```
+
+Since kubernetes and calico perform networking and load balancing at the
+kernel level, this means puppet finds itself with hundreds of rules to
+purge every 30 minutes.
+
+The firewall module also has an undocumented feature that lets you
+manage individual netfilter chains:
+
+```puppet
+# This will remove all `-A INPUT` rules unless they're either (a)
+# managed by puppet or (b) contain `--comment "you can trust me ;)"`.
+firewallchain { 'INPUT:filter:IPv4':
+  ensure => 'present',
+  purge  => true,
+  ignore => [
+    '--comment "you can trust me ;)"',
+  ],
+}
+```
+
+This doesn't work in conjunction with `purge => true` for all firewall
+resources. If we use `firewallchain` resources instead, then any chains
+we don't explicitly define will be ignored by puppet. So, for example,
+if we define only `INPUT:filter:IPv4`, then any existing `OUTPUT` rules
+will be ignored rather than purged by puppet.
+
+Decision
+--------
+
+We will only manage INPUT, OUTPUT, and FORWARD chains on kubernetes
+machines. We will specifically whitelist particular lines known to be
+used in those chains by kubernetes and calico.
+
+Consequences
+------------
+
+As of writing this, it means we now have two firewall profiles
+(`networking::firewall` and `kubernetes::firewall`), and they share most
+of their code. We should probably split that `purge => true` line out of
+the central firewall profile and come up with a single structure that
+works more broadly.
+
+Additionally, docker itself performs its own networking, so we'll want
+to do something like this for the `docker` profile, but with different
+exceptions in the chains we manage.
+
+Security consequences: unknown.
--- a/doc/adr/0008-dedicate-separate-haproxy-machines-for-load-balancing-kubernetes.md
+++ b/doc/adr/0008-dedicate-separate-haproxy-machines-for-load-balancing-kubernetes.md
@@ -1,0 +1,60 @@
+# 8. Dedicate separate HAProxy machines for load-balancing kubernetes
+
+Date: 2019-07-25
+
+Status
+------
+
+Accepted
+
+Context
+-------
+
+With three controller nodes all listening on 6443, they need a single IP
+address they can use to represent "the kubernetes master." Additionally,
+each worker node listens on 30000-32767 in order to expose NodePort
+services to the outside world. Rather than choose a particular worker
+node to talk to, it makes sense to load-balance these requests as well
+to whichever worker nodes are currently online.
+
+What's written above is straightforward enough to add to our current
+HAProxy config.
+
+However, kubernetes doesn't provide any support for claiming IP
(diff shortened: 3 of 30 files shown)


jdbohrman:
What are you hoping to achieve with Kubernetes here? Just wondering, I'd love to help but I'm curious what the goal is.

doc/adr/0007-only-manage-certain-netfilter-chains-in-kubernetes-machines.md
```diff
@@ -0,0 +1,67 @@
...
+(`networking::firewall` and `kubernetes::firewall`), and they share most
+of their code. We should probably split that `purge => true` line out of
+the central firewall profile and come up with a single structure that
+works more broadly.
+
+Additionally, docker itself performs its own networking, so we'll want
+to do something like this for the `docker` profile, but with different
+exceptions in the chains we manage.
+
+Security consequences: unknown.
```
daaang: If anyone has an understanding of how netfilter tables work, I'd really like you to expend on possible security consequences.

manifests/profile/kubernetes/firewall.pp
```diff
@@ -0,0 +1,69 @@
+# Copyright (c) 2019 The Regents of the University of Michigan.
+# All Rights Reserved. Licensed according to the terms of the Revised
+# BSD License. See LICENSE.txt for details.
+
+# Kubernetes firewall settings
+#
+# This is exactly the same as networking::firewall except that it only
+# manages three chains, and it ignores some rules that we're electing to
+# not have puppet manage.
+class nebula::profile::kubernetes::firewall {
```
daaang: This should be combined somehow into `networking::firewall`, but I'm not sure what the best way is to do that. I haven't given it much thought because, until this moment, this was a WIP branch.

manifests/profile/kubernetes/stacked_controller.pp
```diff
@@ -0,0 +1,112 @@
...
+
+  # Controller nodes (stacked with etcd) accept connections from each
+  # other for etcd.
+  Firewall <<| tag == "${cluster_name}_etcd" |>>
+
+  # Controller nodes accept connections from each other for kubelet.
+  Firewall <<| tag == "${cluster_name}_kubelet" |>>
+
+  # Controller nodes should listen for kubernetes connections.
+  Firewall <<| tag == "haproxy-kubectl-${cluster_name}" |>>
```
daaang: I noticed this inconsistent tag naming, and I'm going to fix that before I merge this. Since there's already a cluster up and running and being used, it'll be delicate and involve some temporary duplicate rules, so try to pay that no mind.

manifests/role/kubernetes/controller.pp
```diff
@@ -0,0 +1,14 @@
...
+# BSD License. See LICENSE.txt for details.
+
+# Kubernetes controller plane and etcd server
+class nebula::role::kubernetes::controller {
+  include nebula::profile::base
+  include nebula::profile::work_around_puppet_bugs
+  include nebula::profile::apt
+  include nebula::profile::authorized_keys
+  include nebula::profile::vim
+  include nebula::profile::kubernetes::firewall
```
daaang: All these lines should just be `include nebula::role::minimal`, but I can't do that until the base firewall supports ignoring some chains.

doc/adr/0008-dedicate-separate-haproxy-machines-for-load-balancing-kubernetes.md
```diff
@@ -0,0 +1,60 @@
...
+configure them, but it also provides a simple list of peer IP addresses
+so that they will one day be capable of configuring themselves.
+
+Consequences
+------------
+
+We can make changes to the load balancers for kubernetes without
+worrying about breaking access to our first-class services already in
+production. Even though the details of these servers are very similar,
+this separation let's us view these HAProxy machines not as load
```
botimer: lets

doc/adr/0004-roll-our-own-kubernetes-profile.md
```diff
@@ -0,0 +1,53 @@
...
+--------
+
+We will roll our own kubernetes profiles instead of relying on someone
+else's solution.
+
+Consequences
+------------
+
+While our profiles are sufficient for creating an environment where
+kubernetes can flourish, bootstrapping must be done outside of puppet,
```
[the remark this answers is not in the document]
daaang: If we spin up new kubernetes nodes and set up the hieradata so that puppet sets them up, then, after an hour or so, they'll be ready to be a cluster (the right ports will be open to the right machines, all the needed configuration will be in the right place), but kubernetes itself won't be running. There are a couple commands that one of us will have to run on each machine in the cluster and in a particular order, and we'll have to go through that once per cluster.

So to answer your question, "bootstrapping" is going from "puppet's done configuring these machines to be kubernetes" to "kubernetes is running on these machines." It needs to happen once every time a new cluster is created, and I'm currently punting on whether/how that process should be automated.

It could in theory be automated by puppet (it's what their module does), but I'd rather avoid doing that here, mostly because it feels like it'd be a lot of work (getting puppet to pick one machine to run one command on, look at the output, and then run other commands on other machines based on that output, eeurggh, _and_ never run them ever again on those machines) just to avoid a little repetitive work on our part.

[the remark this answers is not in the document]
daaang: I should add that the bootstrapping process isn't documented (in part because I'm punting on it), but we do have [a ticket for that in jira](https://tools.lib.umich.edu/jira/browse/AEIM-2176) that at least acknowledges that I haven't done this yet.

manifests/profile/docker.pp
```diff
@@ -66,7 +66,8 @@
     include docker
   } else {
     class { 'docker':
-      version => $version,
+      version          => $version,
+      extra_parameters => ['--insecure-registry=hatcher-kubernetes.umdl.umich.edu:32030'],
```
daaang: This is a hack to let us have a private registry immediately. Longer-term, if we want a private registry, it should at the very least run over HTTPS so that we don't need this line. The one that's up is minimal and only exists so that we can have an insightiq pod today.

manifests/profile/networking/firewall.pp
```diff
@@ -21,8 +24,74 @@
...
+
+        $output_ignore = [
+          '-j DOCKER',
+        ]
+
+        $forward_ignore = [
+          '-j DOCKER-USER',
+          '-j DOCKER-ISOLATION',
+          '-i docker0',
+          '-o docker0',
```
daaang: Docker has the same issue (but different rules) when being run on its own. I don't know how much we'll be moving forward with docker-without-kubernetes, but I figured I may as well express this here.

manifests/profile/networking/firewall.pp
```diff
@@ -21,8 +24,74 @@
...
+          '-j cali-FORWARD',
+          '-j KUBE-FORWARD',
+          '-j KUBE-SERVICES',
+        ]
+      }
+
+      default: {
+        $input_ignore = []
+        $output_ignore = []
+        $forward_ignore = []
```
daaang: In this default case, we'll only manage the INPUT, OUTPUT, and FILTER chains, and we'll purge everything on them that we didn't put there. Seems like a pretty safe default to me?

doc/adr/0004-roll-our-own-kubernetes-profile.md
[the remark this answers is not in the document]
botimer: Thanks for the detail -- all sounds good.

doc/adr/0007-only-manage-certain-netfilter-chains-in-kubernetes-machines.md
aelkiss: I think this is pretty reasonable. You have to be root to manage these things anyway, and if you're root, you can just turn puppet off to prevent it from purging. IMO the reason to purge is more so that authorized root users can't just add arbitrary stuff everywhere and then forget about it, than for any security reason.

spec/classes/profile/kubernetes_spec.rb
```diff
@@ -0,0 +1,118 @@
...
+        it do
+          is_expected.to contain_class('nebula::profile::docker').with(
+            version: '5:18.09.6~3-0~debian-stretch',
+          )
+        end
+
+        it do
+          is_expected.to contain_apt__source('kubernetes').with(
+            location: 'https://apt.kubernetes.io/',
+            release: 'kubernetes-xenial',
```
aelkiss: Why xenial, given that kubernetes-stretch appears to exist there?

aelkiss: Also, I think we're now leaning towards trying to minimize these tests that just duplicate static data from the profiles.

manifests/profile/kubernetes/haproxy.pp
```diff
@@ -0,0 +1,162 @@
...
+# This is for the gateway to a kubernetes cluster. In kubernetes terms,
+# this exists outside the actual cluster, but, since we need to be able
+# to access the cluster, this role is essential.
+#
+# @param master Exactly one gateway per cluster should be designated as
+#   the master so that any others will simply exist as failover nodes.
+# @param cluster The name of the cluster to serve as ambassador to. This
+#   defaults to nebula::profile::kubernetes::cluster, and you should
+#   almost certainly set that instead.
+class nebula::profile::kubernetes::haproxy (
```
aelkiss: It might be worth looking into the puppetlabs haproxy module, but this is certainly OK for now.

manifests/profile/networking/firewall.pp
aelkiss: Agreed.

manifests/profile/kubernetes/haproxy.pp
daaang: Could do, but the long-term plan is to have these manage their own configs, rather than have puppet manage them. Once puppet isn't doing that anymore, I doubt there'd be much an HAProxy module can do for us.

spec/classes/profile/kubernetes_spec.rb
daaang: That's a good question, and I think the answer is that I couldn't get stretch to work, although that was months ago, and maybe it just didn't exist at the time. I'll see what happens if I change it by hand on one of them, see if anything breaks.

daaang: Confirmed, stretch doesn't actually offer kubelet or kubeadm (it only offers kubectl, which we don't need or want on the servers):

```
root@hatcher-k8s-work-002:~# apt-cache madison kubeadm
root@hatcher-k8s-work-002:~# apt-cache madison kubelet
root@hatcher-k8s-work-002:~# apt-cache madison kubectl
   kubectl |   1.6.0-00 | https://apt.kubernetes.io kubernetes-stretch/main amd64 Packages
   kubectl |   1.5.6-00 | https://apt.kubernetes.io kubernetes-stretch/main amd64 Packages
```

When I switch back to xenial, there are too many options to reasonably display here:

```
root@hatcher-k8s-work-002:~# apt-cache madison kubeadm | wc -l
110
root@hatcher-k8s-work-002:~# apt-cache madison kubelet | wc -l
151
```

PR #132, opened 2019-05-24
AEIM-2097: sort the results of nodes_for_class

aelkiss:
This should keep anything that relies on the output of nodes_for_class
from seeing arbitrary changes in catalogs based on ordering.


aelkiss:
Please wait until Tuesday to merge this.

PR #133, opened 2019-05-28
AEIM-843: remove rbenv from syslog

aelkiss:


aelkiss:
Tested on malt, looks fine

PR #135, opened 2019-06-03
AEIM-2116 - Open 6443 on haproxy servers

daaang:
Not the best way to do it maybe, but this will get it done, and it'll be easy enough to revisit once we have something working.

### DIFF
--- a/manifests/profile/haproxy.pp
+++ b/manifests/profile/haproxy.pp
@@ -114,4 +114,7 @@ class nebula::profile::haproxy(
     tag    => 'haproxy'
   }
 
+  # HAProxy should listen for kubernetes connections.
+  Firewall <| tag == 'listen_for_kubectl' |>
+
 }
--- a/manifests/profile/networking/firewall/ssh.pp
+++ b/manifests/profile/networking/firewall/ssh.pp
@@ -20,5 +20,14 @@ class nebula::profile::networking::firewall::ssh (
       state  => 'NEW',
       action => 'accept',
     }
+
+    @firewall { "200 kubectl: ${network['name']}":
+      tag    => 'listen_for_kubectl',
+      proto  => 'tcp',
+      dport  => 6443,
+      source => $network['block'],
+      state  => 'NEW',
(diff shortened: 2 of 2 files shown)


manifests/profile/vmhost/prereqs.pp
```diff
@@ -34,7 +34,7 @@
   package { 'qemu':
     ensure => 'installed'
   }
-  
+
   package { 'qemu-kvm':
```
daaang: This was just to make the linter happy.

manifests/profile/haproxy.pp
```diff
@@ -114,4 +114,6 @@
     tag    => 'haproxy'
   }
 
+  nebula::haproxy::kubernetes_cluster { 'just open port 6443 for now': }
```
aelkiss: Why not just put the firewall rules right here?

daaang: My thinking was that we'd need a defined type eventually, so I just went ahead and made one—but looking back I think that for now it'd be better to do this the simpler way and just put the firewall rules here, yeah.

aelkiss:
What's the motivation for using virtual resources here? It seems confusing to have the firewall rules for kubernetes in a profile called 'ssh.pp' -- why can't we just put the firewall resource directly in `manifests/profile/haproxy.pp`?

daaang:
My motivation for putting it there was to have `nebula::profile::networking::firewall::ssh` be the only thing that knows about `nebula::profile::networking::firewall::ssh::networks`, rather than have the haproxy profile loop through `lookup('nebula::profile::networking::firewall::ssh::networks')`. This way, the ssh profile still bears the responsibility of knowing which CIDRs might want access, and then it's up to haproxy and kubernetes to decide to actually listen.

There's probably a better way to do it, but I think it'll necessitate structural changes to the way we open ports for ssh access as well, and that feels out of scope to me.

aelkiss:
I see - let's go with this for now and revisit later for the firewall stuff.

PR #136, opened 2019-06-03
Add nebula::profile::docker

daaang:


manifests/profile/vmhost/prereqs.pp
```diff
@@ -34,7 +34,7 @@
   package { 'qemu':
     ensure => 'installed'
   }
-
```
daaang: This is just to make the linter happy.

manifests/profile/vmhost/prereqs.pp
```diff
@@ -34,7 +34,7 @@
   package { 'qemu':
     ensure => 'installed'
   }
-  
+
   package { 'qemu-kvm':
```
daaang: This is just to appease the linter; it doesn't otherwise relate to this pull request.

PR #138, opened 2019-06-10
AEIM-1540: install ssacli

aelkiss:
moves the hpwdt blacklist to a new profile for all HP hardware
customizations, which includes ssacli as well


aelkiss:
branch was named wrong, see #139

PR #144, opened 2019-07-02
PFDR-178: get chipmunk role working

aelkiss:
adds default parameters for geoip username & license key so it can be
installed but (effectively) unconfigured. the alternative would be
setting it in the specs in hiera, but it makes no sense to force a hiera
configuration for a component we don't even use for chipmunk. longer
term it would be better to include only the dependencies we actually
need for feed.

### DIFF
--- a/manifests/profile/geoip.pp
+++ b/manifests/profile/geoip.pp
@@ -17,8 +17,8 @@
 #     user_id        => '12345
 #   }
 class nebula::profile::geoip (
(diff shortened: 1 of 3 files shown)


aelkiss:
Also uses ensure_packages for the hathitrust dependencies per slack convo - in particular the rbenv profile includes git, which clashes with this (we didn't yet install ruby on the new ht ingest machines)

PR #146, opened 2019-07-06
AEIM-2130 - Create exposed_port resource type

daaang:


manifests/profile/networking/firewall/ssh.pp
```diff
@@ -8,26 +8,9 @@
...
-      tag    => 'listen_for_kubectl',
-      proto  => 'tcp',
-      dport  => 6443,
-      source => $network['block'],
-      state  => 'NEW',
-      action => 'accept',
-    }
+class nebula::profile::networking::firewall::ssh {
+  nebula::exposed_port { '100 SSH':
+    port  => 22,
```
daaang: Since this could potentially break ssh for all our servers, this should be tested in an environment before being merged.

daaang: I tested it on hatcher-k8s-000, and it didn't even affect the firewall (this ended up generating identical resources with identical titles). That's good enough for me if it's good enough for whoever reviews this.

PR #153, opened 2019-07-16
Add mysql backup

malakai97:
This adds the mysql backup as implemented in the mysql_common ansible
role.

### DIFF
--- a/manifests/profile/mysql.pp
+++ b/manifests/profile/mysql.pp
@@ -6,10 +6,13 @@
 #
 # @param password The root password, as stored by root's .my.cnf.
 class nebula::profile::mysql (
-  String $password
+  String $password,
+  String $datadir = "/var/lib/mysql"
 ) {
 
+  #
   # Install and configure mysql server
+  #
   class { 'mysql::server':
     create_root_user        => true,
     create_root_my_cnf      => true,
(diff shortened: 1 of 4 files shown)


aelkiss:
I cleaned this up a little bit -- in particular I exposed a parameter to make the backup directory configurable independent of the data dir -- if travis passes I'm happy merging it.

I think before closing the parent issue we probably want to:

1) ensure that the data dir parameter is actionable - that is it gets set in /etc/mysql/whatever and that any default stuff is moved there

2) review our usual mysql settings; if they are configurable via hiera for mysql::server then we can just set those in LoT. Otherwise we'll need to figure out how to drop those into my.cnf on the server - I think there is a .d setup now, so a drop-in there would probably be the easiest thing

malakai97:
There's a passthrough in the mysql::server module for extra options I believe. I'm not sure if we'd need to expose that on nebula::profile::mysql, or if hiera can set things on mysql::server::blah directly.

botimer:
This is very stale. Is there anything of current value left in this PR?

skorner:
It does capture the current state of our mysql backup process and allow puppet-based management of mysql backups using our current process. That made sense two years ago when our goal was trying to reproduce all current processes using puppet, but it's perhaps not so helpful anymore. Just my two cents.

PR #155, opened 2019-07-16
AEIM-1699 - All roles compile

daaang:


daaang:
I think the route I've gone is going to leave spectacularly unhelpful error messages whenever a role doesn't compile, but the only alternative I could think of would be to add valid hiera representing each of these to default.yaml, but, in those cases, failure to compile in production is a feature, so I didn't want to add a bunch of broken defaults.

I might want to think if we can come up with at least a message better than `should compile into a catalogue without dependency cycles`, perhaps that suggests a condition needs to be added to all_roles_spec.rb? Or maybe just add an error handler for that. I'm gonna wait to see if Travis passes, and then add that if it does.

manifests/profile/aws/filesystem.pp
```diff
@@ -10,24 +10,26 @@
 #   include nebula::profile::aws::filesystem
 class nebula::profile::aws::filesystem {
 
-  if $facts['disks']['xvdb'] {
-    filesystem { '/dev/xvdb':
-      ensure  => present,
-      fs_type => 'ext4',
-    }
+  if $facts['disks'] {
+    if $facts['disks']['xvdb'] {
```
daaang: Was getting an error that `undef` doesn't respond to `[]`. Checking that `$facts['disks']` is defined fixes it, but there may be a cleaner fix.

manifests/profile/mysql.pp
```diff
@@ -1,7 +1,7 @@
 # Copyright (c) 2018 The Regents of the University of Michigan.
 # All Rights Reserved. Licensed according to the terms of the Revised
 # BSD License. See LICENSE.txt for details.
-#
+
```
daaang: Cosmetic—noticed this while looking at the file.

spec/classes/all_roles_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+          elsif role_name.start_with? 'nebula::role::deploy_host'
+            'spec/fixtures/hiera/named_instances_config.yaml'
+          else
+            'spec/fixtures/hiera/default_config.yaml'
+          end
+        end
+
+        let!(:puppetdb_query) do
+          MockFunction.new('puppetdb_query') do |f|
+            f.stubbed.returns([])
```
daaang: A number of roles use `puppetdb_query`, but, luckily, they all seem to be happy with `[]`, which, honestly, they probably always should be.

spec/default_facts.yml
```diff
@@ -17,3 +17,5 @@ augeasversion: '1.4.0'
 
 # required by apt with update => { frequency => 'daily' }
 apt_update_last_success: '1523250000'
+
+ec2_tag_role: nebula::role::aws
```
daaang: This is for the sake of `nebula::role::aws::auto`

spec/classes/all_roles_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+        end
+
+        it do
+          begin
+            is_expected.to compile
+          rescue RSpec::Expectations::ExpectationNotMetError => e
+            # Append a message with a suggestion of how to fix the role.
+            x = RSpec::Expectations::ExpectationNotMetError.new(
+              "You probably need to add a hiera_config to spec/classes/all_roles_spec.rb:\n #{e.message}",
+            )
```
daaang: This is a bit ugly, but at least it'll point to where the error is happening, hopefully with some clarity.

spec/classes/all_roles_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+          elsif role_name.start_with? 'nebula::role::tools_lib'
+            'spec/fixtures/hiera/tools_lib_config.yaml'
+          elsif role_name.start_with? 'nebula::role::chipmunk'
+            'spec/fixtures/hiera/chipmunk_config.yaml'
+          elsif role_name.start_with? 'nebula::role::deb_server'
+            'spec/fixtures/hiera/deb_server_config.yaml'
+          elsif role_name.start_with? 'nebula::role::deploy_host'
+            'spec/fixtures/hiera/named_instances_config.yaml'
+          else
+            'spec/fixtures/hiera/default_config.yaml'
```
daaang: I feel like this could be better represented, like possibly as a separate configuration file, or maybe at least as an array of duples. I'm hesitant to put it in yaml, but also it might be nice anyway to have a yaml file that shows which roles require particular hieradata to exist.

Thoughts?

manifests/profile/aws/filesystem.pp
aelkiss: `if $facts['disks'] and $facts['disks']['xvdb']`?

Alternatively, we could set a default `$facts['disks']` fact - we should be able to rely on that being present everywhere, and it's just an artifact of the tests that it isn't.

spec/classes/all_roles_spec.rb
aelkiss: The simplest thing would probably be just stuffing everything in an 'all roles' yaml, but then we'd have to maintain that list in multiple places, and it wouldn't be very helpful to look at on its own. Maybe extract to some kind of spec helper function that loads the correct hiera for a given role? Would that be useful to use in the individual role specs as well?

spec/classes/all_roles_spec.rb
aelkiss: Seems reasonable. I think in general I'm trying to avoid adding new uses of `puppetdb_query`, and think about how to refactor existing uses to avoid it. There are a few places where it's difficult to avoid, though.

spec/classes/all_roles_spec.rb
aelkiss: We could do this in perhaps a little bit cleaner way by subclassing the `compile` matcher (https://github.com/rodjek/rspec-puppet/blob/master/lib/rspec-puppet/matchers/compile.rb) and overriding `failure_message` and `failure_message_when_negated`

daaang: I think this is way better.

manifests/profile/aws/filesystem.pp
daaang: Done.

spec/classes/all_roles_spec.rb
daaang: Here's what I did:

```ruby
          hiera_fixture = [
            %w[nebula::role::webhost::htvm hathitrust],
            %w[nebula::role::hathitrust hathitrust],
            #...
            %w[nebula default],
          ].select { |base, _| role_name.start_with? base }.first[1]
```

Can you think of a better ruby way than that last line to say "grab the first item in an array where X is true"?

aelkiss: I believe `find` does exactly that: https://ruby-doc.org/core-2.6.3/Enumerable.html#method-i-find

aelkiss: Well, you would still need to append the `[1]` to it, but you wouldn't need the `first`.

PR #156, opened 2019-07-16
AEIM-2180: set GSSAPIAuthenticationYes only if keytab file exists

aelkiss:
On Ubuntu, there are ssh errors with the GSSAPI authentication method if
the keytab file isn't actually there, and in general advertising that
method only leads to confusion if it doesn't actually work.

This change:

- always includes the keytab profile (it is a no-op if there is no
actual keytab file to install)

- only sets GSSAPIAuthentication Yes if the keytab profile installed a
keytab

### DIFF
--- a/manifests/profile/networking.pp
+++ b/manifests/profile/networking.pp
@@ -15,23 +15,13 @@
 
 class nebula::profile::networking (
   Boolean $bridge = false,
-  Boolean $keytab = false,
 ) {
 
   class { 'nebula::profile::networking::sysctl':
     bridge => $bridge,
   }
 
-  if $keytab {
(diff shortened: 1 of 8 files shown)


manifests/profile/networking.pp
```diff
@@ -15,23 +15,13 @@
...
-    include nebula::profile::networking::keytab
-    class { 'nebula::profile::networking::sshd':
-      gssapi_auth => true,
-    }
-  } else {
-    class { 'nebula::profile::networking::sshd':
-      gssapi_auth => false,
-    }
-  }
+  include nebula::profile::networking::sshd
```
daaang: I like giving this logic to the sshd profile.

PR #158, opened 2019-07-22
Pull puppet5 repo instead of PC1 for new VMs

daaang:


manifests/role/aws.pp
```diff
@@ -9,7 +9,12 @@
 class nebula::role::aws {
 
   include nebula::role::minimum
-  package { 'iptables-persistent': }
+
+  ########################################
+  # this needs to be in a profile
+  ensure_packages(['iptables-persistent'])
+  ########################################
```
daaang: I put this note here instead of fixing it because failing master is an emergency, and I'm getting to green as fast as I can.

PR #160, opened 2019-07-29
Prometheus plus node exporters

daaang:
This will affect every non-jessie node we have (that's being managed by puppet at least). This will run `apt-get install prometheus-node-exporter` and will open a port to the Prometheus server at its datacenter (if there is one).

Other than that, the purpose here is to set us up with a Prometheus server at each datacenter, with every node giving it basic hardware monitoring data.

Out of scope is setting up the alert manager and figuring out how best to aggregate everything into such as a pretty dashboard. The alert manager and dashboards are pieces of this that I think would fit well in kubernetes (as well as monitoring other details such as Apache)—but the basic hardware monitoring is definitely easiest to set up through puppet.

### DIFF
--- a/manifests/profile/prometheus/exporter/node.pp
+++ b/manifests/profile/prometheus/exporter/node.pp
@@ -6,16 +6,6 @@ class nebula::profile::prometheus::exporter::node (
   Array $covered_datacenters = [],
   String $default_datacenter = 'default',
 ) {
-  case lookup('role') {
-    'nebula::role::aws::auto': {
-      $role = pick($::ec2_tag_role, 'nebula::role::aws::auto')
-    }
-
-    default: {
-      $role = lookup('role')
-    }
-  }
-
   file { '/etc/default/prometheus-node-exporter':
     content => template('nebula/profile/prometheus/exporter/node.sh.erb'),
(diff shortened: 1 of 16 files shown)


daaang:
This is working in its own environment (I chose ae-files as the guinea pig).

manifests/profile/prometheus/exporter/node.pp
```diff
@@ -0,0 +1,41 @@
+# Copyright (c) 2019 The Regents of the University of Michigan.
+# All Rights Reserved. Licensed according to the terms of the Revised
+# BSD License. See LICENSE.txt for details.
+
+class nebula::profile::prometheus::exporter::node (
+  Array $covered_datacenters = [],
+  String $default_datacenter = 'default',
+) {
+  case lookup('role') {
```
aelkiss: Would this make sense to extract as a function?

daaang: Yes, and done.

daaang:
I don't think it's likely we'll have other cases where puppet needs to calculate the role, since its job is usually just executing the role. But it was a single responsibility violation, and it wasn't hard to fix.

PR #166, opened 2019-08-07
AEIM-2291 - Manage node exporter resources

daaang:
The deb I wrote doesn't provision a user or a systemd service, so this is an attempt to do that more explicitly through puppet. It definitely should be tested in an environment before being merged.


daaang:
Tested on jessie, stretch, and ubuntu (bionic)

manifests/profile/apt.pp
```diff
@@ -15,6 +15,17 @@
   Optional[Hash] $local_repo = undef,
 ) {
 
+  if($facts['os']['family'] == 'Debian') {
+    if $local_repo {
+      apt::source { 'local':
+        *            => $local_repo,
```
daaang: This change is so that our local repo is loaded by Ubuntu nodes as well as Debian nodes. Everything else in this file remains Debian-only.

manifests/profile/prometheus.pp
```diff
@@ -36,6 +48,14 @@
     require => File['/etc/prometheus'],
   }
 
+  $static_nodes.each |$static_node| {
+    concat_fragment { "prometheus node service ${static_node['labels']['hostname']}":
+      tag     => "${::datacenter}_prometheus_node_service_list",
```
daaang: This will let us scrape nodes that aren't managed by puppet (e.g. rhel nodes)

spec/fixtures/hiera/hathitrust.yaml
```diff
@@ -84,6 +85,11 @@ nebula::virtual::users::all_users:
     comment: Solr user
     uid: 123458
     home: /var/lib/solr
+  prometheus:
+    uid: 108
+    gid: prometheus
```
daaang: Adding Prometheus users/groups to the HathiTrust hiera is really just adding more technical debt to nebula, but I couldn't think of a better way to handle this (since hathitrust.yaml overrides this hash instead of merging, which is also our desired behavior for this).

I think the way to actually clean this up is (as with many other problems) to have more than one puppet module so that all these unrelated profiles aren't stepping on each other's toes.

templates/profile/prometheus/exporter/node/defaults.sh.erb
```diff
@@ -0,0 +1,134 @@
+# Managed by puppet (nebula/profile/prometheus/exporter/defaults.sh.erb)
+<%# Set the command-line arguments to pass to the server.
+    Due to shell scaping, to pass backslashes for regexes, you need to double
+    them (\\d for \d). If running under systemd, you need to double them again
+    (\\\\d to mean \d), and escape newlines too. -%>
+ARGS="--collector.diskstats.ignored-devices=^(ram|loop|fd|(h|s|v|xv)d[a-z]|nvme\\d+n\\d+p)\\d+$ \
+      --collector.filesystem.ignored-mount-points=^/(sys|proc|dev|run)($|/) \
+      --collector.netdev.ignored-devices=^lo$ \
+      --collector.textfile.directory=/var/lib/prometheus/node-exporter"
```
daaang: These are Debian's defaults. I converted the comments to erb comments so that the final file will be clean, but since none of us have really looked at our options yet, I figured it'd be good to leave the comments here for the moment.

PR #175, opened 2019-08-21
AEIM-2200: Datamart vhost

aelkiss:
- Splits out www.lib vhosts to separate files
- Refactors handling for access log & error log - now abstracted in the www_lib::vhost defined type

I have verified that datamart at least loads the front page and that I can run a "catalog searchbox" inventory query.


manifests/profile/www_lib/vhosts/default.pp
```diff
@@ -0,0 +1,46 @@
...
+#
+# Default (redirection) vhosts for www.lib.umich.edu environment
+#
+# @example
+#   include nebula::profile::www_lib::vhosts::default
+
+class nebula::profile::www_lib::vhosts::default (
+  String $prefix,
+  String $domain,
+  String $vhost_root,
```
aelkiss: should get rid of this parameter for `www_lib::vhost`; it's only used to construct the default docroot and to allow access to `/www/www.lib/cgi`. It also looks like the ScriptAlias for /cgi is missing, and it seems unlikely that ScriptAlias is really needed by default for all vhosts (rather than just www.lib.umich.edu)

manifests/profile/www_lib/vhosts/www_lib.pp
```diff
@@ -0,0 +1,76 @@
...
+# nebula::profile::www_lib::vhosts::www_lib
+#
+# www.lib.umich.edu virtual host
+#
+# @example
+#   include nebula::profile::www_lib::vhosts::www_lib
+class nebula::profile::www_lib::vhosts::www_lib (
+  String $prefix,
+  String $domain,
+  String $vhost_root,
```
aelkiss: get rid of this, see comment above

manifests/profile/www_lib/vhosts/www_lib.pp
```diff
@@ -0,0 +1,76 @@
...
+  String $ssl_cn = 'www.lib.umich.edu'
+) {
+
+  nebula::apache::www_lib_vhost { 'www.lib-ssl':
+    servername                    => "${prefix}www.${domain}",
+    ssl                           => true,
+    vhost_root                    => $vhost_root,
```
aelkiss: get rid of this and add scriptalias, see comment above

aelkiss: or maybe just make it a local variable here rather than passing through to `www_lib::vhost`

PR #183, opened 2019-08-29
AEIM-1663: Configure TSM via puppet

aelkiss:
Specifically for HT backup servers.

How to include the TSM profile for other nodes running TSM is out of scope for this PR.

### DIFF
--- a/files/tsm/tsm.service
+++ b/files/tsm/tsm.service
@@ -1,0 +1,16 @@
+[Unit]
+Description=Tivoli Storage Manager Client
+After=network.target
+
+[Service]
+Type=simple
+Environment="LANG=en_US.UTF-8"
+Environment="LC_CTYPE=en_US.UTF-8"
+Environment="LC_ALL=en_US.UTF-8"
+ExecStart=/usr/bin/dsmc sched
+RemainAfterExit=no
+Restart=always
+
+[Install]
+WantedBy=multi-user.target
+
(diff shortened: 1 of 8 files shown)


skorner:
Functionally it all looks good to me. I agree with requiring manual password interaction for the initial password saving and optional encryption setup. It's not worth the effort of trying to automate that now, especially as our trajectory is to have fewer VMs running TSM. The VMs that will continue to run TSM are likely to be specialized and the last to be fully puppetized.

Whether we want to continue comprehensive backup of /opt and all of /var is debatable, but not worth answering now.

It might be nice to seed inclexcl with common defaults we want in all cases. I think that might be limited to:

```
exclude         /.../.nfs*
exclude.dir     /.../.snapshot
```

I assume those excludes could also be moved into dsm.sys after the four already in that file. I'm not sure how the current default came to be.

aelkiss:
> It might be nice to seed inclexcl with common defaults we want in all cases. I think that might be limited to:
> 
> ```
> exclude         /.../.nfs*
> exclude.dir     /.../.snapshot
> ```
> 
> I assume those excludes could also be moved into dsm.sys after the four already in that file. I'm not sure how the current default came to be.

Added as the default.

aelkiss:
My recollection with regard to `SErvername` is that there was a reason for this, but nobody really wants to test if this is just chicken-waving or if in the current version it works fine with normal capitalization.

aelkiss:
I'll merge as soon as Travis is happy here.

PR #186, opened 2019-08-30
AEIM-2207 - Add staff.lib vhost

botimer:
There are a couple of items here that may be useful for other purposes.
In particular, the PHP 5 / mod_php stuff is pretty verbose as well as
arcane. The array / comment / loop helps a little.

### DIFF
--- a/manifests/profile/www_lib/vhosts/staff_lib.pp
+++ b/manifests/profile/www_lib/vhosts/staff_lib.pp
@@ -1,0 +1,163 @@
+# Copyright (c) 2019 The Regents of the University of Michigan.
+# All Rights Reserved. Licensed according to the terms of the Revised
+# BSD License. See LICENSE.txt for details.
+
+# nebula::profile::www_lib::vhosts::staff_lib
+#
+# staff.lib.umich.edu virtual host
+#
+# @example
+#   include nebula::profile::www_lib::vhosts::staff_lib
+#
+class nebula::profile::www_lib::vhosts::staff_lib (
+  String $prefix,
+  String $domain,
+  String $ssl_cn = 'staff.lib.umich.edu',
+  String $vhost_root = '/www/staff.lib',
+  String $docroot = "${vhost_root}/web"
+) {
+
+  # These directories are all served by PHP5 using mod_php and require Cosign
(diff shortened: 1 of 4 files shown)


manifests/profile/www_lib/vhosts/staff_lib.pp
```diff
@@ -0,0 +1,168 @@
...
+      },
+      {
+        provider       => 'directory',
+        path           => "${vhost_root}/alida-tmp/current/public",
+        allow_override => ['None'],
+        options        => ['FollowSymlinks'],
+        require        => $nebula::profile::www_lib::apache::default_access,
+      },
+      {
+        # Deny access to raw php sources by default
```
aelkiss: This might be already configured via a top-level apache include, not sure...

manifests/profile/www_lib/vhosts/staff_lib.pp
```diff
@@ -0,0 +1,168 @@
...
+        rewrite_rule => '^(/alida.*)$ http://app-alida:40160$1 [P]'
+      },
+    ],
+
+    custom_fragment               => @(EOT)
+      ProxyPassReverse /alida http://app-alida:40160/
+
+      # Enable http authorization headers
+      <IfModule setenvif_module>
+        SetEnvIfNoCase ^Authorization$ "(.+)" HTTP_AUTHORIZATION=$1
```
aelkiss: We don't need to include the <IfModule> there. We should just be able to do

```
setenvifnocase => [ '^Authorization$ "(.+)" HTTP_AUTHORIZATION=$1' ]
``

manifests/profile/www_lib/vhosts/staff_lib.pp
botimer: Thanks for flagging this. It doesn't appear to be anywhere more general, and I'm not ready to rely on the module for invisible defaults, so I'll leave this here.

manifests/profile/www_lib/vhosts/staff_lib.pp
botimer: I've added this to the www_lib_vhost profile.

PR #187, opened 2019-09-05
AEIM-2313 php

rrotter:
Applies our config changes for fpm's config.ini. We're already installing the php fpm daemon in php73.pp, so no action required there. 

Also fix a regression w/ BMC vhost (missing apache log dir)

### DIFF
--- a/manifests/profile/www_lib/php.pp
+++ b/manifests/profile/www_lib/php.pp
@@ -75,19 +75,4 @@ class nebula::profile::www_lib::php () {
       # XML_Util
     },
   }
-
-  # TODO: migrate this to fpm config
-  #  class { 'php::apache_config':
-  #
-  #    settings     => {
-  #      'PHP/short_open_tag'          => 'On',
-  #      'PHP/max_input_vars'          => '2000',
-  #      'PHP/memory_limit'            => '256M',
-  #      'PHP/error_reporting'         => 'E_ALL & ~E_DEPRECATED',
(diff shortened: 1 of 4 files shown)


manifests/profile/www_lib/php.pp
```diff
@@ -76,18 +76,17 @@
...
-  #      'PHP/short_open_tag'          => 'On',
-  #      'PHP/max_input_vars'          => '2000',
-  #      'PHP/memory_limit'            => '256M',
-  #      'PHP/error_reporting'         => 'E_ALL & ~E_DEPRECATED',
-  #      'PHP/upload_max_filesize'     => '32M',
-  #      'Date/date.timezone'          => 'America/Detroit',
-  #      'mail function/sendmail_path' => '/usr/sbin/sendmail -t -i'
-  #    },
-  #  }
+  php::config { 'fpm php.ini':
```
botimer: Since this is specific to 7.3, we should move it to that profile. I don't see anything here that is www-lib specific.

botimer:
Let's move the config to the php73 profile and and do the rubocop/lint stuff, then I think this is ready.

manifests/profile/www_lib/php.pp
rrotter: Looked at other places where php73 is included and it doesn't look like this should be a problem. Done.

rrotter:
Okay, should have all the lint out this time.

PR #188, opened 2019-09-09
Add heirloom-mailx package

aelkiss:
Some of the cron jobs expect to have the mailx from RHEL, which is known as 'heirloom-mailx' on Debian. It provides some different command line flags and features including being able to attach a file.


aelkiss:
Still getting spammy emails without working file attachments, so I need to figure out in more detail what's going on with this cron job...

aelkiss:
The script wasn't actually calling heirloom-mailx. Now it is, and the mail is working properly.

PR #189, opened 2019-09-09
AEIM-2242 - Fix up tools_lib to succeed on first run

botimer:
There were two small problems popping up when running on a new AWS VM
provisioned with this role tag:

 1. The mail_recipient had no default value, so we needed a node file
 2. The postgres profile didn't have sufficient dependency information
    on the backup directories

This change introduces a nebula::automation_email to parallel root_email
and a few requires to ensure that postgres is installed before the
backup directories and cron entry are managed.

With a mail_recipient value in common.yml, these hosts can be
provisioned with only an AWS tag.

### DIFF
--- a/manifests/profile/tools_lib/postgres.pp
+++ b/manifests/profile/tools_lib/postgres.pp
@@ -40,24 +40,27 @@ class nebula::profile::tools_lib::postgres (
   }
 
   file { $pg_backup_dir:
-    ensure => 'directory',
-    mode   => '2775',
-    owner  => 'root',
-    group  => 'postgres'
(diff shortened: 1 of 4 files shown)


botimer:
@daaang -- can you please double check how I've handled the default? I'm not sure if it will complain on lookup if there is no value in hiera, which we would want it to do.

botimer:
@daaang -- Scratch that... Because the specs run with _only_ the specified hiera fixture, removing the mail_recipient from the one in use and running the tests confirmed that a missing value fails the catalog. Sweet!

PR #190, opened 2019-09-10
correct tools/postgres email default

rrotter:
Don't rely on tools specific email key
Rely on new global key
Inherit said key properly (via confluence manifest)


rrotter:
This successfully runs puppet on a new test.tools instance w/o a node hiera file

PR #193, opened 2019-09-16
AEIM-2209 - Add Publishing/Wordpress vhosts

botimer:


manifests/profile/www_lib/apache.pp
```diff
@@ -119,6 +119,8 @@
     }
   }
 
+  include nebula::profile::www_lib::vhosts::publishing
```
daaang: Any reason this isn't part of the above vhost array?

manifests/profile/www_lib/vhosts/publishing.pp
```diff
@@ -0,0 +1,133 @@
...
+        rewrite_rule => '^(.*)$ https://%{HTTP_HOST}$1 [L,NE,R]'
+      },
+    ],
+  }
+
+  # Note that this vhost listens on 443 but is HTTP, because HAProxy terminates
+  # SSL. We pass an ENV var to signal that the request is over HTTPS for app
+  # URL generation, etc.
+  apache::vhost { 'www.publishing-https':
+    servername      => 'https://www.publishing.umich.edu',
```
daaang: This is the only servername I can see with the protocol baked in—is that on purpose?

manifests/profile/www_lib/apache.pp
botimer: These things don't follow our pattern of flexing the prefix/domain at all, so I decided to not add unused and empty params to the new profile. The resources in there use `apache::vhost` directly, too.

manifests/profile/www_lib/vhosts/publishing.pp
botimer: Yeah, this is a little bit odd, but I'm preserving how it was done before. I think different apps/versions respond differently to the server name (as delivered in the environment for mod_php) and the HTTPS env var we set. Belt and suspenders, I guess.

The effect here is to say to PHP/WordPress: you are on `https://...`, even though that's normally only conditioned on whether SSL is enabled or not. SSL is never enabled in this configuration because it is done at the load balancer.

PR #194, opened 2019-09-16
AEIM-2205 - Add vhost for www.press.umich.edu

daaang:
Other than adding a `setenv` parameter to `nebula::apache::www_lib_vhost`, this doesn't much modify anything that already exists.

### DIFF
--- a/manifests/profile/www_lib/vhosts/press.pp
+++ b/manifests/profile/www_lib/vhosts/press.pp
@@ -1,0 +1,99 @@
+# Copyright (c) 2019 The Regents of the University of Michigan.
+# All Rights Reserved. Licensed according to the terms of the Revised
+# BSD License. See LICENSE.txt for details.
+
+# nebula::profile::www_lib::vhosts::press
+#
+# press virtual host
+#
+# @example
+#   include nebula::profile::www_lib::vhosts::press
+class nebula::profile::www_lib::vhosts::press (
+  String $prefix,
+  String $domain,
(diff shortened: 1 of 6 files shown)


manifests/profile/www_lib/vhosts/press.pp
```diff
@@ -0,0 +1,84 @@
...
+        path          => $docroot,
+        options       => 'FollowSymLinks',
+        allowoverride => 'None',
+        require       => $nebula::profile::www_lib::apache::default_access,
+      }, {
+        provider   => 'location',
+        path       => '/script/press',
+        options    => 'None',
+        require    => $nebula::profile::www_lib::apache::default_access,
+        sethandler => 'fcgid-script',
```
skorner: This requires adding mod_fcgid to our apache config. The alternative is to switch to mod_proxy_fcgi as we did for imgsrv in HT. That is also how we are supporting PHP 7.3. @aelkiss Had you thought about this for the Press?

aelkiss: mod_fcgid isn't available on stretch, I think, so I was expecting that we would use mod_proxy_fcgi.

botimer: Mojolicious is Plack/PSGI-ready, so we set this up as a plackup service and very-much-like-our-ruby rewrite/proxy. Nothing FCGI-specific needed for this vhost.

PR #198, opened 2019-09-20
AEIM-2206: Create redirects vhost

malakai97:
This is a 1:1 buildout of the existing functionality in what we refer to
as the redirects vhost as defined in redirects.conf

I'm particularly concerned with the changes to `theater-historiography`, as my changes here _remove_ some of the functionality.

### DIFF
--- a/manifests/apache/redirect_vhost_https.pp
+++ b/manifests/apache/redirect_vhost_https.pp
@@ -5,11 +5,14 @@
 define nebula::apache::redirect_vhost_https (
   Array[String] $serveraliases = [],
   String $target = "https://www.${title}/",
-  String $ssl_cn = $title,
+  $priority = false,
 ) {
-  nebula::apache::www_lib_vhost { "${title}-https":
-    ssl             => true,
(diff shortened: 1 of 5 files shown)


manifests/apache/redirect_vhost_https.pp
```diff
@@ -5,11 +5,14 @@
...
-  String $ssl_cn = $title,
+  $priority = false,
 ) {
-  nebula::apache::www_lib_vhost { "${title}-https":
-    ssl             => true,
-    ssl_cn          => $ssl_cn,
+  nebula::apache::redirect_vhost_http { "${title}":
+    serveraliases => $serveraliases
+  }
+  apache::vhost { "${title}-redirect-https":
```
aelkiss: While in the _current_ functionality apache doesn't terminate ssl, the replacement should; therefore, I think you'll want to keep what was there before with the `$ssl_cn` parameter that gets passed through to the `nebula::apache::www_lib_vhost` rather than using `apache::vhost` here directly.

PR #199, opened 2019-09-20
Delete prometheus logs but expose error count

daaang:
Sometimes the prometheus node exporter has errors, and those errors fill
up /var/log. This way, they'll be cleared out every 30 minutes, but also
their presence won't be lost.

### DIFF
--- a/lib/facter/prometheus_errors_total.rb
+++ b/lib/facter/prometheus_errors_total.rb
@@ -15,7 +15,7 @@ Facter.add(:prometheus_errors_total) do
     error_count = %r{(?<=error gathering metrics: )\d*(?= error)}
 
     if File.size? log_path
-      [1, File.read(log_path)[error_count].to_i].max
+      [1, Integer(File.read(log_path)[error_count])].max
     else
       0
     end
--- a/lib/facter/prometheus_errors_total.rb
+++ b/lib/facter/prometheus_errors_total.rb
@@ -1,0 +1,23 @@
+# frozen_string_literal: true
+
+require 'yaml'
+
+# This will read the Prometheus node exporter's log and look for errors.
+# The result seen by puppet is nothing more than an integer which it can
+# pass to a file that Prometheus will scrape.
+#
+# We're not interested in details in this case, since it's just a
+# metric. If you want more info than error count, then that's what logs
+# are for.
+Facter.add(:prometheus_errors_total) do
+  setcode do
+    log_path = '/var/log/prometheus-node-exporter.log'
+    error_count = %r{(?<=error gathering metrics: )\d*(?= error)}
+
+    if File.size? log_path
+      [1, File.read(log_path)[error_count].to_i].max
+    else
+      0
+    end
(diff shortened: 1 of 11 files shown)


lib/facter/prometheus_errors_total.rb
```diff
@@ -0,0 +1,23 @@
...
+# We're not interested in details in this case, since it's just a
+# metric. If you want more info than error count, then that's what logs
+# are for.
+Facter.add(:prometheus_errors_total) do
+  setcode do
+    log_path = '/var/log/prometheus-node-exporter.log'
+    error_count = %r{(?<=error gathering metrics: )\d*(?= error)}
+
+    if File.size? log_path
+      [1, File.read(log_path)[error_count].to_i].max
```
malakai97: Prefer `Integer(x)` over `#to_i` as the latter fails silently if the value can't be cast.

manifests/profile/prometheus/exporter/node.pp
```diff
@@ -24,6 +24,7 @@
   Optional[String] $version = undef,
   Array $covered_datacenters = [],
   String $default_datacenter = 'default',
+  String $log_file = '/var/log/prometheus-node-exporter.log',
```
malakai97: Magic string appears twice; here and in prometheus_errors_total.rb. Is there any clean way to avoid that?

manifests/profile/prometheus/exporter/node.pp
```diff
@@ -56,10 +57,33 @@
 
   file { '/etc/systemd/system/prometheus-node-exporter.service':
     content => template('nebula/profile/prometheus/exporter/node/systemd.ini.erb'),
-    notify  => Service['prometheus-node-exporter'],
+    notify  => [Service['prometheus-node-exporter'], Exec['reload systemctl daemons for prometheus']],
```
malakai97: I am a bit on the fence on this. On one hand, I would rather see this as a generic systemctl reload not specific to prometheus, so that additional notifications in unrelated profiles do not result in more than one reload. On the other, this follows the pattern set in our `docker.pp` profile. This is acceptable in its current form.

lib/facter/prometheus_errors_total.rb
daaang: Done!

manifests/profile/prometheus/exporter/node.pp
daaang: Subscribers now exist, which kind of addresses this.

manifests/profile/prometheus/exporter/node.pp
daaang: Unfortunately, custom facts can't read hieradata, since the hieradata is generated by the puppetserver after the puppet agent runs facter. I did at least move this out of the parameter area, so it doesn't imply itself to be something that can or should be ever changed.

malakai97: Good to know. I like this change.

PR #203, opened 2019-09-26
[LITHELP-138]

rrotter:


botimer:
These host aliases work as of now, without this PR. Closing as stale.

PR #204, opened 2019-09-27
AEIM-2414 - Allow path blocking via hiera

daaang:


daaang:
Don't merge this until the corresponding hiera change is made

daaang:
Not doing this.

PR #209, opened 2019-10-10
Condition detailed chipmunk/perl deps on OS release

botimer:
re: PFDR-255


rrotter:
Looks good. I don't like the jessie exclusion boilerplate we've been copy/pasting around, but that's irrelevant since the same line is already elsewhere in the code and we can delete them when Jessie is gone.

botimer:
Totally agreed... This is very smelly and temporary.

PR #211, opened 2019-10-20
AEIM-2442 Add shib profile for www_lib

skorner:
I copied the HT shib profile and defined a new source for the config files on the puppet-master. I populated that source directory on the puppet-master with files from a current web server.


botimer:
I've made adjustments here to rely on the source/fileserver permissions, rather than making them explicit in the profile. This means that, for example, keys can be added without being listed.

manifests/role/webhost/www_lib_vm.pp
```diff
@@ -35,9 +35,12 @@
   include nebula::profile::www_lib::dependencies
   include nebula::profile::www_lib::perl
   include nebula::profile::www_lib::php
-  #  include nebula::profile::www_lib::shibboleth
   include nebula::profile::www_lib::apache
   include nebula::profile::unison
 
+  class { 'nebula::profile::shibboleth':
+    config_source => 'puppet:///shibboleth-www_lib'
```
daaang: `not-a_fan` But otherwise looks good to me.

botimer: Heh! We'll get that later. Thanks!

botimer:
This is not quite ready... we need shib2.conf and to fix one more problem on the modes.

botimer:
This is all set now. Verified as working in test environment.

PR #212, opened 2019-10-22
Block IPv6: purge chains and default to drop

sulemandiwan:
This should make the rules.v6 file block all ipv6 traffic by making chains empty and default to drop.


sulemandiwan:
I haven't tested it on a machine yet but will do that now that the thing is here.

sulemandiwan:
Separation was intentional for a variety of reasons.

PR #214, opened 2019-11-01
allow requests to ht-web-testing from datacenters

aelkiss:


skorner:
Looks good and corrects an oversight.

PR #219, opened 2019-11-12
PFDR-263 - Add PAM/sshd config

botimer:
This brings in a stock stretch /etc/pam.d/sshd config and introduces an
inline template to include that and then add the umask settings.
Specifically, the umask is set to 0002 on ssh logins, so rsync uploads
can preserve setgid and group write permissions readily with:

rsync -rzvtP --no-group --no-perms --chmod=ugo=rwX

This formula is mentioned in the rsync man page, in the --perms section.

### DIFF
--- a/files/pam.d/sshd-stretch
+++ b/files/pam.d/sshd-stretch
@@ -1,0 +1,56 @@
+# PAM configuration for the Secure Shell service
+
+# Standard Un*x authentication.
+@include common-auth
+auth required pam_duo.so
+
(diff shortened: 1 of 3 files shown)


botimer:
Tests are failing for some unrelated reason on Mac only. That doesn't block merge, here -- it's just the first time I've hit this particular failure. It's related to the explicit `ip6tables` provider specified in the firewall profile. I'll take up diagnosis and remediation elsewhere.

PR #223, opened 2019-11-27
HT-379: testing isn't buster any more

aelkiss:
initial installation of jhove/libactivation-java doesn't work on stretch
currently because buster is stable (not testing) now


botimer:
Just a note that this implies that we will not use this role on jessie -- a sound enough assumption that we don't need to deal with it in more detail.

PR #228, opened 2019-12-04
[WIP] AEIM-1985 Add initial Solr file for general case

sulemandiwan:
Note: I will move the solr.pp file one level up and remove the directory. Right now just want to discuss the contents.


manifests/profile/solr.pp
```diff
@@ -0,0 +1,117 @@
+# Copyright (c) 2018-2019 The Regents of the University of Michigan.
+# All Rights Reserved. Licensed according to the terms of the Revised
+# BSD License. See LICENSE.txt for details.
+
+# nebula::profile::solr
+#
+# Provisions a single instance of solr.
+# For each given "shard", this provisions the lib and conf directory 
+# for an "x" and "y" core and symlinks the data directory for the shard.
```
botimer: Only relevant for LSS; remove

manifests/profile/solr.pp
```diff
@@ -0,0 +1,117 @@
...
+#
+# @param timezone  The timezone to use for logs, etc. Should normally be set via Hiera.
+# @param coredata  A map of core names to the full paths on disk for the data
```
botimer: Only for LSS

manifests/profile/solr.pp
```diff
@@ -0,0 +1,117 @@
...
+# @param port   The port solr should listen on
+# @param heap   How much memory to allocate for the Solr heap.
+#
+# @example
+#   class { 'nebula::profile::solr':
+#     timezone                 => 'America/New_York'
+#     coredata                 => {
+#                      'core1' => '/path/to/core_1/data',
+#                      'core2' => '/path/to/core_2/data'
+#                   }
```
botimer: Remove

manifests/profile/solr.pp
```diff
@@ -0,0 +1,117 @@
...
+#   }
+class nebula::profile::solr (
+  String $timezone,
+  Hash[String,String] $coredata = {},
```
botimer: Remove

manifests/profile/solr.pp
```diff
@@ -0,0 +1,117 @@
...
+  String $base = '/var/lib/solr',
+  String $home = "${base}/home",
+  String $logs = "${base}/logs",
+  String $heap = '32G',
```
botimer: Let's start with 8G here.

manifests/profile/solr.pp
```diff
@@ -0,0 +1,117 @@
...
+      group => 'solr';
+    [$base, $home, $logs]:
+      ensure => 'directory',
+      mode   => '0750';
+    $log4j_props:
+      content => template('nebula/profile/hathitrust/solr/log4j.properties.erb');
+    $solr_in_sh:
+      content => template('nebula/profile/hathitrust/solr/solr.in.sh.erb');
+    "${home}/solr.xml":
+      content =>  template('nebula/profile/hathitrust/solr/solr.xml.erb')
```
botimer: Each of these templates will need general versions.

manifests/profile/solr.pp
```diff
@@ -0,0 +1,117 @@
...
+  }
+
+  file { '/etc/systemd/system/solr.service':
+    owner   => 'root',
+    group   => 'root',
+    content => template('nebula/profile/hathitrust/solr/solr.service.erb')
```
botimer: Need general template.

manifests/profile/solr.pp
```diff
@@ -0,0 +1,117 @@
...
+  }
+
+  file {
+    "${home}/lib":
+      ensure  => 'directory',
+      source  => 'puppet:///modules/nebula/solr/lib',
```
botimer: Let's confirm -- I think there are some Java libraries that are used for LSS that are not needed here...

manifests/profile/solr.pp
```diff
@@ -0,0 +1,117 @@
...
+
+        "${subcore_home}/data":
+          ensure => 'link',
+          target => $path;
+
+        "${subcore_home}/conf/schema.xml":
+          source => "puppet:///modules/nebula/solr/schema_${suffix}.xml"
+      }
+    }
+  }
```
botimer: All goes away as managed by separately managed core resources.

manifests/profile/solr.pp
sulemandiwan: Gotcha. Makes sense.

manifests/profile/solr.pp
skorner: It looks like none of our app-supporting general solrs are above 3G in production (search solr is an outlier at 8G), and that's with some of them hosting multiple cores. Even the solr backing the library web site Drupal is only 1G.  I'd recommend 1G as the default for minimal needs and we increase via hiera as indicated.

sulemandiwan: Done.

manifests/profile/solr.pp
skorner: It is correct that everything currently under files/solr_lss/lib is only used by HT solr deployments (LSS, catalog, and maybe feddocs).

sulemandiwan: Okay removing this stanza entirely. If we want a standardized ${home}/lib for some reason later it can be added.

manifests/profile/solr.pp
sulemandiwan: First pass is done. Made a small edit. Doesn't look like it would need further edits.

manifests/profile/solr.pp
sulemandiwan: All of these turned into generic versions also. They didn't need any edits. However, if these are values that need to be altered to fit a minimum generic setup I don't know those.

sulemandiwan: Going to resolve this conversation as this part is done.

sulemandiwan:
@botimer @skorner I don't know how to flag it is ready for another review. So I am saying here it is ready for another review. :)

templates/profile/solr/solr.in.sh.erb
```diff
@@ -0,0 +1,39 @@
...
+
+# To address http://lucene.472066.n3.nabble.com/Re-Several-critical-vulnerabilities-discovered-in-Apache-Solr-XXE-amp-RCE-td4358308.html
+# See AEIM-549
+#
+# 2017-10-17 skorner
+SOLR_OPTS="$SOLR_OPTS -Ddisable.configEdit=true"
+
+# tburtonw May 25 2018: prevent rotating logs every time solr is restarted
+SOLR_LOG_PRESTART_ROTATION=false
+
```
botimer: Remove -- only applied for an LSS thing

botimer:
Okay, based on discussion, let's add this profile to the standalone role and leave it on a branch until we can verify via environment/sandbox. We'll need to confirm that any standalone hosts will not conflict with the profile before we merge to master.

templates/profile/solr/solr.in.sh.erb
sulemandiwan: Did this and repushed the squashed commit.

manifests/profile/solr.pp
```diff
@@ -0,0 +1,48 @@
+# Copyright (c) 2018-2019 The Regents of the University of Michigan.
+# All Rights Reserved. Licensed according to the terms of the Revised
+# BSD License. See LICENSE.txt for details.
+
+# nebula::profile::solr
+#
+# Provisions a basic instance of solr. Cores and specific implementation
```
botimer: Remove the second sentence here -- that cores are elsewhere is implicit and the "specific implementation" bit is fluff.

I would just say: `Install Solr with base configuration.`

templates/profile/solr/solr.in.sh.erb
```diff
@@ -0,0 +1,39 @@
+SOLR_JAVA_HOME="<%= @java_home %>"
+SOLR_HEAP="<%= @heap %>"
+
+GC_LOG_OPTS="-verbose:gc -XX:+PrintHeapAtGC -XX:+PrintGCDetails \
```
skorner: I missed that HT had this enabled by default. It's much too verbose to always be on, and in general GC logging is something we typically only need for troubleshooting (Fulcrum needs it now, for example!). @botimer Can/should we make this something that is enabled only when needed based on a hiera flag?

botimer: TL;DR: Add a parameter to the profile for supplemental JAVA_OPTS and expose this in the systemd unit as an environment variable. Append it to the default in the solr.in.sh script (with bash, not erb). Keep an external recipe book of options we might add for a given node via hieradata.

---

Where possible, we should push these cases toward a unified interface. That's essentially what projects like Tomcat and others have already done in their startup scripts with JAVA_OPTS. They don't know what you'll need or how you'll start things, or even what is supported by the JVM, but they know there is an environment where you can set a value they can put in the right spot.

This also aligns with systemd, where the unit files can hold environment variables directly, or point to an EnvironmentFile. The Twelve-Factor [config section](https://12factor.net/config) discusses this topic as well.

Expanding magic values into the script is better than nothing, but opaque. You can't tell from the concrete script whether a value is intrinsic to the package or injected by the management of the package. It's effectively a custom-built package for that node, and the variability is no longer contained. Think about doing a digest of a directory tree on disk -- the root of the install of the package is now machine-specific. It's better to keep that fixed and have it point to a much smaller, external area declared as volatile; the same as /usr/local vs. /etc.

So, what's that mean here?

We should make sockets in the script and plugs in the systemd unit. Giving special treatment to a couple of key items like max heap is fine, but it's ultimately a big JAVA_OPTS string. Breaking everything out is a lot more work and more likely to add confusion than clarity -- we don't want our management layer to create penalties for expertise with the thing being managed.

We should use puppet to install a fixed package and rely on systemd to set machine-specific environment variables. We can manage those variables either in the unit itself (already a template) or with an EnvironmentFile (a new, companion template).

This approach is all a subclass of separating the code (fixed) from the data (variable), and maximizing the portion of the system that can be held constant. It also favors common mechanisms over specialized, pushing detail outward -- the specialized stuff is transformed into standard stuff as far outside as possible, so the layers are progressively hardened, looking outside-in. In this case, hiera -> env file -> systemd unit -> solr -> jvm.

skorner: Thanks for the thoughts and both long and short form information! That all seems reasonable to me. I imagine for the scope of the current pass we can delete this and introduce it when needed or when we apply the more general option-management solution you listed.

manifests/role/app_host/hyrax.pp
```diff
@@ -0,0 +1,23 @@
...
+# All Rights Reserved. Licensed according to the terms of the Revised
+# BSD License. See LICENSE.txt for details.
+
+# nebula::role::app_host::prod
+#
+# Standalone application host including apache & mysql
+#
+# @example
+#   include nebula::role::app_host::standalone
+class nebula::role::app_host::standalone {
```
botimer: Travis is failing this branch because the filename and role do not match. Also, please make sure the copyright date and comments are correct, then squash it all down to one commit.

botimer: I see there is a weird self-merge in the history... If a pull / rebase doesn't succeed trivially, do it with `merge --squash`:

```
git branch -m this-will-be-squashed
git checkout -b AEIM-1985-general-solr-puppetization master
git merge --squash this-will-be-squashed
git commit
# write a good, small commit message
git push -f
git branch -D this-will-be-squashed
```

sulemandiwan: Also it seems like that in the duplicate entries the newer ones make the changes, while the older ones don't contain any changes but are in the log. 

I also tried the command sequence and it didn't work for me. Will touch base with you tomorrow.

sulemandiwan:
Also it seems like that in the duplicate entries the newer ones make the
changes, while the older ones don't contain any changes but are in the log.

I also tried the command sequence and it didn't work for me. Will touch
base with you tomorrow.

sulemandiwan:
Aside from fixing the git log issues it seems that travis wants me to add entries for "spec/classes" directory in all_roles_spec.rb but likely also in the role folder under it. Let's also talk about this after I get the git log fixed.

manifests/role/app_host/hyrax.pp
sulemandiwan: Okay the branch issues seem fixed now. Changes made and squashed and pushed. Waiting for final Travis. Then it will be ready to be merged tomorrow during the day.

templates/profile/solr/solr.in.sh.erb
sulemandiwan: This is fixed.

manifests/role/app_host/hyrax.pp
sulemandiwan: This is fixed.

sulemandiwan:
Poking at travis a little. Can roll back this change if not needed at all for some reason.

manifests/role/app_host/hyrax.pp
sulemandiwan: Now it is fixed. *koff*

sulemandiwan:
It passed the test. Please verify. It does need the solr test properly added though.

manifests/role/app_host/solr.pp
```diff
@@ -0,0 +1,23 @@
...
+# All Rights Reserved. Licensed according to the terms of the Revised
+# BSD License. See LICENSE.txt for details.
+
+# nebula::role::app_host::prod
+#
+# Standalone application host including apache, mysql, redis, solr
+#
+# @example
+#   include nebula::role::app_host::solr
+class nebula::role::app_host::solr {
```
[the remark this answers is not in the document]
sulemandiwan: This and above two changes done and pushed.

sulemandiwan:
@botimer Could you have a peek at this? The solr_spec.rb file is what needs peekage. :)

spec/classes/profile/solr_spec.rb
```diff
@@ -0,0 +1,69 @@
+# frozen_string_literal: true
+
+# Copyright (c) 2018 The Regents of the University of Michigan.
```
botimer: year

spec/classes/profile/solr_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+      end
+
+      context 'with directory' do
+        it do
+          [
+            '/var/lib/solr',
+            '/var/lib/solr/home',
+            '/var/lib/solr/logs',
+          ].each do |path|
+            is_expected.to contain_file(path).with(
```
botimer: This should be inverted. Each of these should get its own `it do`. This is so they show as itemized examples and anything that fails will do so individually.

botimer: Same applies below.

sulemandiwan: Okay fixed them.

One final blessing and I will merge it on Monday (to avoid Friday Fiascos).

spec/classes/profile/solr_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+              owner: 'solr',
+              group: 'solr',
+              ensure: 'directory',
+              mode: '0750',
+            )
+          end
+        end
+      end
+
+      context 'with config file' do
```
botimer: These are identical to the specs above. Copy pasta?

sulemandiwan: Yeah... Gonna fix it. Also... I realize that hyrax is just using the chipmunk file to get by but it shouldn't be.

sulemandiwan: FIxed that but also that lead to redoing some of solr file settings in the solr.pp and then decided to add a full hyrax test file. So working on the kinks in that.

sulemandiwan:
I rolled back the tests for hyrax since there was a lot of dependency hunting with little return. So that is using chipmunk again. 

However, I have updated the solr.pp file and solr.pp tests to accommodate that file.

@botimer It would be nice if you might do one last test for solr installation on the test VM. Basically just to see if it breaks anything or not. I think it shouldn't but safety and all.

After that it is done and squash and merge I can do.

spec/classes/profile/solr_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+require 'spec_helper'
+
+describe 'nebula::profile::solr' do
+  on_supported_os.each do |os, os_facts|
+    context "on #{os}" do
+      let(:facts) { os_facts }
+
+      it { is_expected.to compile }
+
+      context 'with package' do
```
daaang: This all mostly looks good. My only comment is that this isn't how contexts are meant to be used. A context implies something external to the profile, i.e. a parameter change or a fact about the machine. For example, [kubernetes_spec.rb][1] has a number of contexts, and you can see how each sets a parameter, creating the described context.

[1]: https://github.com/mlibrary/nebula/blob/master/spec/classes/profile/kubernetes_spec.rb

A context is something about the outside world, like "the machine has less than 2G of ram" or "the hostname is set to default.invalid." It isn't meant to be used for categorical grouping of conceptually related tests. The way we do that is just to put them next to each other.

So I'd like to see the contexts removed. To be honest, I'd also get rid of the array loops, since I don't feel like there's much of a need to DRY out

```ruby
it { is_expected.to contain_package('openjdk-8-jre-headless') }
it { is_expected.to contain_package('solr') }
it { is_expected.to contain_package('lsof') }
```

I personally prefer to spell out each test, especially when the tests are so straightforward. When each test is a few lines long, this gets onerous, but with one-line tests, I feel the repeated assertions add clarity that makes up for the minor repetition. But that's a matter of taste, so I'll leave that decision to you.

sulemandiwan: I replaced context. And thanks for clearing it up. I took another look at context and it is making more sense now. :)

PR #230, opened 2019-12-06
AEIM-2479 - Add docker registry and baseline target host 

daaang:


manifests/named_instance/docker.pp
```diff
@@ -0,0 +1,26 @@
...
+  Integer $host_port,
+  Integer $container_port = 3000,
+) {
+  require nebula::profile::docker
+
+  docker::run { $title:
+    image => $image,
+    ports => "${host_port}:${container_port}",
+  }
+}
```
daaang: This will need to grow before it's actually useful—for example it'll need to create an Apache vhost that forwards traffic to the host port. It may also need additional logic for serving static files in some unified manner for containers. For now, all we have is the bare minimum.

PR #237, opened 2019-12-19
AEIM-2521 - Mount CIFS shares in kubernetes

daaang:


manifests/profile/kubernetes/worker.pp
```diff
@@ -27,6 +29,12 @@
     tag     => "${cluster}_proxy_ips",
   }
 
+  $cifs_mounts.each |$mount_title, $mount_parameters| {
+    nebula::cifs_mount { "/mnt/legacy_cifs_${mount_title}":
```
skorner: What makes these cifs mounts "legacy"?

daaang: I put that there to reflect the intention to not handle cifs mounts this way forever. I'm hoping that kubernetes officially supports cifs at some point in the next couple years, and then we can switch to mounting directly with kubernetes instead of mounting via puppet and treating the mounts as local storage, which is kind of an ugly hack.

skorner: Ah, got it. That makes sense.

PR #238, opened 2019-12-20
[WIP] Add tsm to archivematica for Debian 9 onwards

sulemandiwan:


manifests/role/archivematica.pp
```diff
@@ -13,6 +13,12 @@
   include nebula::profile::duo
   include nebula::profile::networking
 
+  # TSM to be included via puppet starting with Debian 9 onwards
+  if $facts['os']['family'] == 'Debian' \
```
skorner: Archivematica systems are all Ubuntu. Also, there are no older puppet-managed instances, so all with this role can have the TSM profile.

skorner:
The TSM aspect is minor and looks good to me. However, I can't speak to the testing portion. I'd like @botimer or @daaang to at least glance at that and sign off.

manifests/profile/tsm.pp
```diff
@@ -41,7 +41,8 @@
   Array[String] $virtualmountpoints = ['/etc','/opt','/var'],
   Array[String] $exclude_dirs = ['/afs/','/net/','/nfs/','/usr/vice/cache/']
 ) {
-  package { 'tivsm-ba': }
+
+  ensure_packages(['tivsm-ba', 'tivsm-api64', 'gskcrypt64', 'gskssl64',])
```
skorner: We should explore why requiring the single package was not enough to pull in the other three as dependencies. It worked on Debian Stretch.

sulemandiwan: Going to double-check this in case it was an artifact of other transient issues.

PR #242, opened 2020-01-09
AEIM-2008 add clearinghouse role

skorner:
This is mostly Aaron's work from the summer to add a puppet-managed apache config for clearinghouse, plus some updates to finalize the apache and php config. We will add puppet-manage mysql when that work is completed. Conor and the stakeholder want to move to the new AWS hosting now though.

The only change with a broader impact is an increase in our default PHP upload_max_filesize from 32M to 64M. It seemed simplest to increase the default rather than duplicate our full PHP 7.3 config for clearinghouse.

### DIFF
--- a/manifests/profile/clearinghouse/apache.pp
+++ b/manifests/profile/clearinghouse/apache.pp
@@ -45,7 +45,7 @@ class nebula::profile::clearinghouse::apache (
   $ssl_params     = {
     ssl            => true,
     ssl_protocol   => 'all -SSLv2 -SSLv3',
-    ssl_cipher     => 'EECDH EDH+aRSA !RC4 !aNULL !eNULL !LOW !3DES !MD5 !EXP !PSK !SRP !DSS',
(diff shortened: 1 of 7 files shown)


aelkiss:
All looks fine to me. I would probably squash at least 90652c1..efb4888, but otherwise looks fine to me. You will definitely want to get an environment running locally where you can run `bundle exec rake lint` and `rubocop` to catch errors before Travis; also note that `puppet-lint -f file1 file2 file3 ...` will auto-fix most of the indentation errors, etc.

skorner:
Thanks @aelkiss! I was planning to squash all my work into one commit, except maybe [05a1ad2](https://github.com/mlibrary/nebula/pull/242/commits/14862cbd9a0e7bc16d2794f3a7fa2a498ace42e7). Agreed regarding my need for minimal dev and testing support locally. I omitted everything in my most recent Debian build and keep thinking I'm only dabbling, but even that needs some tools.

skorner:
I rebased with master and squashed my work into two commits. There were no code changes from what Aaron approved, so I'll merge after tests pass.

PR #245, opened 2020-01-24
AEIM-2568 - Add https reverse proxy to Prometheus

daaang:
Adding an apache vhost to point 443 at localhost:9090, which requires keys. Adding letsencrypt keys, which require an apache vhost on port 80 for passing the letsencrypt challenge. Trying to use the simplest settings possible.

### DIFF
--- a/manifests/profile/letsencrypt.pp
+++ b/manifests/profile/letsencrypt.pp
@@ -8,4 +8,11 @@ class nebula::profile::letsencrypt (
   class { 'letsencrypt':
     email => $email,
   }
+
+  firewall { '200 HTTP':
+    proto  => 'tcp',
+    dport  => 80,
+    state  => 'NEW',
+    action => 'accept',
+  }
 }
--- a/manifests/profile/letsencrypt.pp
+++ b/manifests/profile/letsencrypt.pp
@@ -2,6 +2,10 @@
 # All Rights Reserved. Licensed according to the terms of the Revised
 # BSD License. See LICENSE.txt for details.
 
+# Let's Encrypt System Requirements
+#
+# This installs certbot and sets it up with our contact email. It also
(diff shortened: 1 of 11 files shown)


templates/profile/prometheus/rules.yml.erb
```diff
@@ -67,3 +67,25 @@ groups:
...
+      )
+  - record: datacenter_role:node_cpu_seconds:max_not_idle_mean30s
+    expr: >
+      max without(hostname, instance)(
+        avg without(cpu)(
+          sum without(mode)(
+            rate(node_cpu_seconds_total{mode!="idle"}[30s])
+          )
+        )
+      )
```
daaang: This change isn't related to this pull request. It's just adding federated metrics to our prometheus rules, and I did it here because our scrapers are all in the development environment right now, so this was the easiest way for them to pick up the change.

manifests/profile/letsencrypt.pp
```diff
@@ -0,0 +1,22 @@
...
+# This installs certbot and sets it up with our contact email. It also
+# opens port 80 to the world, which is required for verifying ownership.
+class nebula::profile::letsencrypt (
+  String $email = lookup('nebula::profile::base::contact_email'),
+) {
+  class { 'letsencrypt':
+    email => $email,
+  }
+
+  firewall { '200 HTTP':
```
botimer: Do we need to specify the chain in that protocol specific string format?

daaang: You mean that `200 HTTP` title? It could be anything as long is it starts with something between 002 and 998.

botimer: No, I was thinking of the `'INPUT:filter:IPv4'` stuff, but I think that is just for the chains, not the individual rules. All good!

PR #247, opened 2020-01-24
Update every module to its newest version

daaang:
Separating this out from the https prometheus work, since it's more than enough on its own.


daaang:
This is no longer relevant now that #407 is merged and working.

PR #248, opened 2020-01-27
AE-7176: Copy www.lib vhost to apps.lib

aelkiss:
Copying rather than abstracting seems like the right way to go here --
@skorner reports "In about 5 months we won't serve www.lib from our web
servers anymore and even before then the apps.lib vhost config will
likely start to diverge from what we have locally for www.lib."


skorner:
This looks good to me. We'll be sure to coordinate with D&D so we prune any www-lib directives that are no longer needed as the transition happens later this year.

PR #251, opened 2020-01-30
Enable systemd monitoring in Prometheus

sulemandiwan:


templates/profile/prometheus/exporter/node/defaults.sh.erb
```diff
@@ -109,6 +109,9 @@ ARGS="--collector.diskstats.ignored-devices=^(ram|loop|fd|(h|s|v|xv)d[a-z]|nvme\
    --collector.stat          Enable the stat collector (default: enabled).
    --collector.supervisord   Enable the supervisord collector (default:
                              disabled).
+#%>
+   --collector.systemd
+<%#
```
daaang: Instead of interrupting the comment, just make this a single added line near the top of the file. This comment was just left here as information to copypaste from.

Maybe add it on line 8? And add a backslash so that subsequent lines are read.

sulemandiwan: Sounds good. Thanks.

sulemandiwan: Okay fixed it.

sulemandiwan: Going to squash the comments locally and repush.

daaang: Ah, still needs a backslash at the end of the line so bash won't treat it as the end of the command

sulemandiwan: Heh. Okay. :) Thanks.

sulemandiwan: Okay this should be correct now.

PR #253, opened 2020-01-31
HT-2113: proxy app for useradmin-dev

aelkiss:
Tested & working in development (punch)


aelkiss:
Actually, useradmin is not the best endpoint for this, since it will mask the existing key generation. I need to think about where to put this (and probably re-deploy the app)

PR #255, opened 2020-02-03
HT-2113: proxy app for useradmin production

aelkiss:
I realized at the last minute that /useradmin in production is something else, so (at least until that can be retired) this will be at /usermanage. I don't want to change the name of the instance, and I would expect eventually this can replace what's at /useradmin.


aelkiss:
I have already updated moku-infrastructure, re-deployed, and tested with the development apache.

aelkiss:
will plan to merge Tuesday morning rather than merging & leaving the office

aelkiss:
I should also test the environment on ht-web-testing first

aelkiss:
Tested & working on ht-web-testing. Merging now.

PR #257, opened 2020-02-06
AE-7308 Add AIGA DEC Dialogue redirect to Fulcrum

sulemandiwan:
It is trivial but need a quick blessing.


sulemandiwan:
Sebastien also approved this in Slack so I am merging it.

PR #258, opened 2020-02-06
AE-7308 Add dialogue.aiga.org redirect

sulemandiwan:
Just adding the aiga redirect that was added by @skorner  to this.


sulemandiwan:
This was already merged, probably locally. This is a left over item.

PR #259, opened 2020-02-07
HT-2231: remove get_brittle_books_data

aelkiss:
Tim reports (regarding the file this job processes): "It's definitely
obsolete. We have a umich process that manages the workflow for
locally-digitized volumes (that were "rejected" for scanning by google).
The "allrejects.txt" file was from an earlier iteration of the process,
which was rewritten about 5 years ago. So, Aaron should just remove
references to it. And since it's not a HathiTrust workflow, you
shouldn't have to mess with it either." -
https://github.com/hathitrust/post_zephir_processing/pull/1#issuecomment-573829040

I am wondering though if just removing all references to this job (and then manually purging the cron job on the server it runs on) is a preferred approach, or if we should instead do

```puppet
'get brittle books data':
  ensure => 'absent';
```

and keep it represented in the code in perpetuity.
```

### DIFF
--- a/manifests/profile/hathitrust/ingest_jobs.pp
+++ b/manifests/profile/hathitrust/ingest_jobs.pp
@@ -12,7 +12,6 @@ class nebula::profile::hathitrust::ingest_jobs(
   String $config,
   String $feed_home,
   String $stats_home,
-  Array[String] $dcu_recipients,
   String $crms_renewals_source,
   String $crms_renewals_dest,
   String $recipient,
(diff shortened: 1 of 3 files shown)


aelkiss:
I've changed this to `ensure => 'absent'`. I'm not sure where the conversation ended up about auditing for these, but let me know if there's something quick I can do here, or if you want to leave this for a future AEIM issue.

aelkiss:
Regarding `grin_reports.pl` - This uploaded a file to ctools, which doesn't exist any more; it looks like the script no longer gets to the point of even attempting to do that because the format of the data it's expecting has changed. In the years since this script actually worked, nobody appears to have complained that it was missing.

PR #263, opened 2020-02-17
AE-7412 - Automatically add defined users to deployhosts

daaang:


daaang:
Moku has additional requirements for users have particular shells and home directories that aren't easy to implement at the moment, so I'm closing this.

PR #266, opened 2020-02-24
AEIM-2640 - Increase shibd timeouts on htvm and www_lib_vm

botimer:
This approach adds a couple of parameters to the profile and then specifies concrete values in the role. No hiera needed, but possible -- always moving toward open-closed.

### DIFF
--- a/spec/classes/role/htvm_webhost_spec.rb
+++ b/spec/classes/role/htvm_webhost_spec.rb
@@ -17,8 +17,9 @@ describe 'nebula::role::webhost::htvm' do
       it { is_expected.to contain_mount('/sdr1').with_options('auto,hard,nfsvers=3,ro') }
 
       it do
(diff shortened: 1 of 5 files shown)


spec/classes/role/htvm_webhost_spec.rb
```diff
@@ -17,8 +17,9 @@
       it { is_expected.to contain_mount('/sdr1').with_options('auto,hard,nfsvers=3,ro') }
 
       it do
-        is_expected.to contain_file('/etc/systemd/system/shibd.service.d/increase-timeout.conf')
-          .with_content("[Service]\nTimeoutStartSec=900")
+        is_expected.to contain_class('nebula::profile::shibboleth')
+          .with(startup_timeout: 1800)
+          .with(watchdog_minutes: '*/30')
```
daaang: I'm surprised this passes, given you've used the String type

botimer: Fixed to use Integer.

PR #267, opened 2020-03-04
Blacklist acpi_power_meter, fix hpwdt name typo

sulemandiwan:
Added the blacklist for the acpi_power_meter kernel module that generates the ACPI errors due to a firmware bug (fixable but not something we use so no need). The errors fill logs and clutter console displays.

Also fixed the typo of hpwdt (it was kpwdt).


sulemandiwan:
I think I got them all. The tests passed. I can squash, merge, and push it whenever you give the final blessing but not any sooner than tomorrow morning.

PR #272, opened 2020-03-13
[WIP] AEIM-2663 quod dev

sulemandiwan:
Initial quod-dev profile for tang & co. and some quod ground work.


manifests/profile/quod/perl.pp
```diff
@@ -0,0 +1,284 @@
...
+    'libxml-simple-perl',
+    'libxml-twig-perl',
+    'libxml-writer-perl',
+    'libxml-xpath-perl',
+    'libxml-xpathengine-perl',
+    'libyaml-appconfig-perl',
+    'libyaml-libyaml-perl',
+    'libyaml-perl',
+    'libyaml-syck-perl',
+    'xmlformat-perl']:
```
daaang: Where did this list come from? Why is it different from "dependencies"?

manifests/profile/quod/perl.pp
```diff
@@ -0,0 +1,284 @@
...
+  }
+
+  -> nebula::cpan { [
```
daaang: Do these 13 nebula::cpan packages really depend on all 245 system packages?

manifests/profile/quod/dependencies.pp
```diff
@@ -0,0 +1,28 @@
+# Copyright (c) 2019 The Regents of the University of Michigan.
```
daaang: 2020

manifests/profile/quod/perl.pp
```diff
@@ -0,0 +1,284 @@
+
+# Copyright (c) 2019 The Regents of the University of Michigan.
```
daaang: 2020

manifests/role/app_host/quod_dev.pp
```diff
@@ -0,0 +1,23 @@
+# Copyright (c) 2018 The Regents of the University of Michigan.
```
daaang: 2020

manifests/profile/quod/perl.pp
sulemandiwan: They are based off of what is on the machine. And we don't know the dependency chains.

manifests/profile/quod/perl.pp
sulemandiwan: These are perl packages. Dependencies are prerequisites for these packages as well as for anything else that needs to be installed.

The layout is based on what we have in www_lib.

sulemandiwan: And example would be the oracle auth which apparently is a pre-requisite for some perl packages from what I gathered.

PR #275, opened 2020-03-30
Disable ingest restart jobs

aelkiss:
In rare cases ingest can get stopped in such a way that material ends up
in the repository even if it is incomplete. In the worst case, a zip
file can be put into place and appear intact, but in fact be incomplete
(see HT-1666)


botimer:
We'll want a PR to remove this ensure => absent. Thinking longer term, this kind of thing is probably a use case for Bolt.

PR #277, opened 2020-03-30
Add quod_prod role, quod_dev/quod_prod/common perl profiles

sulemandiwan:


skorner:
@sulemandiwan You probably need to add spec support for the quod_prod role.

sulemandiwan:
@daaang Could you have a peek at the code to see if it is all good. Once it is blessed I will squash and merge it into master. Thanks.

manifests/profile/quod/dev/perl.pp
```diff
@@ -0,0 +1,90 @@
...
+    'libwww-curl-perl',
+    'libxml-twig-perl',
+    'libxml-xpathengine-perl',
+    'libxml-xpath-perl',
+    'libyaml-appconfig-perl',
+    'libyaml-libyaml-perl',
+    'xmlformat-perl']:
+  }
+
+  -> nebula::cpan { [
```
daaang, line 85: Do these CPAN resources really depend on every single one of these packages?

manifests/profile/quod/prod/perl.pp
```diff
@@ -0,0 +1,44 @@
...
+    'libpod-readme-perl',
+    'libregexp-common-perl',
+    'libsoftware-license-perl',
+    'libterm-ui-perl',
+    'libtest-tester-perl',
+    'libtext-soundex-perl',
+    'libtext-template-perl']:
+  }
+
+  -> nebula::cpan { [
```
daaang, line 37: Same question as before; do we really need every single one of those packages installed before we can run `cpan -i` on these?

spec/classes/all_roles_spec.rb
```diff
@@ -65,7 +65,8 @@ def compile_along_with_all_roles(hiera_fixture)
             %w[nebula::role::webhost::www_lib_vm www_lib],
             %w[nebula::role::docker_registry docker_registry],
             %w[nebula::role::archivematica archivematica],
-            %w[nebula::role::app_host::quod_dev quod_dev],
+            %w[nebula::role::app_host::quod_dev quod],
+            %w[nebula::role::app_host::quod_prod quod],
```
daaang, line 69: Nice.

manifests/profile/quod/dev/perl.pp
botimer: Yeah, I guess I might clarify into two questions...

1. Do we know why we are using the debs for these, given that we have a CPAN resource type?
2. Do we need the dependency/order relationship for this to work?

I would hazard that the first is a bit of "it worked before" and the second is some collision between system/package-tracked dependencies and CPAN-tracked dependencies unless we order the installs... but it would be good to sort this out.

It is a _lot_ of cpan-as-deb packages, and I'm wondering if we could do them all as `nebula::cpan` or use [Carton](https://metacpan.org/pod/Carton) to normalize them.

manifests/profile/quod/prod/perl.pp
sulemandiwan: I am basing this off of www_lib format to some degree.

Also, I am not sure which of the 200+ packages are prerequisites for the CPAN modules. 

Is there an issue if these are ordered? Cause find chain of dependencies can quite tedious and time consuming without much benefit normally.

manifests/profile/quod/dev/perl.pp
sulemandiwan: I am basing this off of www_lib format to some degree.

Also, I am not sure which of the 200+ packages are prerequisites for the CPAN modules.

Is there an issue if these are ordered? Cause finding chains of dependencies can quite tedious and time consuming without much benefit normally.

sulemandiwan: Part of the reasoning at least is "it works now as is so let's duplicate what is and not what it could be." Currently we should probably not change this and refactor later on I think if we feel it is important.

And I think @aelkiss or @skorner may have a better idea why we have some as apt packages and some as cpan and whether that can or should be changed. Or at least visiting at this point or in the future.

I think it is a design decision honestly. Error on deb packages for stability and a primary point of management and only the exceptions should be CPAN. 

Carton could be something to consider. But is that adding unnecessary complexity or do we gain from it is something to visit also in the future.

daaang: I guess I'm interested to see what happens if we remove the `->`. I'm completely ok with some packages requiring other packages, but 3 CPAN installs, each requiring 67 apt packages, is a little over the top. For example, what is the worst thing that could happen? Is it that puppet attempts to run these out of order, some succeed while others fail, but then they all succeed 30 minutes later at the next sync?

Or are we worried about something breaking more seriously? What I'd like to avoid is the paranoid copying and pasting of configurations that none of us understand, since all it leads to is the continued copying and pasting of configurations that we may or may not need.

In other words, if we need that `->`, I would like for us to understand why we need it.

aelkiss: Our general principle was to install perl packages from Debian if they were available and use CPAN only if there are not upstream Debian packages.

If some of the CPAN packages were installed before some of the apt packages, then they will automatically bring in their dependencies as CPAN packages; whereas if we have the apt packages installed first they will use the apt packages. For www-lib and HathiTrust I did not particularly want to keep track of the dependencies for perl modules, I just want to ensure that all the Debian-provided packages are installed before the CPAN ones.

While we are using the system perl modules -- rather than going to all-CPAN I would actually argue that it might be better to use `dh-make-perl` to create debian packages that are installed in our local Debian archive. At the least, that way we have more consistent control over versioning -- if we go all-CPAN, then as it is now you will just get whatever random version happens to be released when we first set up a machine and then they'll never be upgraded.

I agree carton would be nice, and I think it's likely I'll look into that for the HT apps that we aren't immediately migrating off perl.

daaang: I see; that makes sense. In general I'm in favor of choosing one or the other, but in the meantime, perhaps a single line

```puppet
Package<| |> -> Nebula::Cpan<| |>
```

somewhere that applies to every system with perl could stand for all?

botimer: I've just added an ADR entry to the PR to capture the discussion above. As Aaron has mentioned elsewhere, this is a long-standing preference for OS-packaged versions where available. The preferred model may be quite different in containers but, for system packages, this is our best option.

I have no position on the Puppet ordering mechanics / syntax.

botimer: @daaang, do you consider this and the following comments resolved?

sulemandiwan: I do like the ordering default given that likely that is our model for approaching things until we start making our own debs always. 

It may be a good thing to do next sprint after we get the upgrades done next week.

I would like to do it as I haven't done something like that before.

daaang: Sounds good.

PR #281, opened 2020-04-07
AEIM-2605 update mocha 1 to debian 9

sulemandiwan:


skorner:
For the fully puppetized www_lib front end, the haproxy binding support was added to the new role rather than making a new profile. Given we are not creating quod VMs with puppet yet, I don't think it matters too much right now where we include the haproxy support for quod. I'll defer to @daaang

PR #283, opened 2020-04-13
AEIM-2604 latte-1 upgrade

sulemandiwan:


manifests/profile/networking/firewall/http.pp
```diff
@@ -1,13 +1,20 @@
-# Copyright (c) 2018 The Regents of the University of Michigan.
+# Copyright (c) 2020 The Regents of the University of Michigan.
```
daaang, line 1: Having correct but redundant copyright dates on files that already have a rich git history is very hard to care about, so I'm not requesting a change here, but in general I've seen these dates used to show when a file has changed. So `copyright 2015` means the file hasn't changed since 2015, `copyright 2015, 2018` means it was created in 2015 and later modified in 2018, while `copyright 2015-2018` means the file was undergoing constant change in 2015, 2016, 2017, and 2018.

So in this case, rather than change `2018` to `2020` (implying the file was created in 2020), the syntax would be to change from `2018` to `2018, 2020` (created in 2018, left alone in 2019, and modified in 2020).

But I've yet to see us have a clear policy around that, nebula is full of incorrect uses, and, as I said, there's also a git history, so my intent with this comment is purely to be informative.

manifests/profile/tsm.pp
```diff
@@ -1,4 +1,4 @@
-# Copyright (c) 2019 The Regents of the University of Michigan.
+# Copyright (c) 2020 The Regents of the University of Michigan.
```
daaang, line 1: Likewise, this would go from `2019` to `2019-2020`. If modified in 2021, it'd become `2019-2021`, but if left alone in 2021 only to be modified in 2022, it'd become `2019-2020, 2022`.

manifests/profile/tsm.pp
```diff
@@ -39,7 +42,8 @@
   Array[String] $inclexcl = ['exclude /.../.nfs*', 'exclude.dir /.../.snapshot'],
   Array[String] $domains = ['/etc','/opt','/var'],
   Array[String] $virtualmountpoints = ['/etc','/opt','/var'],
-  Array[String] $exclude_dirs = ['/afs/','/net/','/nfs/','/usr/vice/cache/']
+  Array[String] $exclude_dirs = ['/afs/','/net/','/nfs/','/usr/vice/cache/'],
+  Array[String] $opt_settings = ['* No custom settings']
```
daaang, line 46: Given the default comment, would `$custom_settings` be a better name? `opt` feels needlessly abbreviated to me, while `optional` collides a bit with puppet's `Optional` syntax, but you say `* No custom settings` right there in the comment, so why not call it that?

spec/classes/profile/tsm_spec.rb
```diff
@@ -61,13 +62,19 @@
...
           )
         end
 
         it do
           is_expected.to contain_file(dsm_opt)
             .with_content(%r{^DOMAIN "/baz"$})
             .with_content(%r{^DOMAIN "/quux"$})
+            .with_content(%r{^first_setting first_value$})
+            .with_content(%r{^second_setting "second_value"$})
         end
```
daaang, line 78: I'd consider splitting this into two tests, as it's testing two different behaviors (domains and opt_settings). For example, if you had

```ruby
it "adds domain settings to the dsm config file" do
  is_expected.to contain_file(dsm_opt)
    .with_content(%r{^DOMAIN "/baz"$})
    .with_content(%r{^DOMAIN "/quux"$})
end

it "adds custom settings to the dsm config file" do
  is_expected.to contain_file(dsm_opt)
    .with_content(%r{^first_setting first_value$})
    .with_content(%r{^second_setting second_value$})
end
```

The advantage of doing it this way is that the errors will be clearer in the future if one of these tests fails. It also shows that, while they look similar, they serve different purposes. If at a later date we ever wanted to change the domain behavior, for example, we wouldn't have to untangle which parts of the test test what.

Not telling you to do this, by the way; I think much of the usefulness is just in considering it. Writing good tests is always a complex web of trade-offs.

manifests/profile/networking/firewall/http.pp
sulemandiwan: Is there a way to automate it? Cause that would help consistency and reduce tedium.

manifests/profile/tsm.pp
sulemandiwan: I thought about that. But there are two types of custom settings. Ones that go into "dsm.opt" and ones that go into "dsm.sys". So my intent was respective opt_settings and sys_settings (latter isn't there cause we don't have any custom setting for that yet).

spec/classes/profile/tsm_spec.rb
sulemandiwan: I do appreciate the info. It helps in learning and getting better. I can split it up; makes sense to do.

sulemandiwan:
Okay separated out the tests. And did minor cleanup of dsm.sys config file (mainly make it look neater).

PR #284, opened 2020-04-15
AEIM-2750 - Create roles for formerly manual_ruby hosts

daaang:


manifests/role/hathitrust/dev/app_host_and_ruby.pp
```diff
@@ -0,0 +1,9 @@
+# Copyright (c) 2020 The Regents of the University of Michigan.
+# All Rights Reserved. Licensed according to the terms of the Revised
+# BSD License. See LICENSE.txt for details.
+
+# Grog
+class nebula::role::hathitrust::dev::app_host_and_ruby {
```
rrotter, line 6: The dev::app_host role only applies to ht-web-testing and (until this change is applied) grog. I don't know how ht-web-testing is an "App Host" when it lacks ruby.

Alternatively dev::app_host could get nebula::profile::ruby and we would add a dev::app_host_without_ruby or similar.

manifests/role/umich_and_ruby.pp
```diff
@@ -0,0 +1,9 @@
+# Copyright (c) 2020 The Regents of the University of Michigan.
+# All Rights Reserved. Licensed according to the terms of the Revised
+# BSD License. See LICENSE.txt for details.
+
+# Frobscottle
+class nebula::role::umich_and_ruby {
```
rrotter, line 6: Thumbs up. This should eventually roll into more specific roles as they emerge.

skorner: One could also make a datamart role that for now is just umich and ruby.

manifests/role/hathitrust/dev_and_ruby.pp
```diff
@@ -0,0 +1,9 @@
+# Copyright (c) 2020 The Regents of the University of Michigan.
+# All Rights Reserved. Licensed according to the terms of the Revised
+# BSD License. See LICENSE.txt for details.
+
+# Buzz
```
skorner, line 5: I'm fine with this, but buzz is essentially a HT dev app host if we want to go that route now.

manifests/role/hathitrust/prod_and_ruby.pp
```diff
@@ -0,0 +1,9 @@
+# Copyright (c) 2020 The Regents of the University of Michigan.
+# All Rights Reserved. Licensed according to the terms of the Revised
+# BSD License. See LICENSE.txt for details.
+
+# Slurpee-3, squishee-3
```
skorner, line 5: These are essentially HT prod app hosts, so we could use that moniker now.

manifests/role/mgetit_log.pp
```diff
@@ -12,4 +12,5 @@
   include nebula::profile::named_instances
   include nebula::profile::nodejs
   include nebula::profile::php73
+  include nebula::profile::ruby
```
skorner, line 15: Before this is merged we need to be sure that the torquebox setup on malt is insulated from the change. This was a factor in the past when we tried to apply puppet-managed ruby to malt. I believe a change was made on malt to insulate TB, but we should verify.

manifests/role/hathitrust/dev/app_host_and_ruby.pp
```diff
@@ -0,0 +1,9 @@
+# Copyright (c) 2020 The Regents of the University of Michigan.
+# All Rights Reserved. Licensed according to the terms of the Revised
+# BSD License. See LICENSE.txt for details.
+
+# Grog, buzz
```
daaang, line 5: Giving buzz `nebula::role::hathitrust::dev::app_host_and_ruby` instead of `nebula::role::hathitrust::dev` will bring in both the `ruby` profile and the `named_instances` profile. Not sure whether adding the named_instances profile will cause problems.

manifests/role/umich_and_ruby.pp
daaang: Done.

manifests/role/hathitrust/dev_and_ruby.pp
daaang: Done, but might cause problems. I haven't looked at the named_instances profile yet to see if adding it to buzz would be a problem.

manifests/role/hathitrust/prod_and_ruby.pp
daaang: Done.

manifests/role/mgetit_log.pp
daaang: Good to know

manifests/role/hathitrust/dev_and_ruby.pp
daaang: Update: we looked through it, it looks ok.

manifests/role/mgetit_log.pp
daaang: We're not messing with malt after all.

manifests/role/hathitrust/dev/app_host.pp
```diff
@@ -9,4 +9,5 @@
 class nebula::role::hathitrust::dev::app_host {
   include nebula::role::hathitrust::dev
   include nebula::profile::named_instances
+  include nebula::profile::ruby
```
daaang, line 12: Now buzz and grog can be regular app hosts, because ht-web-testing no longer depends mysteriously on this role.

manifests/role/hathitrust/dev/app_host_and_ruby.pp
daaang: Got rid of the `_and_ruby` by removing ht-web-testing's requirement entirely.

PR #289, opened 2020-04-21
AEIM-2768 - Uninstall unwanted ruby versions

daaang:
This will check every already-installed version of ruby and compare it
against the list of supported versions. If a version of ruby is
installed but not supported, it will be uninstalled.

### DIFF
--- a/manifests/profile/ruby.pp
+++ b/manifests/profile/ruby.pp
@@ -1,4 +1,4 @@
-# Copyright (c) 2018 The Regents of the University of Michigan.
+# Copyright (c) 2018, 2020 The Regents of the University of Michigan.
 # All Rights Reserved. Licensed according to the terms of the Revised
 # BSD License. See LICENSE.txt for details.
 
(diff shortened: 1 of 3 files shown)


manifests/profile/ruby.pp
```diff
@@ -85,4 +85,20 @@
       }
     }
   }
+
+  $::ruby_versions.each |$version| {
+    unless $version in $supported_versions {
+      unless $version == $global_version {
+        exec { "rbenv uninstall ${version}":
+          command     => "rbenv init -; rbenv uninstall ${version}",
```
botimer, line 93: Does this need to be `eval "$(rbenv init -)"` or equivalent?

daaang:
This shouldn't be merged until it's been tested with a development environment.

manifests/profile/ruby.pp
botimer: @daaang ^?

daaang: It was chicken-waving, so I just got rid of it.

PR #290, opened 2020-04-30
AEIM-2592 - Production kubernetes

daaang:
This change is not backwards compatible and will break hatcher-kubernetes—merge with care.


daaang:
> This change is not backwards compatible and will break hatcher-kubernetes—merge with care.

This is no longer the case. It's worth double-checking, but I'm pretty sure merging this will be safe.

spec/classes/profile/kubernetes/router_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+
+      it do
+        is_expected.to contain_firewall('002 Give external requests our public IP')
+          .with_table('nat')
+          .with_chain('POSTROUTING')
+          .with_jump('SNAT')
+          .with_proto('all')
+          .with_source('172.28.0.0/14')
+          .with_tosource('10.0.0.1')
+      end
```
botimer, line 40: This example is named a little strangely compared with the fixture data of 10.x "public IPs", but checks the right thing.

templates/profile/kubernetes/bootstrap/keys_03_functions.sh.erb
```diff
@@ -0,0 +1,199 @@
...
+  local ca_base="$1"
+  local crt_base="$2"
+  local subject="$3"
+  local single_ipaddress="$4"
+  local single_endpoint="$5"
+  local shared_ipaddress="$6"
+  local shared_endpoint="$7"
+  local kube_api="$8"
+
+  cat <<EOF > tmp.ini
```
botimer, line 121: We borrowed some of this from the gist to test -- we should trim it back to only what we need and cite the gist as a courtesy.

manifests/profile/kubernetes/register_for_keys/controller.pp
```diff
@@ -0,0 +1,14 @@
+# Copyright (c) 2020 The Regents of the University of Michigan.
+# All Rights Reserved. Licensed according to the terms of the Revised
+# BSD License. See LICENSE.txt for details.
+
+class nebula::profile::kubernetes::register_for_keys::controller {
+  $cluster_name = lookup('nebula::profile::kubernetes::cluster')
```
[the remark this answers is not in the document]
daaang: `nebula::profile::kubernetes::cluster` is node hiera; `nebula::profile::kubernetes::clusters` is common.yaml

spec/classes/profile/kubernetes/router_spec.rb
daaang: Yeah, I think that's an unwritten principle from way back, that all our example IP addresses should be within private ranges so that accidentally using example data won't cause data to be sent to some unknown actual server out there. Hence preferring `10.2.3.4` over `1.2.3.4`, etc.

Given how unlikely that is, since we've switched from actual hieradata defaults to fixtures that are only read while running tests, I'm not married to this. It's just a habit.

templates/profile/kubernetes/bootstrap/keys_03_functions.sh.erb
daaang: Good point.

PR #291, opened 2020-05-04
Added umich_mailserver role and profiles for postfix on AEIM-1976

umjnega:
Ready for review from Matt.

### DIFF
--- a/manifests/role/umich_mailserver.pp
+++ b/manifests/role/umich_mailserver.pp
@@ -1,0 +1,35 @@
+# Copyright (c) 2020 The Regents of the University of Michigan.
+# All Rights Reserved. Licensed according to the terms of the Revised
+# BSD License. See LICENSE.txt for details.
+
+# A minimum configuration for a mail server host that logically belongs 
+# the University of Michigan. 
+# 
+#
+# @example
+#   include nebula::role::umich_mailserver
+class nebula::role::umich_mailserver (
+  $bridge_network = false,
+  $internal_routing = '',
+) {
+
+  class { 'nebula::role::minimum':
+    internal_routing => $internal_routing,
+  }
+
+  if $facts['os']['family'] == 'Debian' and $::lsbdistcodename != 'jessie' {
+    include nebula::profile::duo
+    include nebula::profile::postfix
+    include nebula::profile::grub
+    include nebula::profile::ntp
+    include nebula::profile::tiger
+    class { 'nebula::profile::networking':
+      bridge => $bridge_network,
+    }
+  }
+
+  include nebula::profile::dns::standard
+  include nebula::profile::elastic::metricbeat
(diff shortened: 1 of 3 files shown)


manifests/role/umich_mailserver.pp
```diff
@@ -0,0 +1,35 @@
+# Copyright (c) 2020 The Regents of the University of Michigan.
+# All Rights Reserved. Licensed according to the terms of the Revised
+# BSD License. See LICENSE.txt for details.
+
+# The minimum configuration for a host that logically belongs to the
+# University of Michigan. Most--but not all--other roles rely on this
+# one.
```
botimer, line 7: Copypasta error

manifests/role/umich_mailserver.pp
```diff
@@ -0,0 +1,35 @@
...
+class nebula::role::umich_mailserver (
+  $bridge_network = false,
+  $internal_routing = '',
+) {
+
+  class { 'nebula::role::minimum':
+    internal_routing => $internal_routing,
+  }
+
+  if $facts['os']['family'] == 'Debian' and $::lsbdistcodename != 'jessie' {
```
botimer, line 20: This check might not need to be here... I think we'll start seeing these fall away in the coming weeks, and I don't see us applying this role to any jessie hosts.

spec/classes/profile/postfix_spec.rb
```diff
@@ -0,0 +1,16 @@
...
+# All Rights Reserved. Licensed according to the terms of the Revised
+# BSD License. See LICENSE.txt for details.
+require 'spec_helper'
+
+describe 'nebula::profile::postfix' do
+  on_supported_os.each do |os, os_facts|
+    context "on #{os}" do
+      let(:facts) { os_facts }
+
+      it { is_expected.to compile }
```
botimer, line 13: I would add an expectation to contain the postfix package and not contain exim.

manifests/role/umich_mailserver.pp
daaang: Ha! I didn't even look at the comments, good catch

manifests/role/umich_mailserver.pp
daaang: A good point, since this is just for sherry, which is stretch

manifests/role/umich_mailserver.pp
umjnega: I can edit that so its not worded the same the umich role.

manifests/role/umich_mailserver.pp
umjnega: I'll pass on this one until I feel more comfortable with nebula.

spec/classes/profile/postfix_spec.rb
umjnega: Again,  going to note this down and pass until I gain more experience.

PR #292, opened 2020-05-05
AEIM-2672 Add oracle ora config files and relevant support

sulemandiwan:
- Some additional refactoring to use the newly added variables may be nice to do.
- There are corresponding details in the lot as well.

Extra:
------
Another possible option would be to use https://forge.puppet.com/biemond/oradb but this seems a bit heavy for the minimal client-only settings this needs.

### DIFF
--- a/spec/classes/role/www_lib_vm_spec.rb
+++ b/spec/classes/role/www_lib_vm_spec.rb
@@ -68,12 +68,6 @@ describe 'nebula::role::webhost::www_lib_vm' do
       it { is_expected.to contain_host('mysql-web').with_ip('10.0.0.123') }
 
       it do
-        # set via hiera
-        is_expected.to contain_file('authz_umichlib.conf')
-          .with_content(%r{DBDParams\s*user=somebody})
-      end
-
-      it do
         is_expected.to contain_apache__vhost('000-default-ssl')
           .with_aliases([{ 'scriptalias' => '/monitor',
                            'path' => '/usr/local/lib/cgi-bin/monitor' }])
@@ -262,6 +256,26 @@ describe 'nebula::role::webhost::www_lib_vm' do
                 ssl_cert: '/etc/ssl/certs/apps.lib.umich.edu.crt')
           .with_error_log_file('error.log')
           .with_custom_fragment(%r{CookieName skynet})
+      end
+
+      # Files
+      it { is_expected.to contain_file('sqlnet.ora') }
+
+      it 'adds custom params to file authz_umichlib.conf' do
+        is_expected.to contain_file('authz_umichlib.conf')
+          .with_content(%r{DBDParams\s*user=somebody})
+      end
+
+      it 'adds custom params to file tnsnames.ora' do
+        is_expected.to contain_file('tnsnames.ora')
+          .with_content(%r{^ORCL.MYSERVER1_ALIAS1\s+=})
+          .with_content(%r{^ORCL.MYSERVER1_ALIAS2\s+=})
+          .with_content(%r{^ORCL.MYSERVER2_ALIAS1\s+=})
+          .with_content(%r{^ORCL.MYSERVER2_ALIAS2\s+=})
+          .with_content(%r{HOST\s+=\s+myserver1.umdl.umich.edu})
+          .with_content(%r{HOST\s+=\s+myserver2.umdl.umich.edu})
+          .with_content(%r{SID\s+=\s+abcd})
+          .with_content(%r{PORT\s+=\s+1234})
       end
     end
   end
--- a/templates/profile/apache/tnsnames.ora.erb
+++ b/templates/profile/apache/tnsnames.ora.erb
@@ -1,0 +1,48 @@
+<% @oracle_servers.each do | server, server_aliases | -%>
+# <%= server %>
+#
+<% server_aliases.each do | server_alias | -%>
+<%= server_alias %> =
+  (DESCRIPTION =
+    (ADDRESS =
+      (PROTOCOL = TCP)
+      (HOST = <%= server %>.umdl.umich.edu)
+      (PORT = <%= @oracle_port %>)
+    )
+    (CONNECT_DATA =
+      (SERVER = DEDICATED)
+      (SERVICE_NAME = orcl.<%= server %>)
+    )
+  )
+
+<% end -%>
+
+<% end -%>
+
+ORCL.LB =
+  (DESCRIPTION =
+    (ADDRESS_LIST =
+      (FAILOVER = ON)
+      (LOAD_BALANCE = OFF)
+
+<% @oracle_servers.each do | server, server_aliases | -%>
+      (ADDRESS =
+        (PROTOCOL = TCP)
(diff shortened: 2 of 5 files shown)


templates/profile/apache/tnsnames.ora.erb
```diff
@@ -0,0 +1,48 @@
+<% @oracle_servers.each do | server, server_alias_arr | -%>
+# <%= server %>
+#
+<% server_alias_arr.each do | server_alias | -%>
+<%= server_alias.upcase %> =
+  (DESCRIPTION =
+    (ADDRESS =
+      (PROTOCOL = TCP)
+      (HOST = <%= server.downcase %>.umdl.umich.edu)
```
daaang, line 9: Is there a reason to enforce upcase and downcase here? For the principle of least surprising behavior, I'd recommend removing all `.upcase` and `.downcase` calls in favor of upcasing/downcasing the hieradata, so that what's in the hieradata will be copied faithfully into the configuration.

It'd be different if we were sometimes running `server.upcase` or `server_alias.downcase`, but I don't see that anywhere.

templates/profile/apache/tnsnames.ora.erb
```diff
@@ -0,0 +1,48 @@
+<% @oracle_servers.each do | server, server_alias_arr | -%>
```
daaang, line 1: Ruby programmers like to avoid adding types to variable names (e.g. `_arr`). There are a few reasons (e.g. it adds noise that is more likely to confuse than to inform, it can become misleading as the code is extended), but I think the most important one is that it reveals a variable name which could benefit from more thought.

While it's tempting to be clear with a name like `server_alias_arr`, and `server_aliases` (for example) gives less information, all you need to know happens a couple lines later with

```ruby
server_aliases.each do |server_alias|
```

When we see that, we know that `server_aliases` is something (perhaps a dynamically-generated enumerable, maybe a static array) that can be iterated on (i.e. with `.each`), which is all we need to know. This leaves the door open for us to pass anything that can iterate without potentially lying and claiming that it's an array.

tl;dr: I suggest renaming `server_alias_arr` to `server_aliases`

spec/classes/role/www_lib_vm_spec.rb
```diff
@@ -263,6 +257,26 @@
...
+
+      # Files
+      it { is_expected.to contain_file('sqlnet.ora') }
+
+      it 'adds custom params to file authz_umichlib.conf' do
+        is_expected.to contain_file('authz_umichlib.conf')
+          .with_content(%r{DBDParams\s*user=somebody})
+      end
+
+      it 'adds custom params to file tnsnames.ora' do
```
daaang, line 269: I'm glad to see you adding descriptions to the tests (as opposed to `it do`). We should do more of that.

spec/classes/role/www_lib_vm_spec.rb
```diff
@@ -263,6 +257,26 @@
           .with_error_log_file('error.log')
           .with_custom_fragment(%r{CookieName skynet})
       end
+
+      # Files
+      it { is_expected.to contain_file('sqlnet.ora') }
+
+      it 'adds custom params to file authz_umichlib.conf' do
+        is_expected.to contain_file('authz_umichlib.conf')
+          .with_content(%r{DBDParams\s*user=somebody})
```
daaang, line 266: Haha, I was about to comment, saying I was confused about why this was being deleted (took me a minute to realize you just moved and improved it). This is definitely better than

```ruby
it do
  # set via hiera
```

daaang:
I just noticed your comment about the oradb forge module. In theory, I think we should always endeavor to use existing tools like that, but, in practice, it's messier than it's worth to implement our current strange setups with existing tools (see for example all the www-lib apache vhosts).

If we decide to set up a new oracle database, however, I think that would be the time to use the oradb module, rather than manage our own lovingly-crafted config files.

Same with www-lib apache: I wish we'd just represented it with vhost files that we could configure with .erb files, but any **new** websites should definitely be set up with the apache module (or, better yet, the nginx module (or, better yet, kubernetes)).

templates/profile/apache/tnsnames.ora.erb
sulemandiwan: I agree with your use case for certain. That is the main use case and I did think about that in this context.

So server_alias needs to be uppercase and the server name lowercase at least by config file convention if not technical requirement.

What prompted the upcase and downcase was to compensate for human error and cognitive overload of having to remember that we need the server name to be lowercase but the server aliases to be uppercase. With this the code handles that consistency checking and fixing so we don't have to when adding node files.

templates/profile/apache/tnsnames.ora.erb
sulemandiwan: Thanks. :) Makes sense.

templates/profile/apache/tnsnames.ora.erb
sulemandiwan: Thoughts given that info?

Basically if I don't do the code then it will need us to remember to write stuff like:

myserver:
  - SERVERALIAS

And not any variation from that.

templates/profile/apache/tnsnames.ora.erb
sulemandiwan: Done.

templates/profile/apache/tnsnames.ora.erb
daaang: Thanks for the explanation. That makes sense, although I still think it's out of bounds for what puppet is for. Since we're talking about generating a configuration file, we should understand what goes in that file when we're writing configuration regardless of whether that configuration lives in a file on a server or in a yaml file.

I guess what I'm saying is that I think our hiera configuration has more value if we keep it faithful to what it comes out as. If we start adding modifiers, that adds harder-to-predict behavior that I think is worth it when we're removing configuration from hieradata or at least generalizing it.

If you're worried about lowercase letters in the configuration causing breaking changes, a better approach could be to add something that causes the profile to fail with an error instead of generating the file at all, like this:

```puppet
if <any of the aliases contain lowercase letters> {
  fail('Oracle server aliases must be all uppercase')
}
```

This way, we require our configuration to remain faithful to reality.

sulemandiwan: I agree. It wasn't fully settling with me either. 

I like your suggested fix better. :) I will look into this once I finish troubleshooting why it isn't working currently in testing.

daaang: If you want, you could also add a test with `it { is_expected.not_to compile }`

Sometimes it feels pretty good to have a "this shouldn't compile" test and watch it pass.

sulemandiwan:
Slight refactoring to use the variables added. 
Removed default for home as it seems to vary a lot (doesn't have a standard except install directory) so better to not have it I think.

sulemandiwan:
I reverted this branch to a state with no php finagling. The php changes are in a separate branch I have locally stored but I don't think we will need it cause the approach of keeping things separate should eliminate those changes. I will push that up later today after this is taken care of.

PR #298, opened 2020-05-18
AEIM-2815 - Rename kubernetes legacy::kubernetes

daaang:
This is to make room for its radically different production replacement,
since we want the prototype to continue running until we're ready to
replace it.

This requires a s/kubernetes/legacy::kubernetes/g in the hieradata that
definitely needs to be tested on hatcher-k8s-* before this is safe to
merge.


daaang:
Confirmed: the entire prototype cluster has been moved into this development environment, and it had no effect whatsoever on the nodes themselves. This is safe to merge.

PR #306, opened 2020-06-09
AEIM-2849 - www-lib crons

daaang:


skorner:
The three Torquebox cron jobs can be purged forever. 

The shib check is only meaningful for MACC VMs, as only Fulcrum uses shib currently. I think it's fine for that cron to run anywhere though.

daaang:
I removed the torquebox jobs, and I moved the cronjob-by-title tests into the role specs so that they could pick up the GeoIP and shib checks from the profiles that already manage those cronjobs.

manifests/profile/www_lib/cron.pp
```diff
@@ -0,0 +1,65 @@
...
+    'purge apache access logs 1/2':
+      hour    => 1,
+      minute  => 7,
+      command => '/usr/bin/find /var/log/apache2 -type f -mtime +14 -name "*log*" -exec /bin/rm {} \; > /dev/null 2>&1',
+    ;
+
+    'purge apache access logs 2/2':
+      hour    => 1,
+      minute  => 17,
+      command => '/usr/bin/find /var/log/apache2 -type f -mtime +2  -name "*log*" ! -name "*log*gz" -exec /usr/bin/pigz {} \; > /dev/null 2>&1',
```
botimer, line 55: I think we need to ensure_package pigz for this to work... but I would swear that logrotate is already doing gz after a day.

PR #311, opened 2020-06-15
AEIM-2857 search lib mirlyn lib adjust

sulemandiwan:
Adding a location stanza to search-http also (before only in https).
Adding NE to mirlyn vhost settings.


sulemandiwan:
Merging these adjustments. The bring the files more inline.

PR #312, opened 2020-06-16
AEIM-2862 Add med.lib vhost w/ ip restriction param

sulemandiwan:


sulemandiwan:
Needs (sanity) testing in addition to review as I am getting a forbidden error in both testing and production.

manifests/profile/www_lib/vhosts/med.pp
```diff
@@ -0,0 +1,100 @@
...
+#
+# @example
+#   include nebula::profile::www_lib::vhosts::med
+class nebula::profile::www_lib::vhosts::med (
+  Array[String] $ip_restrict,
+  String $docroot = '/hydra/dromedary-production/current/public',
+) {
+  $servername = 'med.lib.umich.edu'
+
+  file { "${apache::params::logroot}/med":
```
botimer, line 17: This looks like it might be out of sync with the vhost below.

sulemandiwan: Good catch. Thanks.

PR #320, opened 2020-06-24
AEIM-2779 - Fix up directory/location directives

botimer:
- Remove the access control for `/cgi/l/login`, which was breaking
 - Move the LocationMatch for SALI to the right section, public ON


botimer:
Note: this diff looks a little weird, like it's moving the docroot directives, but it's really moving the SALI LocationMatch from `cosign_public_access_off_dirs` to `directories` and removing the cgi access controls from `directories`.

PR #325, opened 2020-06-24
Fix up Shibboleth config and first-application dependencies

botimer:


manifests/profile/apache.pp
```diff
@@ -69,13 +69,15 @@
...
   }
 
   file { '/etc/apache2/conf-available':
-    ensure => 'absent',
-    force  => true,
-    purge  => true
+    ensure  => 'absent',
+    force   => true,
+    purge   => true,
+    require => Class['apache'],
```
daaang, line 80: This confuses me, but it doesn't look like it could cause any harm, so it's fine. I just don't see why requiring a directory be absent would need to require the apache class.

botimer: Yeah, that one did not throw an error. Just put it in for parity.

PR #329, opened 2020-06-25
Switch to shorter startup_timeout value

sulemandiwan:
Switching to the updated timeout value as per @skorner.


botimer:
This is old, but the default is still 900, so this looks good, taking out the override.

PR #338, opened 2020-06-29
AEIM-2878 - Add keys to prometheus scrapers

daaang:


daaang:
Tested on hatcher-prometheus, working just fine

PR #342, opened 2020-06-30
Get service list from hieradata

daaang:
This way, we can set in hatcher:

    nebula::profile::www_lib::register_for_load_balancing::services:
    - www-lib

And at the macc:

    nebula::profile::www_lib::register_for_load_balancing::services:
    - www-lib
    - deepblue


daaang:
There are hiera changes that must be made before this is merged. They're currently pushed to remove_deepblue_separate_roles in lens.

PR #351, opened 2020-07-09
AEIM-2929 press web logging path and other activities

sulemandiwan:


sulemandiwan:
The comments for the squashed version:

- Fixed press logging path
- Switched to using logging_prefix variable
- Add log rotation as a server default
- Add log rotation to press vhost

PR #355, opened 2020-07-14
AE-7307 Add support for imgsrv logrotate, change log path

sulemandiwan:


manifests/profile/hathitrust/imgsrv.pp
```diff
@@ -13,8 +13,24 @@
   String $sdrroot,
   String $sdrview,
   String $sdrdataroot,
-  String $bind
+  String $bind,
+  String $logging_prefix = 'imgsrv'
 ) {
+  $log_path = "${apache::params::logroot}/${logging_prefix}"
```
botimer, line 19: This `apache::params::logroot` seems a little sketchy. It's okay for now because we know that those nodes serve both purposes, but it's a bit of incidental coupling. Realistically, I think we will break these out into containers before this changes in puppet config, so no problem -- just noting it for reference purposes.

botimer: I should say: the implicit dependency on `class { 'apache' }` with no fallback is the "sketchy" part.

PR #366, opened 2020-07-30
CVE-2020-10663 - Upgrade json gem

daaang:
This isn't urgent, since it only affects our testing environment, but a
security vulnerability is a security vulnerability, and there's no harm
in keeping our dependencies up to date.


botimer:
Handled in other merges.

PR #375, opened 2020-08-24
AEIM-3022 Expose Graylog via GELF/TCP to datacenter nodes

botimer:
This is straightforward; it adds an HAProxy pair for GELF similar to how
we support https_alt.

### DIFF
--- a/manifests/profile/kubernetes/destination_port/gelf_tcp.pp
+++ b/manifests/profile/kubernetes/destination_port/gelf_tcp.pp
@@ -1,0 +1,14 @@
+# Copyright (c) 2020 The Regents of the University of Michigan.
+# All Rights Reserved. Licensed according to the terms of the Revised
+# BSD License. See LICENSE.txt for details.
+
+class nebula::profile::kubernetes::destination_port::gelf_tcp {
+  $cluster_name = lookup('nebula::profile::kubernetes::cluster')
+
+  @@concat_fragment { "haproxy kubernetes gelf tcp ${::hostname}":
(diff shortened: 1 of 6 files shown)


botimer:
Thanks, I definitely missed that. Added and rebased.

manifests/profile/kubernetes/destination_port/gelf_tcp.pp
```diff
@@ -0,0 +1,14 @@
...
+# All Rights Reserved. Licensed according to the terms of the Revised
+# BSD License. See LICENSE.txt for details.
+
+class nebula::profile::kubernetes::destination_port::gelf_tcp {
+  $cluster_name = lookup('nebula::profile::kubernetes::cluster')
+
+  @@concat_fragment { "haproxy kubernetes gelf tcp ${::hostname}":
+    target  => '/etc/haproxy/services.d/gelf_tcp.cfg',
+    order   => '02',
+    content => "  server ${::hostname} ${::ipaddress}:12201 check\n",
```
daaang, line 11: This 12201 should instead be whatever nodeport you'll be using.

spec/classes/profile/kubernetes/destination_port_spec.rb
```diff
@@ -8,6 +8,7 @@
 [
   ['api',        6443],
   ['etcd',       2379],
+  ['gelf_tcp',  12201],
```
daaang, line 11: This should also be a nodeport instead of the external facing port

PR #378, opened 2020-09-01
added info for yarn repo

umjnega:


manifests/role/app_host/prod.pp
```diff
@@ -13,4 +13,5 @@
   include nebula::profile::ruby
   include nebula::profile::nodejs
   include nebula::profile::named_instances
+  include nebula::profile::yarn
```
daaang, line 16: I know we want this on bulleit-1, but do we also want it on bulleit-2, julep-1, juice, hatcher-apps-001 and macc-apps-001? If not, we should split out a new role just for bulleit-1.

manifests/role/app_host/dev.pp
```diff
@@ -17,4 +17,5 @@
   include nebula::profile::ruby
   include nebula::profile::nodejs
   include nebula::profile::named_instances
+  include nebula::profile::yarn
```
daaang, line 20: Along the same line, we want this (and it's already installed) on nectar, but do we also want it on punch, champagne, and search-prep?

manifests/profile/yarn.pp
```diff
@@ -0,0 +1,24 @@
...
+    key          =>  {
+      id     => 'B90F6449FEBC20F00DB13ED8212659B22565CA86',
+      source => 'https://dl.yarnpkg.com/debian/pubkey.gpg'
+    },
+    release      => 'stable',
+    repos        => 'main',
+    architecture => $::os['architecture'],
+  }
+
+  package { ['yarn']: }
```
daaang, line 22: Might be worth adding a requirement here, since the package requires the apt repo, like this

```puppet
package { 'yarn':
  require => Apt::Source['yarn'],
}
```

This will ensure that the puppet agent doesn't try to apt-get install yarn until after it's added the new repo.

umjnega: Ah yes, we did talk about this.  Thanks for the example.  I agree, I like this approach.  I will change it.

PR #385, opened 2020-09-10
add rewrite condition and rule to redirect

umjnega:
Not sure I have the rewrite condition correct.   But its a start.

### DIFF
--- a/manifests/profile/www_lib/vhosts/deepblue.pp
+++ b/manifests/profile/www_lib/vhosts/deepblue.pp
@@ -74,6 +74,10 @@ class nebula::profile::www_lib::vhosts::deepblue (
         rewrite_cond => '%{REQUEST_URI} !^((\/?|/index.html)$|/splash/|/cosign/valid)',
         rewrite_rule => '^(.*)$	http://bulleit-2.umdl.umich.edu:8080$1 [P]'
       },
+      {
+        comment      => 'Deep Blue Preservation redirect',
+        rewrite_rule => '^/static/about/deepbluepreservation.html https://www.lib.umich.edu/about-us/policies/digital-repository-services-digital-preservation-policy/registered-formats-and [R=permanent,L]'
+      },
     ],
 
     directories                   => [


manifests/profile/www_lib/vhosts/deepblue.pp
```diff
@@ -74,6 +74,11 @@
         rewrite_cond => '%{REQUEST_URI} !^((\/?|/index.html)$|/splash/|/cosign/valid)',
         rewrite_rule => '^(.*)$	http://bulleit-2.umdl.umich.edu:8080$1 [P]'
       },
+      {
+        comment      => 'Deep Blue Preservation redirect',
+        rewrite_cond => '^(.*)$ https://deepblue.lib.umich.edu/static/about/deepbluepreservation.html [P]',
```
botimer, line 79: We won't need a condition here, just the path portion moved to the pattern in the rule.

manifests/profile/www_lib/vhosts/deepblue.pp
```diff
@@ -74,6 +74,11 @@
...
+        rewrite_rule => '^(.*)$	https://www.lib.umich.edu/about-us/policies/digital-repository-services-digital-preservation-policy/registered-formats-and [P]'
```
botimer, line 80: The flags should be `[R=permanent,L]`, which makes it a (permanent) 301 [R]edirect and makes this the [L]ast rule in the evaluation -- immediate client redirect.

The pattern won't need any matching groups since it's a literal redirect. It should have the head anchor, though. That is, I think `^/static/about/deepbluepreservation.html` is about the best pattern.

botimer (changes requested):

manifests/profile/www_lib/vhosts/deepblue.pp
umjnega: Got it, change rewrite_cond to path....

manifests/profile/www_lib/vhosts/deepblue.pp
umjnega: I think I have this corrected now.  cleaned up that spacing gap too. (probably a tab)

botimer (approved):
Looks good. Clean and squashed.

PR #390, opened 2020-09-27
Rotate wtmp and btmp logs weekly instead of monthly

daaang:
Sometimes wtmp fills up disks on our more thinly-provisioned servers. By
setting this to weekly instead of monthly, we expect that the log files
will take up less disk space.

### DIFF
--- a/manifests/profile/logrotate.pp
+++ b/manifests/profile/logrotate.pp
@@ -12,7 +12,7 @@ class nebula::profile::logrotate {
   logrotate::rule {
     default:
       missingok    => true,
-      rotate_every => 'month',
+      rotate_every => 'week',
       create       => true,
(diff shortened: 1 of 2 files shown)


botimer:
This looks fine to me, though I'm unaware of this being an issue any time recently. What's the current opinion?

skorner:
It's likely that the problematic servers had the setting manually changed or the problem mitigated through partition cleanup. A more intentional and comprehensive local log rotation strategy and settings that takes into account whatever external log aggregation we have in place would be more satisfying if there is not immediate pressure for a one-off band-aid.

botimer:
Merged since this is here and records our intent. If the intent changes, updating/removing is easy.

PR #392, opened 2020-10-05
AE-7753 Add heartofdarknessarchive vhost / update redirects

sulemandiwan:


umjnega (approved):
I read this as .com .net redirect to .org.  The cert is matched with .com.  Looks good to me.

PR #394, opened 2020-10-09
[WIP] new www_lib_cron_vm role for www.lib maintenance

rrotter:
Create a VM that is a minimal subset of www.lib apache servers for running crons and maintenance.

Needs to support dnd's use of drush.


botimer:
This looks like a meaningful start for an automation server role... What is the status?

rrotter:
This will never be completed as is, but could be useful reference material when we pick up the need for www_lib devops machine in the future.

PR #396, opened 2020-10-14
1st commit for mysql prometheus exporter

umjnega:


daaang (commented):
Yeah this all looks good. I guess the next steps are to figure out which profiles/roles should include this new profile, and also to put all the exported resources in place. Nice!

manifests/profile/prometheus/exporter/mysql.pp
```diff
@@ -25,5 +25,12 @@
     content => template('nebula/profile/prometheus/exporter/mysql/defaults.sh.erb')
   }
 
+  @@concat_fragment { "prometheus mysql service ${::hostname}":
```
daaang, line 28: This will need to be imported by the prometheus profile—right now, it's being exported without being imported anywhere.

manifests/profile/prometheus/exporter/mysql.pp
```diff
@@ -25,5 +25,12 @@
...
+    tag     => "${::datacenter}_prometheus_mysql_service_list",
+    target  => '/etc/prometheus/mysql.yaml',
+    content => template('nebula/profile/prometheus/exporter/mysql/target.yaml.erb')
+  }
+
+  Firewall <<| tag == "${::datacenter}_prometheus_mysql_exporter" |>>
```
daaang, line 34: This will import all `firewall` resources with the tag `${::datacenter}_prometheus_mysql_exporter`, and currently no such resource is being exported anywhere, so it will find 0 results. (The prometheus profile needs to export it.)

templates/profile/prometheus/exporter/mysql/target.yaml.erb
```diff
@@ -1,5 +1,5 @@
-- targets: [ '<%= @ipaddress %>:9101' ]
+- targets: [ '<%= @ipaddress %>:9100' ]
```
daaang, line 1: Does the mysql exporter really use 9100? If so, we'll need to not use the default, because this port is already taken by the node exporter.

templates/profile/prometheus/exporter/mysql/target.yaml.erb
```diff
@@ -1,5 +1,5 @@
...
   labels:
     datacenter: '<%= @datacenter %>'
     hostname: '<%= @hostname %>'
-    priority: '<%= @master ? 'primary' : 'backup' %>'
+    role: '<%= @role %>'
```
daaang, line 5: I'm confused by this change. While `priority` is certainly not the right word, it does seem like it'd be useful to track whether a database is a master or slave (although, maybe even that is superfluous given our master-master architecture). In any case, I'm not sure there's much value in tracking a server's role, in the context of mysql, but I haven't given it much thought.

umjnega: I wasn't sure what to do in this spot.  I figured the priority was wrong, since all mysql servers are technically writable.   I defaulted to "role" without really understanding what it means.    Perhaps, nothing is necessary here?

templates/profile/prometheus/exporter/mysql/target.yaml.erb
umjnega: Ah, good catch.  I wasn't making that connection.   Port is actually 9104.  Will push with my next update.

manifests/profile/prometheus/exporter/mysql.pp
umjnega: I think i understand what was missing here.  Will wait till my next PR passes, to validate that with you.

daaang: Yeah you got it.

templates/profile/prometheus/exporter/mysql/target.yaml.erb
daaang: Hm yeah, maybe just delete the line. The main reason we're doing this is to detect replication errors that require our attention, and there isn't really an easy way to rate different servers—even if we're talking about some bronze-level service, replication errors are bad news in danger of getting worse and worse, so we want to resolve them anytime they happen.

With haproxy, there was a good reason to know whether something was primary/backup, since backup problems are only bad if the primary is also bad.

Actually, I've come full circle, and I back your decision to add the role (although I bet right now, as written, this template will always produce `role: ''`). I could easily imagine wanting to say, collect database metrics for www-lib, and that should hinge on a role, rather than e.g. needing to change every time we replace a database server.

manifests/profile/prometheus/exporter/mysql.pp
daaang: This still isn't being imported by prometheus—I'd look for `Concat_fragment <<|` and `concat_file` in the prometheus profile for examples to imitate. I think you'll also need to edit the prometheus config template file to tell it to read the resulting concat file.

manifests/role/app_host/dev_mysqlexp.pp
```diff
@@ -0,0 +1,16 @@
+# Copyright (c) 2019 The Regents of the University of Michigan.
```
daaang, line 1: 2020

manifests/role/app_host/prod_mysqlexp.pp
```diff
@@ -0,0 +1,16 @@
+# Copyright (c) 2019 The Regents of the University of Michigan.
```
daaang, line 1: 2020

manifests/role/aws_mysqlexp.pp
```diff
@@ -0,0 +1,12 @@
+# Copyright (c) 2020 The Regents of the University of Michigan.
+# All Rights Reserved. Licensed according to the terms of the Revised
+# BSD License. See LICENSE.txt for details.
+
+# Minimal aws with prometheus mysql exporter
+#
+# @example
+#   include nebula::role::aws
+class nebula::role::aws_mysqlexp {
```
daaang, line 9: Since there's already an aws directory, I'd move this to aws/mysqlexp.pp and call it aws::mysqlexp, to be consistent with other aws roles.

daaang (changes requested):
Nice! I'm not sure about mysqlexp as a name for this—maybe mysql_metrics? But other than that and a couple missing resources, this looks pretty much exactly right to me.

manifests/profile/prometheus/exporter/mysql.pp
umjnega: Sorted this out and made sure docker was mounting the file.

daaang (approved):
Yeah awesome.

PR #399, opened 2020-11-04
AEIM-3108 - Throttle with 429 instead of 509

daaang:
We've been serving 509s when throttling connections, but 5xx errors
imply something wrong with the server. The correct HTTP status code is
429, which correctly implies disallowed behavior on part of the client.

### DIFF
--- a/manifests/haproxy/service.pp
+++ b/manifests/haproxy/service.pp
@@ -64,11 +64,11 @@ define nebula::haproxy::service(
   $nonempty_whitelists = $whitelists.filter |$whitelist,$exemptions| { $exemptions.length > 0 }
 
   if $max_requests_per_sec > 0 {
-    file { "/etc/haproxy/errors/${service}509.http":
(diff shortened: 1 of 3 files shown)


manifests/haproxy/service.pp
```diff
@@ -64,11 +64,11 @@
   $nonempty_whitelists = $whitelists.filter |$whitelist,$exemptions| { $exemptions.length > 0 }
 
   if $max_requests_per_sec > 0 {
-    file { "/etc/haproxy/errors/${service}509.http":
+    file { "/etc/haproxy/errors/${service}429.http":
       ensure => 'present',
       mode   => '0644',
       notify => Service['haproxy'],
-      source => "https://${http_files}/errorfiles/${service}509.http"
+      source => "https://${http_files}/errorfiles/${service}429.http"
```
daaang, line 71: I created deepblue429.http and quod429.http on ae-files, so we should be covered.

PR #401, opened 2020-11-05
Add default mysql prometheus datacenter

daaang:


botimer:
It seems to me this is covered by other, more recent work. Is that true?

rrotter:
Stale. Closing.

PR #402, opened 2020-11-05
add mysql down alert

umjnega:


daaang (commented):
In the wild I don't think I've ever seen alert thresholds shorter than 5 minutes. I'd set the page alerts to 5m and the ticket one to 30m, at least to start with. Otherwise these look right to me!

If we want to add an experimental alert for behind, I'd start with a 5m ticket alert of `mysql_slave_status_master_server_id > 0 and ON (instance) (mysql_slave_status_seconds_behind_master - mysql_slave_status_sql_delay) > 300`. 5m to be intentionally noisy, ticket to be low stakes for now.

templates/profile/prometheus/rules.yml.erb
```diff
@@ -160,6 +160,33 @@ groups:
...
       summary: 'Dark Blue repository storage, {{$labels.device}} is out of available space.'
       description: '{{$labels.device}} has had less than 100GB of available space for more than 5 minutes.'
+  - alert: MysqlDown
+    expr: >
+      mysql_up == 0
+    for: 5m
+    labels:
+      severity: page
+    annotations:
+      summary: 'MySQL is DOWN on {{$lables.hostname}}'
```
botimer, line 170: labels

templates/profile/prometheus/rules.yml.erb
```diff
@@ -160,6 +160,33 @@ groups:
...
+      description: 'MySQL is DOWN on {{$labels.hostname}}'
+  - alert: MysqlSlaveSqlDown
+    expr: >
+      mysql_slave_status_slave_sql_running  == 0
+    for: 5m
+    lablels:
+      severity: page
+    annotations:
+      summary: 'MySQL Slave SQL is Not Running'
+      description: 'MySQL Slave SQL is not Running on {{$labels.hostname}}'
```
botimer, line 180: Second 'SQL' is redundant

templates/profile/prometheus/rules.yml.erb
```diff
@@ -160,6 +160,33 @@ groups:
...
+  - alert: MysqlMaxPreparedStatment
+    expr: >
+      mysql_global_status_prepared_stmt_count / mysql_global_variables_max_prepared_stmt_count > 0.75
+    for: 30m
+    lables:
+      severity: ticket
+    annotations:
+      summary: 'MySQL Prepared Statment Count is apparoaching MAX on {{$lables.hostname}}'
+      description: 'MySQL Prepared Statment Count is approaching MAX on {{$labels.hostname}}'
```
botimer, line 189: typos

templates/profile/prometheus/rules.yml.erb
```diff
@@ -160,6 +160,33 @@ groups:
...
+    labels:
+      severity: page
+    annotations:
+      summary: 'MySQL is DOWN on {{$lables.hostname}}'
+      description: 'MySQL is DOWN on {{$labels.hostname}}'
+  - alert: MysqlSlaveSqlDown
+    expr: >
+      mysql_slave_status_slave_sql_running  == 0
+    for: 5m
+    lablels:
```
botimer, line 176: labels

templates/profile/prometheus/rules.yml.erb
```diff
@@ -160,6 +160,33 @@ groups:
...
+      severity: page
+    annotations:
+      summary: 'MySQL Slave SQL is Not Running'
+      description: 'MySQL Slave SQL is not Running on {{$labels.hostname}}'
+  - alert: MysqlMaxPreparedStatment
+    expr: >
+      mysql_global_status_prepared_stmt_count / mysql_global_variables_max_prepared_stmt_count > 0.75
+    for: 30m
+    lables:
```
botimer, line 185: labels

botimer (changes requested):
Minor typo changes to make. The shape and variables look good.

templates/profile/prometheus/rules.yml.erb
daaang: For what it's worth, we don't use the description at all. We only use the summary, so if the slave isn't running on two different servers, we'll receive an alert that looks like this:

```
MySQL Slave SQL is Not Running
MySQL Slave SQL is Not Running
```

I think the hostname should be in the summary, and I don't really have any feedback around what should or shouldn't go into a description, because we don't use it.

templates/profile/prometheus/rules.yml.erb
umjnega: fixed.

templates/profile/prometheus/rules.yml.erb
umjnega: Brain and fingers not in agreement when I wrote this one.  fixed.

templates/profile/prometheus/rules.yml.erb
umjnega: Fixed and learned that summary is used, description is not.

templates/profile/prometheus/rules.yml.erb
umjnega: wow. fixed.

templates/profile/prometheus/rules.yml.erb
umjnega: wow wow. fixed.

PR #403, opened 2020-11-16
Dynamic weighting for haproxy back ends

aelkiss:
Uses "set-weights.rb" from ae-utils to set weights dynamically based on
server load

Tested on ictc-haproxy-001 with ictc-ht-web-*


daaang (approved):
I can't say I'm thrilled about adding more secrets as files on the puppet server, but I can't think of a better way to do this with what we have. Otherwise, this looks sensible enough.

aelkiss:
Likewise -- if we had a better place to manage secrets I'd be happy to use that instead.

PR #407, opened 2020-12-11
[WIP] AEIM-3241 - Upgrade to puppet 6 and update dependencies

daaang:


Dockerfile
```diff
@@ -1,12 +1,23 @@
...
+WORKDIR ${APP_HOME}
 
-ENV BUNDLE_PATH /gems
 RUN bundle install
 
-COPY . /usr/src/nebula/
-CMD ["/bin/bash"]
+COPY --chown=$UID:$GID . ${APP_HOME}
+
+CMD ["bundle exec rake spec"]
```
botimer, line 23: I think these need to be split out in the array.

botimer (commented):
This looks good to me, though I am unclear of the best approach to rolling it out and checking the net effect on the catalogues. Nice work!

PR #408, opened 2020-12-15
HT-2523, HT-2749, HT-2433: Add dependencies for additional backup locations

aelkiss:


aelkiss:
Just adds one package for HT machines.

FWIW, I think we are probably close to being able to migrate ingest to k8s, but I don't want to take that on quite yet or for that to block getting up and running with this.

aelkiss:
@daaang  Any concern with merging this? Just adds one package.

aelkiss:
We never merged the HT-2433 branch; I cherry-picked that commit here and hopefully we can just get that merged soon?

rrotter (approved):
I don't have any concerns about this.

PR #409, opened 2020-12-22
Add a cronjob to ensure press FCGI is running

daaang:


skorner (approved):
Looks good! It's not clear if a restart of the systemd unit works as expected in all failure modes, but this is progress that can be built on.

PR #412, opened 2021-01-08
AE-9092 Fix ignored status; Update to use pushgateway

sulemandiwan:


daaang (approved):
Pending a merge of #413, this looks exactly right. Whether this'll catch lockfiles is yet to be seen, but that's for later.

daaang (changes requested):
I merged the change to the pushgateway helper, so this is ready, but I just realized it won't do much good if you don't also add the alert rule e.g. https://github.com/mlibrary/nebula/blob/master/templates/profile/prometheus/rules.yml.erb#L135-L141

daaang (approved):
Yeah, I think a 30m window is perfect, and `severity: ticket` is also exactly right, given how long this seems to go unnoticed.

PR #422, opened 2021-01-27
HT-2749: install clamav-daemon on ingest machines

aelkiss:
There are some cases where we want HT ingest to virus scan material prior to ingest. This adds clamav and enables the clamav-daemon for use by https://github.com/hathitrust/feed/blob/master/lib/HTFeed/PackageType/EMMA/VirusScan.pm


manifests/profile/hathitrust/clamav.pp
```diff
@@ -0,0 +1,30 @@
+# Copyright (c) 2018 The Regents of the University of Michigan.
```
daaang, line 1: Almost hit you with a LGTM, but 2018

aelkiss: Fixed

aelkiss:
@daaang Just pinging you if we can merge this now.

PR #427, opened 2021-02-19
set upload_max_size and post_max_size to 128M

umjnega:


manifests/profile/www_lib/php.pp
```diff
@@ -101,7 +101,8 @@
       'PHP/max_input_vars'          => '2000',
       'PHP/memory_limit'            => '256M',
       'PHP/error_reporting'         => 'E_ALL & ~E_DEPRECATED',
-      'PHP/upload_max_filesize'     => '64M',
+      'PHP/upload_max_filesize'     => '128M',
+      'PHP/post_max_size'	    => '128M',
```
botimer, line 105: The linter is going to bark about alignment here.

botimer (changes requested):
Once the linter is happy, this looks good to me.

manifests/profile/www_lib/php.pp
umjnega: ah, yeah tab instead of spaces.  Thanks.  Fixed.

umjnega:
Fixed linter, it was a tab vs spaces.  My bad.

PR #429, opened 2021-03-02
Stop using the tiger auditing package

skorner:
Tiger is not providing actionable audit and monitoring
output for A&E, and systems now have the ITS mandated
CrowdStrike Falcon performing a similar role.


aelkiss (approved):
🪓🐅

PR #431, opened 2021-03-09
Shutdown VMs in parallel and with a 60s timeout

daaang:
Previously, they were sending suspend instead of shutdown, in serial,
with a 300s timeout. Also previously, we were not managing the
/etc/default/libvirt-guests file, and we are starting now.


botimer:
This looks great -- exactly as discussed. I especially like that you included only the values in the template, and not all the commented, boilerplate examples we can get anywhere else.

PR #436, opened 2021-03-24
Support for private DNS zone in dnsmasq for k8s cluster

aelkiss:
This is to support Isilon SmartConnect, which relies on delegating
queries for a particular zone to its private DNS server.

### DIFF
--- a/manifests/profile/kubernetes/dns_server.pp
+++ b/manifests/profile/kubernetes/dns_server.pp
@@ -9,6 +9,7 @@ class nebula::profile::kubernetes::dns_server {
   $kube_api_address = $cluster['kube_api_address']
   $node_cidr = $cluster['node_cidr']
   $private_domain = $cluster['private_domain']
+  $private_zones = $cluster['private_zones']
 
   package { 'dnsmasq': }
 
@@ -51,6 +52,15 @@ class nebula::profile::kubernetes::dns_server {
       content => template('nebula/profile/kubernetes/dns/hosts_06_ipv6_debian.erb'),
       order   => '06',
     ;
+  }
+
+  if $private_zones {
+    $private_zones.each |Hash $zone| {
+      file { "/etc/dnsmasq.d/${$zone[name]}":
+        content => "server=/${$zone[domain]}/${$zone[resolver]}\n",
+        notify  => Service['dnsmasq']
+      }
+    }
   }
 
   firewall {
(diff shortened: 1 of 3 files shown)


spec/classes/profile/kubernetes/dns_server_spec.rb
```diff
@@ -69,6 +69,11 @@
           .with_content("::1 localhost ip6-localhost ip6-loopback\n")
       end
 
+      it do
+        is_expected.to contain_file('/etc/dnsmasq.d/smartconnect')
+          .with_content("server=/sc.default.invalid/192.0.2.7\n")
```
daaang, line 74: I'd like to see another test for when the private zones aren't set, so that that behavior is mapped out. No need to like, expect it not to contain a file, but I'd at least like to be sure it compiles. An example of testing multiple cluster configs: https://github.com/mlibrary/nebula/blob/master/spec/classes/profile/kubernetes/kubeadm_spec.rb

manifests/profile/kubernetes/dns_server.pp
```diff
@@ -53,6 +54,12 @@
     ;
   }
 
+  $private_zones.each |Hash $zone| {
+    file { "/etc/dnsmasq.d/${$zone[name]}":
+      content => "server=/${$zone[domain]}/${$zone[resolver]}\n"
```
daaang, line 59: Do you want this to notify the dnsmasq service? Or does it not need a restart when the configuration changes?

daaang (changes requested):
Looks pretty straightforward. My only concern is about what happens in the default case where no private zones are defined, since it'll affect the other cluster.

manifests/profile/kubernetes/dns_server.pp
aelkiss: Yes, it should notify the dnsmasq service -- good catch.

spec/classes/profile/kubernetes/dns_server_spec.rb
aelkiss: Thanks -- wasn't sure of the best place to test it. FWIW `second_cluster` does not have it configured in `spec/fixtures/hiera/kubernetes/default.yaml` - are you suggesting another test here in `dns_server_spec.rb`?

aelkiss:
Feedback addressed, squashed & force pushed. Ready for re-review

daaang (approved):
Thanks for incorporating that, this looks great.

botimer:
Is this still in good shape?

aelkiss:
Is that a question for me or for @daaang  ?

botimer:
I guess it's a question for both -- is it still needed and acceptable for merge as-is?

aelkiss (commented):
Yes, this is still needed, and AFAIK it should be fine as-is.

aelkiss:
@botimer  IDK if you're still waiting on me on this, but it's fine from my perspective to merge this any time.

botimer:
Everything looks good here to me, and it sounds like it was just sitting with approval. Merging.

PR #437, opened 2021-03-30
Rename useradmin to otis

aelkiss:
Since otis admins more than just users


aelkiss:
I think this is ready to merge; there is a corresponding lensoftruth branch that needs to be merged concurrently. Once this is merged (and configuration is applied everywhere) we can purge useradmin from moku & the hatcher cluster.

daaang (approved):
Looks like a straightforward rename, cool.

PR #439, opened 2021-04-07
SYS-171 - Stop the pre-check for Test::NoWarnings from failing

botimer:
This was exiting 254, causing the exec to run every time, churning the
"changed" data. I found a way to get it to load the module and exit 0.


aelkiss:
:sob: 

I just wasted a bunch of time looking in the HT code base for things that required this, only to realize eventually it isn't even in that profile.

That said, you're already bringing in the Debian package `libtest-nowarnings-perl`, so why is this even being installed with CPAN?

aelkiss (changes requested):
This is unnecessary; in fact the perl modules pulled in at https://github.com/mlibrary/nebula/blob/79ae4439873c85ae32362acd51939054dada80d8/manifests/profile/www_lib/perl.pp#L138-L146 are all duplicated by Debian packages: https://github.com/mlibrary/nebula/blob/79ae4439873c85ae32362acd51939054dada80d8/manifests/profile/www_lib/perl.pp#L93-L101

botimer:
Okay, the extraneous CPAN modules are gone, relying on the debs. Should be 💯 .

aelkiss (approved):

daaang (approved):

PR #442, opened 2021-04-08
Sys 143 correct specific php versions

umjnega:


manifests/profile/php73.pp
```diff
@@ -19,16 +19,14 @@
...
-  'php-msgpack',
+  package { ['php7.3-geoip',
+  'php7.3-igbinary',
+  'php7.3-imagick',
+  'php7.3-memcached',
+  'php7.3-msgpack',
   #  'php-pear',
-  'php-redis',
-  'php-xdebug',
-  'php-xml',
```
botimer, line 31: Looks like maybe this one didn't make it over. What's the story for php-xml?

botimer (commented):
Looks pretty good, but I want to be sure that php-xml is handled. For the record, I did just double check that all of these versioned packages are in packages.sury.org repository, including php7.3-xml.

manifests/profile/php73.pp
umjnega: are 'php-redis' and 'php-xdebug' necessary if they are specificed with 'php7.3-redis' and 'php7.3-xdebug'?   The notes on SYS-108 led me to believe they might cause the conflict?

botimer (changes requested):
Cool, looks good. Let's squash this down to one commit with the issue key in the summary, and it'll be ready to go.

PR #444, opened 2021-04-20
SYS-183 Add Apt Mirror Profile/Role

sulemandiwan:


sulemandiwan:
This is fully done and tested. Seems to work. So just need a review before I can merge into production. If either of you want to bless it. @botimer @daaang

files/apt-mirror/apt-mirror-sync.sh
```diff
@@ -0,0 +1,54 @@
...
+# architecture
+ARCH=amd64
+
+# method
+METHOD=rsync
+
+# verbose
+#VERBOSE=-v
+
+# log timestamp
```
botimer, line 35: I think we should tighten this up and remove the one-line variables/comments because they don't really tell us anything more than the line itself. Specifically:

* We "import keys" but it's not clear why the debian keyring is to be imported, or where `trustedkeys.gpg` is.
* We mention "Fuse mountpoint", but there's nothing clearly about Fuse here, and the directory under `/mnt` is enough to see that we are making the apt-mirror mount point.

But really, I don't see any need for these variables because each argument is broken onto its own line and the variables used once. This script doesn't have defaults/overrides or take any arguments, so the variables don't buy us anything. If there is an essential reason for one of the _values_, that could be worth a comment before the invocation. In particular, I'm wondering whether the rsync has anything special to do with s3fs (I don't think so, but maybe).

The only value that is not obvious from usage here is the bucket name, which I would rename to `BUCKET_NAME`. The only value reused is `/mnt/apt-mirror`, which I would call `BUCKET_MOUNT` and use throughout. I would declare both as the first two lines, then inline everything else. This would make the only unlabeled argument to debmirror `"${BUCKET_MOUNT}/debian`, which puts the details in the "right" places and brings the file density closer to the conceptual density: two configured values (remote bucket and local path to mount it), and a few commands based on them.

manifests/profile/apt_mirror.pp
```diff
@@ -0,0 +1,54 @@
...
+# Profile for apt repo mirror
+#
+# @example
+#   include nebula::profile::apt_mirror
+class nebula::profile::apt_mirror (
+) {
+
+  package {
+    [ 'debmirror', 'debian-keyring', 's3fs' ]:
+  }
```
botimer, line 16: At least debian-keyring and s3fs are dependencies, but have very little to do with making an apt mirror, so we should use `ensure_packages` here to avoid incidental conflicts in other cases where we might use them.

manifests/profile/apt_mirror.pp
```diff
@@ -0,0 +1,54 @@
...
+
+  user { 'apt mirror user':
+    ensure     => 'present',
+    name       => 'apt-mirror',
+    comment    => 'apt mirror user',
+    gid        => '996',
+    groups     => ['apt-mirror'],
+    home       => '/home/apt-mirror',
+    password   => '!',
+    shell      => '/bin/bash',
```
botimer, line 32: Shouldn't this user's shell be `/bin/false`?

manifests/role/apt_mirror.pp
```diff
@@ -0,0 +1,16 @@
...
+# nebula::role::apt_mirror
+#
+# Debian apt mirror
+#
+# @example
+#   include nebula::role::apt_mirror
+class nebula::role::apt_mirror {
+  include nebula::role::aws
+  include nebula::profile::krb5
+  include nebula::profile::duo
```
botimer, line 14: Is there any need for regular user logins for a one-off machine like this? I think these can go away.

botimer (changes requested):

manifests/profile/apt_mirror.pp
sulemandiwan: Let me check that. It may be a leftover bit from when I was testing and isn't actually needed anymore.

files/apt-mirror/apt-mirror-sync.sh
sulemandiwan: I will redo the comments and the variable names which should tighten it up some. I have found having variables defined even if used once makes scripts easier to read, maintain, and edit for other people and for future self, but I will ponder about what you said.

manifests/role/apt_mirror.pp
sulemandiwan: I guess I need clarification on this. Will remove those two still allow us to log in as us? Our accounts are created separate right?

manifests/profile/apt_mirror.pp
sulemandiwan: Changed this.

manifests/profile/apt_mirror.pp
sulemandiwan: Updated this.

files/apt-mirror/apt-mirror-sync.sh
sulemandiwan: Adjusted the script. Comments make more sense now rather than repeating what the variable is. Condensed and cleaned it up a bit.

manifests/role/apt_mirror.pp
sulemandiwan: Also done.

botimer (approved):

files/apt-mirror/apt-mirror-sync.sh
```diff
@@ -0,0 +1,56 @@
...
+# remove them from the mirror because of how debmirror works.
+DIST=stretch,buster,stretch-updates,buster-updates,stretch-backports,buster-backports
+ARCH=amd64
+
+# rsync is the recommended standard for synchronizing repos.
+METHOD=rsync
+
+# Troubleshooting flags
+#DRYRUN=--dry-run
+#VERBOSE=-v
```
botimer, line 35: This won't be run by hand, please remove.

files/apt-mirror/apt-mirror-sync.sh
```diff
@@ -0,0 +1,56 @@
...
+ --host=${HOST} \
+ --root=/debian \
+ --dist=${DIST} \
+ --section=main,contrib,non-free,main/debian-installer \
+ --i18n \
+ --arch=${ARCH} \
+ --method=${METHOD} \
+ --passive --cleanup \
+ --getcontents \
+ $VERBOSE $DRYRUN
```
botimer, line 51: Won't be run by hand, please remove.

botimer (changes requested):
We could set the user's home dir to /var/local/apt-mirror, but that doesn't matter much. If we want to be able to run this with additional flags by hand, please use a trailing `"$@"` on the command or similar. Modifying the file to test something is bad news.

files/apt-mirror/apt-mirror-sync.sh
sulemandiwan: Removed.

files/apt-mirror/apt-mirror-sync.sh
sulemandiwan: Removed.

sulemandiwan:
> We could set the user's home dir to /var/local/apt-mirror, but that doesn't matter much. If we want to be able to run this with additional flags by hand, please use a trailing `"$@"` on the command or similar. Modifying the file to test something is bad news.

Sure. I changed the home directory and added the trailing $@. Intent wasn't to modify the file directly. Just to copy and modify. But this is cleaner though not as much at finger-tips.

botimer (approved):
:shipit:

PR #451, opened 2021-05-17
SYS-177 ruby health check

umjnega:
Added template files and node exporter code


templates/profile/prometheus/exporter/node/rubyHealthCheck-testall.sh.erb
```diff
@@ -0,0 +1,24 @@
...
+	./testruby.sh $VER > /dev/null 2>&1
+	if [ $? -ne 0 ]; then
+		echo "ruby_health_check{version=\"${VER}\"} 0"
+	else
+		echo "ruby_health_check{version=\"${VER}\"} 1"
+	fi
+
+	#Clean up ruby version test directories
+	rm -r $VER
+done
```
botimer, line 24: Tabs vs spaces

templates/profile/prometheus/exporter/node/rubyHealthCheck-cronjob.sh.erb
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
```
botimer, line 1: We usually put the small "managed by puppet" banner on files we drop in.

templates/profile/prometheus/exporter/node/rubyHealthCheck-testall.sh.erb
```diff
@@ -0,0 +1,24 @@
...
+if [ $? -ne 0 ]; then
+  echo "rbenv version can't run... source the profile setup before trying to test"
+  exit 1
+fi
+#./versions.sh | xargs -I{} ./testruby.sh {} | tee -a report.txt
+
+echo '# HELP ruby_heath_check Check Health of Ruby Version'
+echo '# TYPE ruby_health_check gauge'
+
+for VER in $(./versions.sh)
```
botimer, line 13: We can inline the versions.sh here and save some puppet code. It was standalone for convenience while putting the pieces together, but it's not really useful as such for the long term.

templates/profile/prometheus/exporter/node/rubyHealthCheck-testruby.sh.erb
```diff
@@ -0,0 +1,15 @@
...
+VER="$1"
+if [ -z "$VER" ]; then
+  echo 'Usage ./testruby.sh <version>'
+  exit 1
+fi
+mkdir "$VER"
+pushd "$VER"
+echo "$VER" > .ruby-version
+curl ae-files/ruby-check.tar | tar -x --strip-components=1
+./run.sh
```
botimer, line 11: These relative paths were convenient in piecing things together, but they could be brittle. I guess they already assume that they are being run in the script dir, so it's fine -- just noting that this is less robust than using dirname for script-relative or absolute.

I think the most robust answer here would be to have the cron job run in a clean temp directory and call the script from there. In that case, resolving the path to the running script is probably necessary. As always, it's more complicated than you would hope, but there is a copypasta recipe: https://stackoverflow.com/a/246128

botimer (changes requested):

templates/profile/prometheus/exporter/node/rubyHealthCheck-testall.sh.erb
umjnega: Cleaned up.

templates/profile/prometheus/exporter/node/rubyHealthCheck-cronjob.sh.erb
umjnega: added a comment.

templates/profile/prometheus/exporter/node/rubyHealthCheck-testall.sh.erb
umjnega: I think I corrected this.  In testing its working.

templates/profile/prometheus/exporter/node/rubyHealthCheck-testruby.sh.erb
umjnega: I may not be understanding this one completely.  I get that absolute paths are better.  However, when adding that copypasta (assuming i have it in the right place), it tries to run ./run.sh from outside the version directory.  I'm not entirely sure where to put it.    

Perhaps just doing "/usr/loca/rubytests/$VER/run.sh" instead of "./run.sh" in this case is acceptable?

botimer: Well, the cron job will invoke the testall script, but should probably cd to a temp dir. From there, we know where testruby will be, so we should invoke that absolutely. It naturally cd's into the specific version, so the relative ./run.sh there is no problem.

templates/profile/prometheus/exporter/node/rubyHealthCheck-testall.sh.erb
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+#Managed By Puppet
+#
+
+cd /usr/local/rubytests/
+export PATH=/l/local/rbenv/shims:/l/local/rbenv/bin:/usr/local/bin:/usr/local/sbin:$PATH
+export RBENV_ROOT=/l/local/rbenv
```
botimer, line 7: We should pull these out to the cronjob itself so they apply for anything we run. We're mostly interested in `rbenv` and `bundle`, which are placed on the path by bin and shims, respectively. Then we can invoke them directly.

umjnega: added to cronjob, rather than testall script.

botimer (approved):
After pairing and running linter/tests, I think this is ready to go.

PR #452, opened 2021-05-19
create tsm profile specific for sidecar

umjnega:
Copied prep_app_host to prep_app_host_tsm.  This new role will be applied to sidecar in order for backups to happen.

### DIFF
--- a/manifests/role/hathitrust/prep_app_host.pp
+++ b/manifests/role/hathitrust/prep_app_host.pp
@@ -14,4 +14,5 @@ class nebula::role::hathitrust::prep_app_host {
   include nebula::profile::python
   include nebula::profile::ruby
   include nebula::profile::docker
+  include nebula::profile::tsm
 }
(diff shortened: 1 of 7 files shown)


aelkiss:
I think it is more appropriate to add tsm to prep_app_host rather than copying. sidecar is the only prep_app_host currently, and I would expect that anything in the future with a similar role to sidecar (although I find it unlikely we would do anything quite like that rather than using k8s) would require TSM.

umjnega:
> I think it is more appropriate to add tsm to prep_app_host rather than copying. sidecar is the only prep_app_host currently, and I would expect that anything in the future with a similar role to sidecar (although I find it unlikely we would do anything quite like that rather than using k8s) would require TSM.

Yeah, i went back on my original plan.  Just added TSM to the prep_app_host role.  Had to do some clean up of the heira default files to make it all happy.  Thanks to @daaang for the help.

daaang (approved):
Long as the appropriate hiera is set in lensoftruth, this looks good to me. Glad to see quod.yaml and archivematica.yaml disappear.

PR #453, opened 2021-05-21
HT-2959: Shibboleth + Dex OIDC <-> SAML proxy

aelkiss:
This adds the Apache configuration to:

- reverse proxy to Dex from Apache, providing relevant Shibboleth headers
- automatically force authentication with Shibboleth at the correct time when authenticating through Dex
- Consume an entityID IdP hint parameter and redirect the user to log in through Shibboleth

There are certainly some duplications here with otis; if we were committed to puppet-apache + proxied access to apps in the long term I would want to consider refactoring this à la what we did for moku apps, but my sense is that we aren't committed to puppet-apache in the long term, that we will hopefully not need to use the apache basic auth for proxied apps all that much longer, and that eventually (at least for HathiTrust) I'll likely eventually want to containerize Apache itself, I don't think that refactoring would be worthwhile.

### DIFF
--- a/manifests/profile/hathitrust/apache/babel.pp
+++ b/manifests/profile/hathitrust/apache/babel.pp
@@ -19,6 +19,8 @@ class nebula::profile::hathitrust::apache::babel (
   String $gwt_code,
   String $otis_endpoint,
   String $otis_basic_auth,
+  String $dex_endpoint,
+  String $dex_basic_auth,
   Array[String] $cache_paths = [ ],
 ) {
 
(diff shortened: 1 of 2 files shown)


aelkiss:
I have tested this via `preview.babel.hathitrust.org` and verified that I can go through a complete authentication workflow (using `example-app` from Dex), so I think this should be good to go whenever we want to merge it.

daaang (approved):
Yeah, as you say, basic http auth isn't ideal, but it works, and if this is adding complexity to our IaC, the wolf's been in this particular barn for a long while now. Anything that encourages us to move away from every application needing to solve authentication on its own is a win in my book.

PR #459, opened 2021-07-14
AE-10100 - Increase jira ProxyTimeout from 60s to 300s

daaang:
On some of the bigger projects, it's taking longer than 60 seconds for
jira to load pages, so mod_proxy is timing out, since it's currently
defaulting to the value of Timeout, which the puppet apache module sets
to 60 seconds by default.

Left to its own, apache has a default timeout of 300s, so we're not
really doing anything all that weird by having a 5 minute proxy timeout.


skorner:
The rationale is sound and shouldn't lead to resource contention from too many open connections in this non-public use case.

PR #460, opened 2021-07-15
Redirect my-account to account.lib

daaang:


skorner:
This looks like it will accomplish what is needed (not redirecting favorites and redirecting all other paths as requested. With Apache 2.4 you could in theory put all the rewrites in an "If" expression block rather then have two no-op rewriterules, but that experimentation is not worth the expediency need right now. A rewrite condition pair won't gracefully work since you need it to apply to all the subsequent rewrites .

PR #463, opened 2021-07-26
HT-3073: Run catalog cron job getall.sh at both sites

aelkiss:
This stems from an error made when originally migrating these jobs, and perhaps before that; the getall.sh cron job should in fact run at both sites.

The catalog cron job profile is only included in the HT global primary
role; the mdp_misc one is included in the site primary.

I wonder if perhaps we should rename mdp_misc to "once per site" and
catalog to "once anywhere", or if that isn't really worth it.


daaang (approved):
I didn't read your comment, and what I was going to say was "looks like you're moving this cronjob from the `global_primary` role to the `site_primary` role, so `catalog` and `mdp_misc` might not be the right names for the profiles."

Sounds like we're on the same page. I don't think a rename needs to happen now, but probably next time any of this changes.

aelkiss:
With any luck we can just put these all in k8s rather than change more of it in puppet.

PR #465, opened 2021-07-29
Configure script to merge application logs in mdp_misc crontab

respinos:


manifests/profile/hathitrust/cron/mdp_misc.pp
```diff
@@ -57,7 +57,11 @@
     'translation maps':
       minute  => '58',
       hour    => '15',
-      command => "${catalog_home}/derived_data/getall.sh ${catalog_home}/derived_data"
+      command => "${catalog_home}/derived_data/getall.sh ${catalog_home}/derived_data";
+
+    'merge application logs':
+      minute  => 05,
+      command => "${home}/mdp-misc/scripts/merge_application_logs.pl";
```
daaang, line 64: Given that `${home}` is `/htapps/babel/mdp-misc`, I'm betting this should be `"${home}/scripts/merge_application_logs.pl"`. Other than that, this looks fine, long as the intended behavior is for this to run every hour. I'll admit that I'm curious about what this cronjob does—I'll admit that I'm not used to reading perl, so it's hard for me to understand what this is doing just by looking at the script itself.

respinos: Oops, yes. Sorry. It should be `/scripts/` in that path. (Bad edit when editing the longer path to use `$home`.)
The TL,DR: the applications are writing to logfiles that are specific to the PID to avoid using semphores/locking to improve performance. With CGI scripts that's a metric ton of files being generated, so they're being "staged" in hour-based directories. This script appends the previous hours to maintain one logfile per day.

PR #467, opened 2021-08-10
HT-3042: add PTSEARCH_SOLR env var

aelkiss:
This seems to be the easiest way to inject a different host for ptsearch
solr at different data centers. Right now deployment assumes
configuration will be the same at both data centers and differences will
be handled by `/etc/hosts`. However, deploying ptsearch solr in Kubernetes at
multiple data centers with proxied ingress via HTTPS means we need
separate hostnames so that letsencrypt can successfully generate a
cert. We considered several options including disabling cert
verification, alternative ingress setups, etc, and this seems this least
disruptive; it also gives us an easy path to vary the solr for
test/preview in addition to production.

### DIFF
--- a/manifests/profile/hathitrust/apache/babel.pp
+++ b/manifests/profile/hathitrust/apache/babel.pp
@@ -21,6 +21,7 @@ class nebula::profile::hathitrust::apache::babel (
   String $otis_basic_auth,
   String $dex_endpoint,
   String $dex_basic_auth,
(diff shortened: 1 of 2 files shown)


daaang (approved):
I don't speak apache well enough to understand what all these SetEnv directives are applying to, but I don't really need to understand that to see that this change can't possibly hurt anything.

PR #468, opened 2021-08-17
SYS-239 - Alert when Fedora is down for Deep Blue Data

botimer:
We check Fedora every minute. If it does not respond for 15 minutes, we
send a page level alert to investigate promptly.

### DIFF
--- a/templates/profile/prometheus/rules.yml.erb
+++ b/templates/profile/prometheus/rules.yml.erb
@@ -218,6 +218,14 @@ groups:
     labels:
       severity: ticket
 
+  - alert: DeepBlueDataProdFedoraDown
+    annotations:
+      summary: 'Fedora is unresponsive on node {{$labels.hostname}}.'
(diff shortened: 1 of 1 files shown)


daaang (changes requested):
I think it's weird for this to be 15m and a page, while the press check is 30m and a ticket. They're both "production," so if they're different, I'd like us to understand why they're different. So the changes I'm requesting are:

1. Either this alert waits 30m or the press one waits 15m.
2. Either this alert is `severity: ticket` or the press one is `severity: page`.

If either of the above is wrong, then a clear comment explaining the difference is fine too.

botimer:
We have redundancy for press, and the restart there is failsafe. Fedora being single instance and occasionally having lock file problems requiring manual intervention were my reasons for this being different. I'm not set on these differences, though.

botimer:
Per discussion, we have a few other places that might not be following the best patterns now, but the details and expectations for this one are documented.

PR #471, opened 2021-09-08
Reduce cron email notifications

billdueber:
First time dealing with puppet, so I really need someone to check my work.

All in mdp_misc.pp

* Set an $sdr_environemnt variable without MAILTO, and
  use it to define a default env ($sdr_environment +
  MAILTO=${mail_recipient}`) that can be overridden for items that
  take care of their own mail
* Change "harvest proxy downloads", "data api log monitor", and
  "manage mbook sessions" to both:
  * use the new, mailto-free environment
  * only send email on error (e.g., use || instead of |)

### DIFF
--- a/manifests/profile/hathitrust/cron/mdp_misc.pp
+++ b/manifests/profile/hathitrust/cron/mdp_misc.pp
@@ -16,16 +16,18 @@ class nebula::profile::hathitrust::cron::mdp_misc (
   String $sdr_data_root = '/sdr1',
   String $home = '/htapps/babel/mdp-misc',
   String $catalog_home = '/htapps/catalog/web',
-  Integer $mdp_sessions_minute = 5
+  Integer $mdp_sessions_minute = 5,
+  $sdr_environment = [
+    "SDRROOT=${sdr_root}",
+    "SDRDATAROOT=${sdr_data_root}",
+    "HOME=${home}"
+  ]
 ) {
 
   cron {
     default:
       user        => $user,
-      environment => ["MAILTO=${mail_recipient}",
-      "SDRROOT=${sdr_root}",
-      "SDRDATAROOT=${sdr_data_root}",
-      "HOME=${home}"];
+      environment => $sdr_environment + ["MAILTO=${mail_recipient}"];
 
     # Mail /htapps/babel/logs/assert/hathitrust=email-digest-current at 15 minute intervals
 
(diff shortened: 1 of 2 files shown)


billdueber:
OK, well the failure is (I think?) due to formatting issues. But `bundle install` fails when I try to do so to run the tests locally.

```
Bundler could not find compatible versions for gem "puppet-syntax":
  In snapshot (Gemfile.lock):
    puppet-syntax (= 2.6.1)

  In Gemfile:
    puppet-module-posix-dev-r2.6 was resolved to 1.1.0, which depends on
      puppet-syntax (~> 3.0)

    rspec-puppet-utils was resolved to 3.4.0, which depends on
      puppetlabs_spec_helper was resolved to 2.15.0, which depends on
        puppet-syntax (>= 2.0, < 4)

Running `bundle update` will rebuild your snapshot from scratch, using only
the gems in your Gemfile, which may resolve the conflict.

Bundler could not find compatible versions for gem "rake":
  In snapshot (Gemfile.lock):
    rake (= 13.0.1)

  In Gemfile:
    rake (>= 12.3.3)

    puppet-module-posix-dev-r2.6 was resolved to 0.5.3, which depends on
      dependency_checker (~> 0.2) was resolved to 0.2.0, which depends on
        rake (~> 12.3)

    rspec-puppet-utils was resolved to 3.4.0, which depends on
      puppetlabs_spec_helper was resolved to 2.15.0, which depends on
        puppet-syntax (>= 2.0, < 4) was resolved to 2.6.1, which depends on
          rake

Running `bundle update` will rebuild your snapshot from scratch, using only
the gems in your Gemfile, which may resolve the conflict.

```

manifests/profile/hathitrust/cron/mdp_misc.pp
```diff
@@ -16,16 +16,18 @@
   String $sdr_data_root = '/sdr1',
   String $home = '/htapps/babel/mdp-misc',
   String $catalog_home = '/htapps/catalog/web',
-  Integer $mdp_sessions_minute = 5
+  Integer $mdp_sessions_minute = 5,
+  $sdr_environment = [
```
aelkiss, line 20: Do we want this in the parameters section? This would allow us to override the defaults; not sure if/when we would want to do that.

billdueber: Let's leave it alone for the moment. I need to re-jigger those scripts anyway so they're less...bad.

aelkiss: :heavy_check_mark:

aelkiss:
Tests should be fixed now.

PR #473, opened 2021-10-05
HT-3127: remove kubernetized & superceded cron jobs

aelkiss:


aelkiss:
Cron jobs are migrated here:

https://github.com/hathitrust/ht_tanka/blob/main/environments/ingest/cron_jobs_macc/main.jsonnet

PR #481, opened 2022-01-06
Add log rotation to fulcrum.

daaang:
These settings are copied right from bulleit-1 for heliotrope-production
logging.


daaang:
I looked at the error in the tests, and it's from spec/classes/role/fulcrum_spec.rb which was already failing before adding logrotate.

botimer (approved):
Nice! Thanks!

PR #489, opened 2022-02-02
Unset Shib headers before setting to prevent spoofing

aelkiss:
When we always set the X-Shib header (regardless of whether the env var
from mod_shib was present), any client-supplied header would get wiped
out. When conditionally setting the header, a client-supplied header
could be injected if mod_shib does not supply a value. This is not the
behavior we want.

I've tested this on beta-1.babel.hathitrust.org and verified that 1) this pattern (`RequestHeader unset` followed by a conditional `RequestHeader set`) still passes through the Shib-supplied env vars and that 2) with this pattern, client-supplied headers don't make it through.

### DIFF
--- a/manifests/profile/hathitrust/apache/babel.pp
+++ b/manifests/profile/hathitrust/apache/babel.pp
@@ -346,6 +346,15 @@ class nebula::profile::hathitrust::apache::babel (
     ",
 
     request_headers             => [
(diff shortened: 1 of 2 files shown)


botimer (approved):
Nice. A little clumsy that we have to double touch everything, but there's nothing complex here.

PR #490, opened 2022-02-02
Add validate_cmd for shibboleth2.xml

aelkiss:
This should, in theory, validate proposed changes to shibboleth2.xml before putting them in place. I thought that shibd would refuse to reload an invalid shibboleth2.xml (and that may be true), but it also (surprisingly to me) causes apache / mod_shib to fail. 

This does NOT validate changes to other Shibboleth configuration files (e.g. attribute-map.xml or attribute-policy.xml); I'm not sure how we would do that with either `shibd -t` or with the puppet configuration that just puts the entire directory in place, but I'm also reasonably certain that I've pushed out changes to those files with errors and that it does not bring down the web server, so... :shrug: 

The upcoming change to shibboleth SP version 3 in Debian 11 might give us a good chance to rethink workflow here. If nothing else, it would be ideal to 1) have the configuration files for shibboleth managed somewhere version-controlled and 2) to have a way to easily try out a change on e.g. `test.babel.hathitrust.org` without applying it everywhere. That concern is somewhat orthogonal to actually pushing out a broken config, though.

### DIFF
--- a/manifests/profile/shibboleth.pp
+++ b/manifests/profile/shibboleth.pp
@@ -79,10 +79,11 @@ class nebula::profile::shibboleth (
   }
 
   file { '/etc/shibboleth/shibboleth2.xml':
-    mode   => '0440',
-    owner  => '_shibd',
-    group  => 'nogroup',
-    source => "${config_source}/shibboleth2.xml"
+    mode         => '0440',
+    owner        => '_shibd',
+    group        => 'nogroup',
+    source       => "${config_source}/shibboleth2.xml",
+    validate_cmd => '/usr/sbin/shibd -t -c %'
   }
 
   file { '/etc/systemd/system/shibd.service.d':
--- a/manifests/role/webhost/htvm.pp
+++ b/manifests/role/webhost/htvm.pp
@@ -8,6 +8,7 @@
 #   include nebula::role::webhost::htvm
 class nebula::role::webhost::htvm (
   String $private_address_template = '192.168.0.%s',
+  String $shibboleth_config_source = 'puppet:///shibboleth'
 ) {
   include nebula::role::hathitrust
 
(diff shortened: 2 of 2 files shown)


aelkiss:
We should also verify this works on e.g. test.babel.hathitrust.org -- perhaps by making a trivial change like adding whitespace to `shibboleth2.xml` and verifying that validation succeeds and the file gets put into place.

daaang:
Yeah. I guess my other thought would be to verify that it does indeed reject the file from earlier today.

daaang (approved):

aelkiss:
I verified on ht-web-testing that running `shibd -t -c /tmp/shibboleth2.xml` with a broken file returns nonzero status. I'm not sure how to test that with the puppet validate_cmd without accidentally breaking production.

aelkiss:
Next step is running with this environment on ht-web-testing with a trivial change to shibboleth2.xml and verifying that we see the validate_cmd being run and (ultimately) the changed file being put in place.

aelkiss:
Well, puppet doesn't seem to show output from running the validate command... Would be nice if we had a way to actually verify in testing that puppet won't put the broken config in place. Maybe it's worth being able to specify the location of config per-server..

aelkiss:
Actually, it looks like we already support that... `nebula::profile::shibboleth::config_source`... I'll give it a try.

aelkiss:
OK, I was able to test this by overriding the shibboleth config source for ht-web-testing and verify that it runs the test and does not put an invalid file in place.

botimer (approved):
This looks ready to go. Any last checks before we merge?

aelkiss:
I don't think so. I'll go ahead and merge the corresponding lensoftruth branch.

aelkiss:
Since this is approved, I'm going to go ahead and hit merge.

PR #491, opened 2022-02-10
always use preseed for installing debian vms

rrotter:
This appears to have been a special case we were using to support jessie
vms, which are no longer supported. It was breaking bullseye support, and would
also break any future releases.


rrotter:
Going to rewrite this based off of #492.

PR #499, opened 2022-03-11
Debian 11 HT web VM

aelkiss:
Per discussion in Slack, and not wanting to make a PR that (if merged) will break a bunch of stuff, I've made this against a new bullseye branch.

### DIFF
--- a/spec/classes/role/htvm_test_webhost_spec.rb
+++ b/spec/classes/role/htvm_test_webhost_spec.rb
@@ -12,13 +12,6 @@ describe 'nebula::role::webhost::htvm::test' do
       require 'pry'
       include_context 'with setup for htvm node', os_facts
 
-      it { is_expected.to compile }
-      it { is_expected.not_to contain_package('php5-common') }
-      it { is_expected.not_to contain_package('php5-dev') }
-
-      if(os == 'debian-11-x86_64')
-        it { is_expected.not_to contain_package('libapache2-mod-shib2') }
-      end
     end
   end
 end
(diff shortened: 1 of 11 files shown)


aelkiss:
Tests are currently failing because of a conflict between requiring `mariadb-client` in `manifests/profile/hathitrust/dependencies.pp` and something the chipmunk role is bringing in. I think using `ensure_packages` should fix it.

aelkiss:
Assuming we merge this to the `bullseye` branch, we should follow this up with a change to:

- add a puppet.conf on ae-files that defaults to the bullseye environment
- update https://github.com/mlibrary/nebula/blob/7166267a24b1c1acfbc5b8a88b4d41879968f79f/templates/virtual_machine/bullseye.cfg.erb#L486 to use that puppet.conf by default

aelkiss:
@skorner  says:

> have you noticed the logrotate failure emails the past week or so for your new VMs? The common issue is a duplicate config for `btmp` and `wtmp` (both seem to come from puppet based on the conf.d file header, but one prefixed with `debian_` ). The Fulcrum VMs have a separate issue wrt the perms on the app log path.

spec/classes/role/htvm_test_webhost_spec.rb
```diff
@@ -9,8 +9,9 @@
 describe 'nebula::role::webhost::htvm::test' do
   on_supported_os.each do |os, os_facts|
     context "on #{os}" do
+      require 'pry'
```
aelkiss, line 12: need to remove this

rrotter (commented):
All looks fine to me style-wise. Before we merge this I'd like to hear from @daaang on the dependency updates so we don't conflict with upstream.

aelkiss:
Next step is to get this to a point where as much as possible except for incompatible php module updates can be merged. Will make a fresh branch/PR for that work.

aelkiss:
Superceded by #512 , #513

PR #500, opened 2022-04-04
HathiTrust DEV-234 umich cosign factor

aelkiss:
We need this to tell if a user is a Michigan non-Friend user, which we
use in computing the actual user ID. This is a legacy of having used
uniqnames as the user identifier with cosign and never having fully
migrated away from that.

See https://github.com/hathitrust/mdp-lib/blob/master/Utils.pm#L92-L104


aelkiss:
FYI @respinos  @moseshll

We should perhaps consider if there is a way to move away from this practice...

botimer (approved):
This looks totally fine to me. Are you ready to merge?

aelkiss:
Yes, should be fine to merge any time.

PR #502, opened 2022-04-11
Move command argument to the right place; add alert

daaang:
I added an argument to `docker run` that should have been an argument to
the running container. I noticed the problem because I happened to
check; the pushgateways have been down since that change, and we haven't
heard anything because we didn't have an alert for any exporters being
down other than the node exporters.


botimer (approved):
:shipit:

PR #503, opened 2022-04-11
HT-3127: remove full zephir diff job

aelkiss:
- tested full zephir diff job in kubernetes
- prune removed cron jobs


aelkiss:
@botimer  This is ready to merge whenever.

PR #504, opened 2022-04-13
HELIO-4186 - Add default vhost for monitoring

botimer:
HAProxy checks /monitor/monitor.pl for backend liveness, and that is set
up in apache::base. However, we did not have a default vhost for
requests via IP/fqdn, so the checks were failing. This adds the default
vhost to the fulcrum+apache role. Ideally, it would be in the profile,
but that would collide with the standard inclusion via misc on general
servers.

### DIFF
--- a/manifests/role/webhost/fulcrum_www_and_app.pp
+++ b/manifests/role/webhost/fulcrum_www_and_app.pp
@@ -20,6 +20,15 @@ class nebula::role::webhost::fulcrum_www_and_app (
   include nebula::profile::www_lib::apache::base
   include nebula::profile::www_lib::apache::fulcrum
 
+  # Include a default vhost to catch monitoring requests by IP/fqdn.
+  # This is here rather than in the profile because it would be duplicate
+  # on www_lib_vm.
+  class { 'nebula::profile::www_lib::vhosts::default':
+    prefix => '',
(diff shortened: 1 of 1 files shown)


manifests/role/webhost/fulcrum_www_and_app.pp
```diff
@@ -20,6 +20,15 @@
   include nebula::profile::www_lib::apache::base
   include nebula::profile::www_lib::apache::fulcrum
 
+  # Include a default vhost to catch monitoring requests by IP/fqdn.
+  # This is here rather than in the profile because it would be duplicate
+  # on www_lib_vm.
+  class { 'nebula::profile::www_lib::vhosts::default':
```
daaang, line 26: It won't surprise you that I don't like parameters in roles, but the alternative would be creating new classes that map 1:1 to this role and to he www-lib role, and I think that path would be unreasonable and would gain us nothing of value. :shipit:

daaang (approved):

manifests/role/webhost/fulcrum_www_and_app.pp
botimer: Agreed, thanks. This can go away once fulcrum migrates out of www_lib_vm.

PR #510, opened 2022-04-26
Add badrobots support for Deep Blue Data and Documents

skorner:
Proxied requests to the apps bypass the general badrobots check for the vhost. Add a RewriteCond to apply badrobots to proxied app requests.

### DIFF
--- a/manifests/profile/www_lib/vhosts/deepblue.pp
+++ b/manifests/profile/www_lib/vhosts/deepblue.pp
@@ -72,11 +72,12 @@ class nebula::profile::www_lib::vhosts::deepblue (
       },
       {
         comment      => 'Deep Blue Data',
+        rewrite_cond => '%{ENV:badrobot} !(^true$)',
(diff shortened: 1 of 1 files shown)


botimer (commented):
Syntactically, this looks fine, but there is a scarce trail to the rationale. We apply badrobots.conf, which cannot be referenced directly. We can infer from this change that the `badrobot` env var is set to `true` if the requester is somehow offensive, but it's unclear why these are behaving differently from the usual `require not env badrobot` treatment.

Are the normal rules not applying because these rewrites fall outside of any `<Directory>` or `<Location>`? Could we cover with them with a location to get them on regular behavior rather than add specialized `RewriteCond`?

skorner:
I believe that is correct as to why the normal rules are not applying. In my quick research, it appears that using a RewriteCond is the standard approach when you need to key off of an env var value for a RewriteRule. It's possible we could achieve the same result with a suitable <Directory> or <Location>, but it was not obvious to me what to try.

botimer:
Okay, good enough for me. It's a small, strange thing, but no reason to hold this up.

PR #512, opened 2022-05-11
Debian 11 support

aelkiss:
This should add most of what's needed for debian 11 support for HT without breaking anything for debian 9. This is mostly ported from https://github.com/mlibrary/nebula/tree/ht-debian-11 but keeping conditionals to support both stretch and bullseye. ~Still to-do is to add back some of the tests from the ht-debian-11 branch that are not yet replicated here.~ The tests are present now.

### DIFF
--- a/templates/profile/dns/smartconnect/named.conf.options.erb
+++ b/templates/profile/dns/smartconnect/named.conf.options.erb
@@ -3,7 +3,8 @@
 options {
   directory "/var/cache/bind";
 
-  dnssec-validation auto;
+  dnssec-enable no;
+  dnssec-validation no;
 
   auth-nxdomain no;    # conform to RFC1035
 
--- a/.fixtures.yml
+++ b/.fixtures.yml
@@ -18,7 +18,7 @@ fixtures:
       ref: "6.5.0"
     apache:
       repo: "puppetlabs/apache"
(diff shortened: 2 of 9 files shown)


templates/profile/dns/smartconnect/named.conf.options.erb
```diff
@@ -3,7 +3,8 @@
 options {
   directory "/var/cache/bind";
 
-  dnssec-validation auto;
+  dnssec-enable no;
```
aelkiss, line 6: I can't imagine this would break anything on stretch since AFAIK we aren't using it anyway, but it would be worth verifying on e.g. test.babel.hathitrust.org before applying in production.

aelkiss:
Compared to the `ht-debian-11` branch, this adds some conditionals to try to ensure continued support for stretch (https://github.com/mlibrary/nebula/compare/ht-debian-11...debian_11_support)

templates/profile/dns/smartconnect/named.conf.options.erb
aelkiss: I did try this out on `test.babel.hathitrust.org` and verified that the Isilon mounts still work after the change to `named.conf`.

PR #515, opened 2022-05-25
added repo profile for tesseract

umjnega:
Copied from the yarn.pp repo set up.   I dont see a spec test anywhere, but I'm thinking i should have one? maybe for practice...


skorner:
We'll see what others think, but my instinct is to name the profile based on what it is adding support for (tesseract) rather than the official repo name in this case, as that is more meaningful. I'd also include adding this profile to the [quod_dev role](https://github.com/mlibrary/nebula/blob/master/manifests/role/app_host/quod_dev.pp) in the same PR.

umjnega:
Ah yeah, i had thought about naming it "tesseract" too. Wasn't sure about those fields and how they would apply to creating the repo file and what not.    I can update it and add to the quod_dev role.  Thanks for the feedback.

umjnega:
Probably failing because we dont have a test set up for Debian Buster (10)

nebula::role::app_host::quod_dev
  on ubuntu-20.04-x86_64
    is expected to compile using default hieradata (FAILED - 1)
  on debian-9-x86_64
    is expected to compile using default hieradata (FAILED - 2)
  on debian-11-x86_64
    is expected to compile using default hieradata (FAILED - 3)

PR #517, opened 2022-05-26
20220525 logrotate bugfix

skorner:
This PR includes a commit that fixes fulcrum application log rotation and a commit that fixes our override of btmp and wtmp log rotation.


botimer (approved):
This looks good and I believe it is safe based on outside conversation.

PR #519, opened 2022-06-02
Remove support for legacy geoip database

aelkiss:
- Remove geoip entirely from www_lib; according to this thread
https://mlit.slack.com/archives/C0PPPC8GK/p1648147000168719 it is no
longer being used

- Remove geoipupdate from HathiTrust servers; the databases are now
managed and updated centrally in a shared location (per maxmind best
practices) rather than pulled on each VM


skorner:
Looks good to me. I am on board with not forcing a purge on existing www-lib VMs, given they will be replaced soon anyway.

PR #520, opened 2022-06-02
Use puppet client cert for apache cert auth

aelkiss:
This is the first step (adding the client cert); then we'll need update the ingresses to accept the client cert; then we can remove the basic auth.


aelkiss:
Not sure what tests (if any) would be useful to add for this.

aelkiss:
One thing this won't cover is handling what to do when the puppet CA cert expires - I'm not sure what that will look like in terms of renewing node certificates or in terms of updating the allowed CA on the k8s side. I assume we'll be able to renew it and use that renewed cert without needing to reissue client certs, but we should make sure we understand that well before the current cert expires (Nov 25, 2024)

aelkiss:
Removed basic auth from this commit, and added new ingresses that support TLS auth for otis and dex.

ptsearch will continue to use basic auth; we aren't proxying to that from apache but rather it's from the application side; also if we were using the solr auth instead of k8s-provided auth via ingress it would still be basic auth.

The testing & rollout procedure will look like:

- use this on the `client-certs-bullseye` branch (which has this commit as well as the puppet php module version 8) in hathitrust development, with a corresponding lensoftruth branch that changes the endpoints to the new ones that accept tls auth instead of basic auth
- once that is working, merge this branch to master
- then, rebase the `debian_11_with_puppet_php_8` branch to include this new commit; update the corresponding lensoftruth branch with the new endpoints; push both

@daaang  Does that all make sense to you? If so, go ahead and approve the PR, and I'll get it all rolled out.

aelkiss:
The client_certs_bullseye branch is working fine with e.g. test.babel.hathitrust.org.

botimer (approved):
We've tested this approach and I can confirm it works as expected. The certificate and key are combined properly into a single PEM, and both curl and Apache can present it as a client certificate from the one file.

If the receiving server has, and trusts, the CA cert used by the Puppet server for validating client certificates, the proxy works beautifully.

PR #521, opened 2022-06-07
Remove ingest service

aelkiss:
There is a corresponding nebula change to remove the ingest indexing primary role as well.


aelkiss:
This branch has been running on macc-ht-ingest-* for several days with no known issues.

daaang (approved):
:party_parrot:

PR #522, opened 2022-06-07
Add developer conveniences for dev/testing

aelkiss:
Per request from @respinos


daaang (approved):
Makes a lot of sense to me. Also I'm figuring that role::webhost::hathitrust::test is no more because it's been replaced functionally by htvm::test.

aelkiss:
Correct.

PR #528, opened 2022-06-17
Revert to traditional hosting for the time being

botimer:
There are some oddities in sending email. Sort them out in staging
first.


umjnega:
Going back to the previous server.. look good to me.

PR #529, opened 2022-06-20
HathiTrust DEV-316: CRMS_INSTANCE only set for production

aelkiss:
This supports the need to vary `CRMS_INSTANCE` based on whether a vhost is development or production (-ish).

- `CRMS_INSTANCE=production` for actual production as well as for `preview.babel.hathitrust.org`
- Not set by default elsewhere

This isn't the prettiest fix. Currently, I'm expecting to override this via hiera where needed -- while it might be nice to set it via the test vs. prod role, that would (I think) necessitate changes to a long chain of included files:

- `nebula::role::webhost::htvm::test` and `::htvm::prod` both include
- `nebula::role::webhost::htvm` includes
- `nebula::profile::hathitrust::apache` includes
- `nebula::profile::hathitrust::apache::babel`

If anybody has a more elegant way to vary this than what's given here, I'm all for it, but I don't think it's worth spending a lot of time on at this point.

### DIFF
--- a/manifests/profile/hathitrust/apache/babel.pp
+++ b/manifests/profile/hathitrust/apache/babel.pp
@@ -21,6 +21,7 @@ class nebula::profile::hathitrust::apache::babel (
   String $dex_endpoint,
   String $ptsearch_solr,
   String $ptsearch_solr_basic_auth,
+  Boolean $prod_crms_instance = true,
   Array[String] $cache_paths = [ ],
 ) {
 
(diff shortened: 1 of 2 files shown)


botimer (approved):
Without pulling too hard on the thread, I take this to be _not_ setting the association for the "current" hostname to production unless so designated, which lets the app figure it out.

I think this is another case where the abstract and concrete collide in our mega-module pattern. A control repo with the actual infrastructure represented would be more direct. Here, we've kind of got concrete mixed into the abstract, with nowhere to put the actual concrete (.pp), and so we have to make an escape valve in hiera for when to apply it/override.

I'm sure this feels like an annoyance with puppet-apache, but I think it is probably more the case of doing in a module some of what should be in a control repo.

In any case, if I understand what is actually happening, 👍 :shipit: .

PR #531, opened 2022-06-22
Export haproxy backends with max instead of min

daaang:
When an haproxy server is down, it exports 0 backends, so alerts based
on these metrics were firing every time any single haproxy server went
down. Using max instead of min will mean that if any haproxy server is
detecting a backend as up, that'll be enough.


skorner:
This make sense to me, given the alert is about availability of the back ends and not haproxy.

PR #532, opened 2022-06-24
add new httpchk line

umjnega:
the test doesn't appear to have a check for the contents of the file that I edited... just that the file exists and has the right perms.. will find out.


daaang (approved):
I'm not a huge fan of inlined multi-line content, but 2 lines isn't enough to make me demand a separate file. If it were 10 lines, I'd be requesting changes, but it's 2 lines, and I think that's exactly right.

PR #535, opened 2022-06-29
Add xmllint, xsltproc for ht dev environmnet

aelkiss:
Per @respinos


aelkiss:
@umjnega  just need a quick sign-off on this -- would you or somebody else in A&E be able to review & approve soon?

umjnega:
Looks ok to me.. but I did notice that you didn't end the last item with a comma.....  looking at an example from another perl.pp that is currently in use.. we end the list of packages with a comma on the last package..  Is this maybe not necessary?

aelkiss:
I don't think it's necessary -- passed CI & puppet-lint

PR #537, opened 2022-07-07
Update quod-prod profile perl dependencies for bullseye

skorner:
- remove libcpan-meta-perl (not present in bullseye and no replacement needed)


daaang (approved):
**Matt** has automatically approved of this **+0 –1** PR.

PR #538, opened 2022-07-07
Set UseCanonicalName

aelkiss:
This is required for validating the "target" parameter in Shibboleth
sign on.

This also splits out crms-training to a separate vhost; this is required
so that Shibboleth knows the proper return target rather than having to
differentiate between crms-training.babel.hathitrust.org and
babel.hathitrust.org

### DIFF
--- a/manifests/profile/hathitrust/apache.pp
+++ b/manifests/profile/hathitrust/apache.pp
@@ -112,6 +112,7 @@ class nebula::profile::hathitrust::apache (
   }
   class { 'apache::mod::proxy_fcgi': }
   class { 'apache::mod::reqtimeout': }
+  class { 'apache::mod::setenvif': }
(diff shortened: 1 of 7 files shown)


aelkiss:
See also #529  which this supercedes in part for the `CRMS_INSTANCE` handling.

skorner (approved):
This looks good to me. As you noted, enabling UseCanonicalName is required to allow us to leverage desired Shib SP functionality during session creation. Bundling this change with the necessary creation of a dedicated crms-training vhost makes sense to me. The tests pass, but if you want someone to review the code changes more carefully than I can, please feel free to request it.

aelkiss:
I'm having issues with this disabling the `setenvif` module. I'll see if I can resolve that before merging.

PR #541, opened 2022-07-20
HathiTrust DEV-345: fix redirect to WAYF in dev

aelkiss:
We rarely rely on the SP to use the configured WAYF session initiator;
rather, the applications redirect to the WAYF as necessary. However,
there are cases when we do, especially with Dex. This change ensures
that we redirect to the WAYF for the vhost we're accessing rather than
always redirecting to https://babel.hathitrust.org


aelkiss:
`ShibRequestSetting discoveryURL` is covered here: https://shibboleth.atlassian.net/wiki/spaces/SP3/pages/2065334723/ContentSettings

skorner (approved):
Thanks for the background context and pointers to the documentation. This seems like an ideal solution to the problem, given our fixed shib config.

PR #545, opened 2022-08-01
Extract client_cert profile

daaang:


antmoth:
?

botimer:
I did not see/remember this one as I built and merged https://github.com/mlibrary/nebula/pull/710.

I think this is a slightly better version, and would not be opposed to merging this and porting the quod usage. However, I will note the naming I went for, which is potentially more generic and accurate -- machine_cert rather than client_cert. That is, we use the machine cert _as_ a client cert, but it is, itself, not a "client cert".

PR #547, opened 2022-10-04
Add missing MARC support perl modules for HT

rrotter:
Install from CPAN: MARC::Record MARC::File::XML


aelkiss (approved):
Looks fine to me!

PR #551, opened 2022-11-02
Add MARC::File::MiJ Perl module for HathiTrust

moseshll:
Once caveat: trying to install this module on dev-2.babel with the `cpan` command resulted in an error seemingly related to the `ex::monkeypatched` dependency (yeah, nothing at all sketchy there). But, I was trying to do a local install on a chunk of filesystem under my control, which is nontypical usage.


rrotter (approved):
We don't have any tests that CPAN will actually run (kind of out of scope, but `install MARC::File::MiJ` in cpan WFM on dev-2. I'm fine with merging this.

PR #552, opened 2022-11-03
Generate etcd yaml with erb

rrotter:
Config is written to /tmp on etcd nodes, iff etcd_initial_cluster var is defined for the present cluster.

This approach requires moving /tmp/etcd.yaml manually. Could also write directly to /etc if we aren't concerned about etcd coming up w/o proper certs (or add additional conditional).

### DIFF
--- a/manifests/profile/kubernetes/bootstrap/etcd_config.pp
+++ b/manifests/profile/kubernetes/bootstrap/etcd_config.pp
@@ -1,9 +1,15 @@
-# Copyright (c) 2020 The Regents of the University of Michigan.
+# Copyright (c) 2020, 2022 The Regents of the University of Michigan.
 # All Rights Reserved. Licensed according to the terms of the Revised
 # BSD License. See LICENSE.txt for details.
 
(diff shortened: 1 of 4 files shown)


rrotter:
Writing the initial_cluster string in common.yaml isn't the most elegant; but I think it's a little more future proof than writing a bash generator script. If we later change how we define clusters it's still valid with little/no reworking.

daaang (approved):
I guess I wrinkle my nose a little at the addition of a long line of ip addresses and ports full of generatable information doubled in hiera just because we can't leave off a trailing comma. It definitely offends my sensibilities, but, in practice, we're probably never actually going to have etcd nodes that aren't 10..14, and anyone changing that won't be under any illusion that puppet is going to magically make it easy, and it already won't make it easy. This is pretty simple, and if we change it for one cluster, it won't break other clusters, **and** it has the benefit of making etcd more easily adaptable if we wanted to, since all we have to do is `cp /tmp/etcd.yaml /etc/kubernetes/manifests` to get a correct config again, should anything change (as oppose to going by hand and editing 5 similar yaml files). Ship it.

PR #553, opened 2022-11-15
Move let's encrypt's files into /l/local/ssl-certs

daaang:


daaang:
To do:

1. Add the current list of 11 certs to hiera
2. Remove all 11 certs from /sysadmin/archive/ssl
3. Document the steps for moving additional incommon certs to let's encrypt and aws.

manifests/profile/certbot_route53.pp
```diff
@@ -0,0 +1,37 @@
+# Copyright (c) 2022 The Regents of the University of Michigan.
+# All Rights Reserved. Licensed according to the terms of the Revised
+# BSD License. See LICENSE.txt for details.
+
+class nebula::profile::certbot_route53 (
+  Hash[String, Array[String]] $certs = {}
+) {
+  $certs.each |$service, $domains| {
+    $domains.each |$domain| {
+      file { "/l/local/ssl-certs/${domain}.crt":
```
daaang, line 10: Parameterize this so it's something like `"${cert_dir}/${domain}.crt"`

manifests/profile/certbot_route53.pp
```diff
@@ -0,0 +1,37 @@
...
+        group  => "puppet",
+        source => "/etc/letsencrypt/live/${domain}/fullchain.pem"
+      }
+
+      file { "/l/local/ssl-certs/${domain}.key":
+        group  => "puppet",
+        source => "/etc/letsencrypt/live/${domain}/privkey.pem"
+      }
+
+      concat { "/l/local/ssl-certs/haproxy/${service}/${domain}.pem":
```
daaang, line 20: Parameterize this too; maybe `"${haproxy_cert_dir}/${service}/${domain}.pem"`?

daaang (commented):
1. I think we should avoid hard-coding `/l/local` where we can. It's not that I consider it proprietary; but rather I'd prefer we move away from it, and parameterizing it is a step in that direction.
2. It would be good if the steps we performed by hand were represented in addition to the work of keeping LE certs in sync. So this profile should install `awscli`, `certbot`, and `python3-certbot-dns-route53`.
3. The files in `/root/.aws` should be distributed by puppet with the secrets in lensoftruth.
4. I'm not sure if there's a way to set up `/etc/letsencrypt` in such a way that we bypass the interactive questions (agreeing to terms and conditions and declining to sign up for EFF news).
5. If we can do 4, then we could add something to make puppet run `certbot certonly --dns-route53 -m "$our_cert_email" -d "$main_domain,*.$main_domain,$second_domain,*.$second_domain,..."`

I think only 1-3 are easy/important enough. 4 and 5 would be more important if we didn't intend to replace the puppetserver with a container when we move it to the macc. But as it is, we don't actually intend to duplicate this setup, so it's not really important that we be able to.

PR #555, opened 2023-01-05
Update general web hosting role/profile for Debian 11 (bullseye) compatibility

skorner:
The changes are limited to what was necessary for bullseye compatibility plus some cleanup that made sense to include in one PR. I split the cleanup work into a separate commit.


daaang (approved):
I read through all this just now (sorry I took so long!). Looks mostly technical, moving shibboleth, java 11, some perl dependency changes, changing a variable name, adding a directory, and removing some dead cosign configuration. I don't see anything that looks insane, so LGTM!

PR #556, opened 2023-01-11
Set ae-files to manage its own certs

daaang:


skorner (approved):
Looks good and makes sense to me. Maybe consider explicitly mentioning in the commit message or PR description that port 80 will be used for LE cert management and 443 for serving the content? It's also fine to me as is though.

rrotter (approved):
anyone know why this wasn't merged?

rrotter:
Discussed on slack. Couple of non-show-stopping issues:

1. I think the hold up was waiting shake out any port 80 dependencies, which is long done in 7 months.
2. This sends some spam from a cron, which is a bug, but it's outside of this PR (in nebula::profile::letsencrypt and/or letsencrypt::renew, I think)

So, merging this.

PR #563, opened 2023-02-17
support for chrony ntp client and ptp_kvm clock

rrotter:
* replace ntpd on kubernetes nodes
* add chrony on prometheus nodes
* use ptp_kvm clock if we install chrony on kvm guest


rrotter:
This was to eliminate clock skew issues we had on systems that are behind nat. This could optionally be applied on all hosts. Unless this gets a wontfix, I'd like this done before puppet upgrade, tagged accordingly.

rrotter:
Seems like the clock skew was fixed by other means.

PR #570, opened 2023-03-30
Allow additional paths for X-SendFile for Fulcrum

botimer:


antmoth:
?

botimer:
This should be irrelevant now, with the old paths configured out and the load balancer / web server switch complete.

PR #571, opened 2023-04-04
Manage openjdk java

umjnega:
Creating a pull request to revisit this and discuss what else would need to be done.


botimer:
This looks like the right approach. We can include the profile in an appropriate role (or app profile). If an application depends on a given JDK version, we can specify it in a puppet file. If a given (grab-bag) host needs additional JDKs or an override, we can do it in hiera.

PR #572, opened 2023-05-01
HathiTrust DEV-731: take down usdocs registry

aelkiss:
Many dependencies for the registry are out of date, the usage is very low, and HathiTrust has agreed to take it down rather than update the dependencies.


manifests/profile/hathitrust/apache.pp
```diff
@@ -110,6 +110,7 @@
   class { 'apache::mod::php':
     extensions => ['.php','.phtml']
   }
+  class { 'apache::mod::proxy': }
```
aelkiss, line 113: The tests throw a bunch of errors without this :shrug: I suspect the `proxy_pass` block below pulled it in, and however it's being used elsewhere (probably via RewriteRules with the P flag) doesn't.

rrotter (approved):
Looks fine, see comment on apps-ht host entry.

aelkiss:
Great. Assuming tests pass I will merge on Monday.

PR #577, opened 2023-05-10
change matomo subdomain from matomo to www.matomo

rrotter:


aelkiss (approved):
I'll wait for CI and then merge.

PR #578, opened 2023-05-17
Add jemalloc and auto-restarts for Fulcrum

botimer:
We have thoroughly proven out jemalloc to mitigate the OOM issues for the app, and it requires no application changes, only environment. We're also adding automatic restarts to the systemd unit in case of other kinds of failure.


botimer:
For references, see HELIO-3772 and https://medium.com/motive-eng/we-solved-our-rails-memory-leaks-with-jemalloc-5c3711326456.

skorner (approved):
As you indicated, jemalloc is proven to stabilize this app's memory use and auto restart on failure for Fulcrum is an essential safety net for unexpected transient problems.

PR #581, opened 2023-05-24
Purge Cosign from the www.lib fallback vhost

skorner:
There is no valid weblogin use case for www.lib, so we can remove it.

Also purge Cosign from mgetit.lib where it was never in use.

### DIFF
--- a/manifests/profile/www_lib/vhosts/www_lib.pp
+++ b/manifests/profile/www_lib/vhosts/www_lib.pp
@@ -21,16 +21,9 @@ class nebula::profile::www_lib::vhosts::www_lib (
     servername                    => "${prefix}www.${domain}",
     ssl                           => true,
     usertrack                     => true,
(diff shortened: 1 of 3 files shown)


manifests/profile/www_lib/vhosts/www_lib.pp
```diff
@@ -21,16 +21,9 @@
     servername                    => "${prefix}www.${domain}",
     ssl                           => true,
     usertrack                     => true,
-    cosign                        => true,
+    cosign                        => false,
     docroot                       => $docroot,
     directories                   => [
-      {
```
skorner, line 27: Removing this is not related to cosign, but rather addressing an oversight when I cleaned up the www.lib vhost for fallback duty last year. There is no CGI within the current library website static build.

daaang (approved):

PR #582, opened 2023-05-24
Switch kubernetes from docker to containerd.io

daaang:


manifests/profile/containerd.pp
```diff
@@ -0,0 +1,27 @@
+# Copyright (c) 2023 The Regents of the University of Michigan.
+# All Rights Reserved. Licensed according to the terms of the Revised
+# BSD License. See LICENSE.txt for details.
+
+class nebula::profile::containerd {
```
daaang, line 5: Part of installing containerd involves altering /etc/containerd/config.toml, at least for kubernetes, so that file should probably be managed by puppet. That said, it's kind of a monster, and I don't know how or whether it varies between hosts, so I'm leaving that for later.

PR #584, opened 2023-06-07
Prevent badrobots etc at location /

aelkiss:
There is one particular IP address that has been issuing a large number of requests like

```
OPTIONS /cgi/pt/error
```

For whatever reason this wasn't being caught by the `DirectoryMatch` directives. This change should ensure that it does. I will verify on preview.babel that this doesn't interfere with any Shibboleth stuff.


aelkiss:
Looks good on preview.babel.hathitrust,.org.

rrotter (approved):
Like much in the puppet Apache module, I don't really understand this, but if I don't see any issues with it.

aelkiss:
@rrotter Could you go ahead and merge?

PR #586, opened 2023-06-20
Modify press_fcgi_check to report total restarts

daaang:
We don't currently know how often this script actually does restart the press service. With this change, the script will increment a counter on the pushgateway every time it actually runs systemctl restart press, and if we're ever interested, we'll be able to see how often it comes up, if ever.


skorner (approved):
Looks good and potentially useful. Thanks!

PR #587, opened 2023-06-22
Add mod_auth_openidc support and migrate from cosign to auth_openidc for www_lib profile and vhosts

skorner:
The auth related changes are in two commits to reflect general mod_auth_openidc support in apache vs changes specific to www_lib.

The press_fcgi_check script commit is included as that change was needed while this branch was live in production.

### DIFF
--- a/templates/profile/apache/auth_openidc.conf.erb
+++ b/templates/profile/apache/auth_openidc.conf.erb
@@ -1,0 +1,38 @@
+OIDCProviderMetadataURL <%= @oidc_metadata %>
+OIDCClientID <%= @oidc_client_id %>
+OIDCClientSecret <%= @oidc_client_secret %>
+OIDCCryptoPassphrase <%= @oidc_crypto %>
+
+# Essential, as the default client_secret_basic is not supported by ITS' shib OIDC provider config
+OIDCProviderTokenEndpointAuth client_secret_post
+
+# Defaults to code, but be explicit for transparency
+OIDCResponseType code
+
+OIDCScope "openid email profile"
+OIDCRemoteUserClaim preferred_username
+
+# Always protect the default redirect URI in all vhosts
+<Location /openid-connect/callback>
+    AuthType openid-connect
+    Require valid-user
+</Location>
+
+# Setup file-based session cache
+OIDCCacheType file
+OIDCCacheDir "/var/cache/apache2/mod_auth_openidc/oidc-sessions"
+OIDCCacheFileCleanInterval 60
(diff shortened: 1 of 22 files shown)


skorner:
This branch has been running in production for a week now with no problems reported. 

A corresponding LoT branch must be merged at the same time to make available required OIDC settings. 

After merging, the production web servers need to be moved back to the regular puppet env. In addition, the cosign module and cosign-related cron cleanup job must be manually removed from the web servers. The module will no longer be active at the time.

templates/profile/apache/auth_openidc.conf.erb
```diff
@@ -0,0 +1,38 @@
...
+
+# Indicates whether POST data will be preserved across authentication requests.
+# Preservation is done via HTML 5 local storage. Note that this can lead to private
+# data exposure on shared terminals, that is why the default is "Off". Can be
+# configured on a per Directory/Location basis.
+OIDCPreservePost On
+OIDCRefreshAccessTokenBeforeExpiry 300 logout_on_error
+
+# A redirect URI must be fined to load the module successfully
+OIDCRedirectURI "https://something.lib.umich.edu/openid-connect/callback"
```
botimer, line 38: This looks configured for all locations where the handler is configured, but I wonder if it would be better to set this default to a static page somewhere with a misconfiguration message.

botimer (approved):
Really nice. I understand what's going on here and I don't see any problems. It uninstalls the cosign module, switches config to oidc, unrolls the protected directories that had a special config hash, and specifies the callbacks for each site. 🚢

templates/profile/apache/auth_openidc.conf.erb
skorner: I like the idea, but I'm not sure it would work as intended. The module won't check if there is actual content at the redirect URI (which it expects there isn't) before including it in the authn request and sending the user off to the IdP. The IdP will just throw an error about an unregistered redirect URI rather than send the user to it.

If there is no OIDCRedirectURI and the module is active in the config, Apache will fail to start. My rationale for including a false entry here was to allow mod_auth_openidc to be loaded into an apache config without requiring it be active for any vhost (for whatever reason). That seemed harmless, but I suppose both ways can lead to surprises.

botimer: Ahh, yes. In that case, I think we should make the URL effectively the error message, then. Something like https://missing-callback.invalid/check-the-apache-oidc-configuration. Probably won't come up, but a good general habit over "foo" values.

PR #588, opened 2023-06-22
add bookworm support

rrotter:
merged the preseed. Puppet doesn't yet have a bookworm dist in their app repo. I think I'll wait on that for merging for now.


rrotter:
Closing this. Still no upstream bookworm support in puppet for bookworm. We'll start over when that comes.

PR #589, opened 2023-06-30
Always trust umich::networks::all_trusted_machines for SSH

daaang:
Not sure why, but the `is_publicly_accessible()` function appears to be flapping, which occasionally causes inaccessibility due to machines incorrectly reporting to be private-only machines.

This way, all machines will always trust ssh coming from trusted CIDRs. It should have no impact on private hosts, since they're already behind a NAT router that should make them unroutable.

And if public-ip-address hosts occasionally flap and get a few 10-net ssh allowances? No harm done, really.


botimer (approved):
Seems fine, though I think we could drop the conditional entirely -- always allowing private bastion host ssh seems consistent and acceptable.

PR #592, opened 2023-07-19
DEV-864 - install current lts node for ht dev hosts

aelkiss:
We need node/npm installed to support building the production css/js components, but do not currently need/want it in production.

nebula::role::webhost::htvm::test is the role used by the dev/test/preview HT web VMs.


aelkiss:
Working on test.babel. Please merge ASAP.

PR #593, opened 2023-07-21
HT DEV-849 - run wordpress cron from cron

aelkiss:
* Runs wordpress cron once an hour
* Hits production vhost via localhost (htaccess restricts access to wp_cron.php from elsewhere)
* redirect from http://old.www.hathitrust.org -> https://old.www.hathitrust.org


aelkiss:
Please plan to merge this on 7/26 between 09:00 and 10:00 EDT. Thanks

PR #595, opened 2023-08-01
Move wordpress cron to global primary

aelkiss:
It needs to run on only one machine globally, and does not make sense to merge with any of the existing lists of cron jobs that run in that fashion.

### DIFF
--- a/spec/classes/profile/hathitrust/cron/mdp_misc_spec.rb
+++ b/spec/classes/profile/hathitrust/cron/mdp_misc_spec.rb
@@ -57,12 +57,6 @@ describe 'nebula::profile::hathitrust::cron::mdp_misc' do
                   ],
                   minute: 30)
         end
-
-        it do
-          is_expected.to contain_cron('wordpress cron')
-            .with(command: %r{.*wp-cron.php.*},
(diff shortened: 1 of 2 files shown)


aelkiss:
Tested on nas-macc / nas-ictc. @daaang  Please merge ASAP once tests pass.

manifests/role/webhost/htvm/global_primary.pp
```diff
@@ -10,4 +10,11 @@
   include nebula::role::webhost::htvm::site_primary
   include nebula::profile::hathitrust::cron::statistics
   include nebula::profile::hathitrust::cron::catalog
+
+  cron {
+    'wordpress cron':
+      user    =>  'nobody',
```
daaang, line 16: Looks like this is changing from libadm to nobody, which I assume is fine.

daaang (approved):

manifests/role/webhost/htvm/global_primary.pp
aelkiss: Yeah, I figure we can generally rely on `nobody` being present, and that's the user the web server runs as anyway.

PR #596, opened 2023-08-09
add support for certbot cloudflare plugin

rrotter:


umjnega:
Looks good to me.   Though i'm still getting the hang of puppet code.

PR #597, opened 2023-08-15
Begin removing offensive master/slave terminology

aelkiss:
MariaDB 10.5 uses "replica" as a synonym for "slave", but does not remove the offensive terminology in all locations.


estelendur (approved):
yay!

PR #601, opened 2023-08-25
Allow Search to manage its own auth via OIDC

botimer:
There were some odd conflicts for Search, where people were being taken through reauthentication. It has something to do with multiple data centers and server-side sessions/caches. The app can now manage OIDC logins, so we are bypassing the module here.


skorner (approved):
Looks good! These changes do exactly what was said in the commit and PR message: remove apache-level authn/z for search.lib so that the application has full control to manage authn/z.

PR #604, opened 2023-08-26
update actions/checkout, add dependabot

rrotter:
- actions/checkout v2 => v3
- dependabot for gh workflows
- added parallel_tests gem, make `bundle exec rake parallel_spec` the default for ci

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,7 @@
+---
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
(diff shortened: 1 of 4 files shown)


rrotter:
This should stop GH actions complaining that we're trying to run an out of date version of node.js (which is used by actions/checkout). I can split out parallel_spec into a separate PR if preferred, I lumped together these two small changes last week since it's updating the behavior of the ci action.

botimer (approved):

PR #606, opened 2023-08-30
Test workflow dont merge

rrotter:
testing the behavior of conditionals on gh action.


rrotter:
Great, looks like the new labels are recognized on each push, but not on a re-run. That's reasonable enough.

PR #607, opened 2023-08-30
Test faster

rrotter:
parallel_test gives each thread an equal number of files to test, so extremely large/slow spec files prevent the tests from getting much parallel speed-up. With that in mind, I split up the tests in the two slowest running spec files: htvm_webhost_spec and all_roles_spec

- htvm_webhost_spec: much of this was effectively testing the profiles required by this role, so I created new profile specs and moved the tests there, and moved tests to existing profile specs where possible. There is more that could be done, I only moved the bits that were obvious, but this is still cuts roughly 20% of the runtime off this file and (I hope) makes the spec a little easier to understand.
- split all_roles_spec into 5 slices: make the logic in all_roles_spec into a function, call it from 5 new spec files

### DIFF
--- a/spec/classes/all_roles_spec.rb
+++ b/spec/classes/all_roles_spec.rb
@@ -37,46 +37,55 @@ module RSpec::Puppet
   end
 end
 
-`find manifests/role -name '*.pp'`.each_line do |file_path|
-  role_name = puppet_role_name_from(file_path)
-  describe role_name do
-    on_supported_os.each do |os, os_facts|
-      context "on #{os}" do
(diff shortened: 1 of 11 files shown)


spec/classes/all_roles_spec.rb
rrotter, line 1: wrapped this all in a block, no real changes to the logic, just an added layer of indentation

aelkiss (approved):
All makes sense to me.

daaang (approved):
Hecking rad

PR #609, opened 2023-09-05
add firewall $advanced_rules

rrotter:
This adds $advanced_rules to the firewall profile. These rules get evaluated without any defaults, which allows us to leave some fields empty.

Original post below the fold.

> I'm not going to merge this as is, clearly. Is there any way to have an entry in $rules that ignores everything in $defaults? Two of the three default fields must be empty (not merely different) for a rule I need to add. Tried undef, empty string, and the puppet Default type, none of these worked.
> 
> Worst case, I'll rename $better_tests to something, uh, better, add some tests, and make this a real PR.
> 
> > Error: Failed to apply catalog: Only one of the parameters 'action' and 'jump' can be set
> 
> > Error: Failed to apply catalog: Parameter action failed on Firewall[999 mysql port forwading]: Invalid value "undef".
> 
> > Error: Failed to apply catalog: Parameter action failed on Firewall[999 mysql port forwading]: Invalid value "".
> 
> > Error: Failed to apply catalog: Parameter action failed on Firewall[999 mysql port forwading]: Invalid value "default".

### DIFF
--- a/manifests/profile/networking/firewall.pp
+++ b/manifests/profile/networking/firewall.pp
@@ -14,6 +14,7 @@
 class nebula::profile::networking::firewall (
   String $internal_routing = '',
   Hash $rules = {},
+  Hash $advanced_rules = {},
 ) {
   # Include standard SSH rules by default
   include nebula::profile::networking::firewall::ssh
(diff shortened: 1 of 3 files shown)


daaang (dismissed):
As you already understand, `better_rules` is not a good name. I suggest either going with Microsoft and `advanced_rules` or with Linux and `verbose_rules`, for this pattern.

Obviously I'd rather not add even more "firewall in hiera" infrastructure, but this is far from the worst offender, and we need it ASAP. When the time comes to replace musar, I will want its replacement to be configured differently.

rrotter:
Used a better name for the var, added tests, renamed the PR appropriately.

PR #610, opened 2023-09-06
optimal ci runner sizing

rrotter:
Pushed a few times with different config to benchmark runners of various sizes.
A "fast" run now cost ~0.15 instead of ~0.25, and takes ~5m instead of ~4m.

Keeping the label for fast running because often we really don't care if it's slow, and I don't want anyone to think twice when making a commit because it costs money, even it the money involved is trivial.

Exact results results below in comments.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,7 +7,7 @@ on:
 
 jobs:
   test:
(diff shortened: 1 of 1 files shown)


rrotter:
3:51 on 16 cores, 10:51 on 2 cores. Still taking about 60s to download deps, that hurts.

... and 4:49 on 8 cores. I think we have a winner.

PR #613, opened 2023-09-18
Add public/private awareness to prometheus node service discovery

daaang:
Until now, prometheus's profile has only been aware of IP addresses in the sense of "every host has exactly 1 IP address, and it's public." That assumption is no longer true; some hosts have 0 public addresses and 1 private address, and most hosts have 1 public address and 1 private address.

1. When the prometheus server is crossing datacenters, it must use a **public** IP address.
2. When the node has 0 public addresses, the prometheus server must use a **private** IP address.
3. The prometheus server should prefer to use a **private** IP address when possible.
4. The prometheus server must be able to fall back to a **public** IP address if there is no available private address.

### DIFF
--- a/manifests/profile/prometheus.pp
+++ b/manifests/profile/prometheus.pp
@@ -131,18 +131,69 @@ class nebula::profile::prometheus (
     block => 'umich::networks::all_trusted_machines',
   }
 
-  @@concat_fragment { "02 pushgateway advanced url ${::datacenter}":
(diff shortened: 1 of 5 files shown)


daaang:
I think this is basically done except that it won't work for pushgateways anymore. I should probably do that with an IP address instead of an FQDN, and then it's just a matter of exporting and opening to the right IP address with that.

estelendur (approved):
Jody and I looked at these with Matt and everything seemed fine. shipit

PR #614, opened 2023-09-18
Update modules to support Puppet 7

rrotter:


rrotter:
Can't update puppetlabs/apache past 7.0.0, it breaks hundreds of tests, probably thanks to these:

https://github.com/puppetlabs/puppetlabs-apache/pull/2267
https://github.com/puppetlabs/puppetlabs-apache/pull/2253

This doesn't block Puppet 7, but will block Puppet 8.

All other modules should be on the final version that supports Puppet 6.

manifests/profile/work_around_puppet_bugs.pp
```diff
@@ -1,34 +0,0 @@
...
-  # This was allegedly fixed in puppet 5.5.7, so we should be able to
-  # delete this once we aren't running any older versions of puppet.
-  #
-  # The bug is that the state.yaml file tracks every resource the agent
-  # has ever seen, even if it only existed, say, as part of a tidy
-  # resource saying to delete the file. So this file can get bloated on
-  # servers that do a lot of tidying.
-  #
-  # When state.yaml gets too big, it starts taking a long time (e.g.
-  # sometimes over 35 minutes) for agents to perform what should be a
```
daaang, line 24: Ah memories

daaang (approved):
Thank you for doing the tedious work of clicking around the forge website 28 times and then doing whatever's needed to pass the tests. Also, good catch on removing that old state file workaround.

PR #616, opened 2023-09-20
update dependencies post puppet7 upgrade

rrotter:
Moves tests to ruby 2.7 and puppet 7, update modules.


rrotter:
This is as far as we can update modules with little to no code. Critically, it resolves a conflict introduced during the puppet7 update between the lvm and stdlib module versions.

metadata.json
```diff
@@ -7,34 +7,34 @@
...
-    {"name": "puppetlabs/mysql",        "version_requirement": ">= 13.3.0 < 14.0.0"},
-    {"name": "puppetlabs/ntp",          "version_requirement": ">= 9.2.2  < 10.0.0"},
-    {"name": "puppetlabs/postgresql",   "version_requirement": ">= 8.3.0  < 9.0.0"},
-    {"name": "puppetlabs/puppetdb",     "version_requirement": ">= 7.13.0 < 8.0.0"},
-    {"name": "puppetlabs/reboot",       "version_requirement": ">= 4.3.1  < 5.0.0"},
-    {"name": "puppetlabs/sshkeys_core", "version_requirement": ">= 2.4.0  < 3.0.0"},
-    {"name": "puppetlabs/stdlib",       "version_requirement": ">= 8.6.0  < 9.0.0"},
-    {"name": "saz/resolv_conf",         "version_requirement": ">= 5.1.0  < 6.0.0"},
-    {"name": "stm/debconf",             "version_requirement": ">= 5.0.0  < 6.0.0"}
+    {"name": "jdowning/rbenv",          "version_requirement": "3.0.0"},
```
botimer, line 10: I'm not entirely clear on the change from optimistic ranges to exact values. The modules have been very reliable in flagging compatibility issues with major versions and this change would eliminate any automatic updates we are currently receiving. I do like the synchronization with the fixtures.

Have you outlined any thoughts on inventory/updates with exact versions?

rrotter: I've been bitten by breaking changes introduced in modules on minor or teeny versions in the last two days. apt 9.0.1 doesn't even compile unless stdlib is updated to a version newer than what we're using and *much* newer than the version it claims to support.

rrotter: TBH this change was half due to the problems with puppetlabs/apt and half because the syntax was tedious (no `~>`!), which isn't a great reason. I also didn't realize that librarian-puppet was actually optimistically grabbing the newest rather than oldest compatible version.

I'll limit this stricture to where there is a known breaking change.

rrotter:
Don't merge until all hosts are on puppet7+

estelendur (approved):

PR #618, opened 2023-09-21
Switch chipmunk from cosign to mod_auth_openidc

botimer:


botimer:
I believe this still needs a www-data chown on the session cache. I'm not sure whether that needs to be a parameter or can be literal.

skorner:
I addressed the user/group issue by specifying that the dir be owned by the apache user/group info available via the puppet module. I'm going to close this PR and start a new one though.

PR #619, opened 2023-09-22
add librarian-puppet to ci

rrotter:


botimer (approved):
I like this as a step in the regular CI and especially that the step is conditional on dependency changes.

PR #620, opened 2023-09-28
Switch Dark Blue (chipmunk) from cosign to mod_auth_openidc and purge all vestigial cosign support in nebula

skorner:
There is one change that will affect any roles that use the auth_openidc profile. Specifically, this PR includes a change to the auth_openidc profile to ensure the auth_openidc cache directory is owned by the apache user:group rather than hard coded, which should not be appropriate in any use case.

Purging all remaining cosign support is included in this PR, as chipmunk was the last cosign use case in nebula. Much of the cleanup is within named_instance, which will be more comprehensively purged soon, but that's OK.


manifests/profile/hathitrust/apache/babel.pp
skorner, line 1: @aelkiss FYI since this change touches a hathitrust profile, but is presumably of no impact.

botimer (approved):
This looks good to me. The commented-out firewall include might be a little confusing, but as you say, that entire profile is going away shortly.

PR #623, opened 2023-10-13
kubelet deps

rrotter:
Install dependencies and configuration to allow initialization of new vms with kubelet on Debian Bullseye


rrotter:
Test k8s cluster nodes seem happy with this so far. Fine after applying, still fine after a reboot of the nodes (tried at least one each of controller, etcd, and worker nodes).

rrotter:
This actually wasn't working on the worker nodes. Added the full default `containerd` config and squashed.

PR #624, opened 2023-10-19
Add unattended upgrades to kubernetes private hosts

daaang:


daaang:
This is running on the first two etcd (10, 11), controller (20, 21), and worker (30, 31) nodes on macc-test-kubernetes. Next week before merging, take a look at `/var/log/apt/history.log` for those versus their non-auto-upgrading peers for a sense of the changed behavior.

rrotter (approved):

PR #625, opened 2023-10-19
define private_address_template only once for HT

rrotter:
`private_address_template` is defined all over the place, but it's the same for all HathiTrust hosts. Allow it to be defined in one shared profile.

The advantage of this isn't so much that we get to delete a few lines from the hiera data (which is nice), but that we won't have to add another for each future VM type.


aelkiss (approved):
Seems like a clear win to me. I don't anticipate any reason why we would need this to vary between role in the future (and if we did, we have a lot of ways we could override this per host or per role if needed.)

PR #626, opened 2023-10-25
Support standalone server certificate issuance

botimer:
Here we extend the profile for managing certificates with Certbot and Route 53. The existing "certs" parameter expects there to be an HAProxy service in place and handles the combined PEM file. The new "simple_certs" paramter supports standalone hosts where we only need the .crt and .key files in "ssl-certs".

Both parameters support SANs, so the main CN is structured as a hash. If there are no SANs, simply supply an empty array. One difference to note is that the "certs" behavior automatically adds a wildcard to the primary domain (CN). This is not assumed for simple_certs.


umjnega (approved):
Looks good to me.   Nice explainations along the way.

PR #630, opened 2023-11-02
nebula::exposed_port: handle duplicate blocks

rrotter:
At some point we decided it was important that if we use network block lists like `hathitrust::networks::staff` somewhere other than apache, that we properly handle the case that a block is listed twice:

https://github.com/mlibrary/nebula/blob/962b9b213a0fadd474af7e0aa0e13ecd53815dd6/spec/fixtures/hiera/hathitrust.yaml#L44-L60

I actually need to use `hathitrust::networks::staff` with `nebula::exposed_port`. So, in an effort to _not break_ this canary/test, I added support for duplicate blocks in `nebula::exposed_port`

- `nebula::exposed_port` merges identical blocks
- will still fail if a block name is repeated, but with different content

### DIFF
--- a/manifests/exposed_port.pp
+++ b/manifests/exposed_port.pp
@@ -60,7 +60,7 @@ define nebula::exposed_port(
   String $block,
   String $protocol = 'tcp',
 ) {
-  lookup($block).flatten.each |$cidr| {
+  lookup($block).flatten.unique.each |$cidr| {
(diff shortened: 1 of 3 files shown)


daaang (approved):
Nice tests, and yeah I can't think of any way this could blow up. Seems about as safe as could be.

PR #633, opened 2023-11-06
Silence letsencrypt cron emails by directing them to logs

daaang:
This should append `2>&1 | logger -t letsencrypt-renew` to all cert renewal cronjobs.


skorner (approved):
This seems like the best practice option: output goes to the system logger where it can be accessed and managed like all other logs rather than vanishing or filling Inboxes and mail spools.

PR #634, opened 2023-11-07
Remove an old unneeded puppetdb query

daaang:


skorner (approved):
We've confirmed this is not needed, and even if it was we'd need to implement it differently.

PR #639, opened 2023-11-22
Add kubelet and ipmi exporter to sysadmin boxes

daaang:
This also creates a generic kubelet profile outside kubernetes, and much of kubernetes's kubelet resources have been replaced with this profile. It's not a complete rewrite, but it's a start.

This should definitely be tested in a development environment before merging.

On the way, this should also fix the naming collision between two different prometheus node exporter packages by pinning to a specific version.


botimer (approved):
I haven't traced all of the port stuff super closely, but it all looks handled. Node and IMPI exporters on separate ports; firewall for each; preferring private IP; adding scrape config... Ship it!

PR #642, opened 2023-12-11
Fulcrum demo

estelendur:
Get fulcrum standalone app host role working


botimer (changes requested):
We have a little bit of work to do before this can be merged... We'll need to ensure that the www-lib servers are ready for the Adoptium change. There is also some fixup for the Shibboleth config.

skorner:
I noticed in the tomcat logs (/var/log/fedora) that tomcat hasn't been able to unpack the fedora war file due to the perms on the webapps directory.

This reason I was looking around is that before this update is final, we'll also need a way to specify that the fedora binary storage is on the NAS volume while the other aspects of running fedora are on local disk. There is a hook for the binary storage directory in repository.json which presumably pulls a definition from /etc/default/fedora.

PR #644, opened 2023-12-18
Rename MysqlReplicationNotRunning alert

daaang:
MysqlReplicaSqlDown was confusing to at least one person (me!), who thought it meant "the database is down."


umjnega:
I like the change of wording.  Makes sense to me.

PR #645, opened 2023-12-21
Force login for auth system user management app

botimer:
This was another case of passive auth running afoul of authz_umichlib.


skorner (approved):
That should be what this app needs to work again. Thanks!

PR #650, opened 2024-01-10
drop www to match cert file name

umjnega:


manifests/profile/www_lib/apache/misc.pp
```diff
@@ -44,7 +44,7 @@
       'www.press.umich.edu',
       'www.publishing.umich.edu',
       'www.theater-historiography.org',
-      'www.heartofdarknessarchive.org',
+      'heartofdarknessarchive.org',
```
skorner, line 47: A minor nit...I'd move this up alphabetically now that www is gone.

skorner (approved):
Assuming tests pass and the order change, looks good.

PR #652, opened 2024-01-23
Use PROXY protocol from gateway to ingress controllers

botimer:
This sends the original client IP in additional metadata outside of the HTTP request. It must be "accepted" by the ingress controller to function.

See also:

* https://www.haproxy.com/documentation/haproxy-configuration-tutorials/client-ip-preservation/enable-proxy-protocol/
* https://docs.nginx.com/nginx/admin-guide/load-balancer/using-proxy-protocol/
* https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/configmap/#use-proxy-protocol

### DIFF
--- a/spec/classes/profile/kubernetes/destination_port_spec.rb
+++ b/spec/classes/profile/kubernetes/destination_port_spec.rb
@@ -9,7 +9,6 @@ require 'spec_helper'
   ['api',        6443],
   ['etcd',       2379],
   ['http',      30080],
-  ['https',     30443],
   ['https_alt', 31443],
   ['gelf_tcp',  32201],
 ].each do |service, port|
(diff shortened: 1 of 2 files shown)


aelkiss:
Nice, very excited about this if it means that the ingress controller (& ultimately downstream services) can log the original client IP

spec/classes/profile/kubernetes/destination_port_spec.rb
```diff
@@ -36,3 +35,30 @@
     end
   end
 end
+
+[
+  ['https',     30443],
+].each do |service, port|
+  describe "nebula::profile::kubernetes::destination_port::#{service}" do
```
daaang, line 42: big oof

daaang (approved):
Nice!

spec/classes/profile/kubernetes/destination_port_spec.rb
botimer: Yes, I share this oof. Left as an open question of how best to deal with a now-leaky abstraction.

PR #654, opened 2024-01-26
Make HAProxy better at knowing when to pull kubernetes controller nodes from load balancing

daaang:
HAProxy's TCP check must be essentially no more than `nc -v`, which was succeeding on a controller node that was very much unavailable. `curl -k` (and `ssh`) was timing out, so this pull request aims to make HAProxy do something closer to `curl -k`. When controller nodes are up and working, as of writing, `curl -k https://addr:6443/` without any attempt at valid client certs yields a prompt `403` status.

### DIFF
--- a/templates/profile/kubernetes/haproxy/services.d/api.cfg.erb
+++ b/templates/profile/kubernetes/haproxy/services.d/api.cfg.erb
@@ -7,5 +7,6 @@ frontend kubernetes-api-front
 
 backend kubernetes-api-back
   mode tcp
-  option tcp-check
   balance roundrobin
+  option httpchk HEAD /
+  http-check expect status 403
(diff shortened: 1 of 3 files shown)


templates/profile/kubernetes/haproxy/services.d/api.cfg.erb
```diff
@@ -7,5 +7,6 @@ frontend kubernetes-api-front
 
 backend kubernetes-api-back
   mode tcp
-  option tcp-check
   balance roundrobin
+  option httpchk HEAD /
+  http-check expect status 403
```
daaang, line 12: I'm not especially happy running essentially `curl -k` and considering `403 Forbidden` to be "success," but the aim here is not to test our PKI but rather to check that a server is ready to receive requests, and if not, remove it from load balancing. Assuming that we **don't** have valid certs, 403 is in fact the expected behavior.

If some later version of kubernetes decides to switch to `401` or something, I believe it will be impossible not to notice in the test cluster.

PR #658, opened 2024-02-09
hathitrust catalog and lss roles

rrotter:
Reworked the solr-vm branch so these two roles can share most of their code, and also ease setup of future roles without AFS.


rrotter:
wip, still need to do manual testing and add catalog release scripts.

files/solr_lss/conf/1000common.txt
aelkiss, line 1: I'm assuming you removed all these because they're in `/htsolr/lss/shared` and don't need to be managed with puppet?

manifests/profile/hathitrust/solr6/catalog.pp
```diff
@@ -0,0 +1,62 @@
...
+  $core_data_dir_template = 'data'
+  $core_link_prefix = ""
+  $is_catalog = true
+  $is_primary_node = true # catalog solr is only one node per site
+  file { "/usr/local/bin/index-release":
+    owner   => "root",
+    mode    => "755",
+    content => template("nebula/profile/hathitrust/solr6/index-release.sh.erb"),
+  }
+  if ($is_primary_site) {
```
aelkiss, line 50: This appears to be an accurate translation of what's happening currently...

templates/profile/hathitrust/solr6/index-release.sh.erb
```diff
@@ -0,0 +1,105 @@
...
+  fi
+done
+
+# check today's snap for 'busy' flag
+if [ -f $BASE/flags/${SNAP}/busy ]; then
+  /bin/echo "indexing and/or optimization ocurring...skip release today"
+  exit 1
+fi
+
+<% if not @is_primary_site %>#<% end %># check whether release happened at mirror site
```
aelkiss, line 31: I would prefer structuring these all as e.g.
```bash
<% if not @is_primary_site %>
# check whether release happened at mirror site
if ! curl -A SOLR -s --retry 5 --fail https://babel.hathitrust.org/flags/web/<%= @solr_name %>-release-${TODAY} --resolve "babel.hathitrust.org:443:<%= @mirror_site_ip %>" >& /dev/null; then
  /bin/echo "Mirror site index release appears to have failed...skip release today"
  exit 1
fi
<% end %>
```
rather than repeating the check on every line and adding commented-out things in the output file.

templates/profile/hathitrust/solr6/log4j.properties.erb
```diff
@@ -1,5 +1,6 @@
+# Managed by puppet (nebula/profile/hathitrust/solr6/log4j.properties.erb)
 #  Logging level
-solr.log=${solr.solr.home}/../logs
+solr.log=<%= @solr_home %>/logs
```
aelkiss, line 3: Might as well -- reduces the indirection and makes this much clearer when actually looking at the realized config

aelkiss (commented):
All looks fine to me. The only thing I want to change is the conditionality in `index-release.sh.erb` so that we condition entire blocks rather than comments for individual lines.

templates/profile/hathitrust/solr6/index-release.sh.erb
rrotter: Yeah, this logic tortured. Pushed at update to change it.
IIRC I was trying to make the output have a consistent line count to make it easier to diff in development.

aelkiss: Thanks, looks much clearer now.

PR #659, opened 2024-02-13
move shib require to top location block

umjnega:
Went ahead and looked into the fulcrum issue from yesterday and I think I correclty updated the change @skorner made.

### DIFF
--- a/manifests/profile/www_lib/vhosts/fulcrum.pp
+++ b/manifests/profile/www_lib/vhosts/fulcrum.pp
@@ -19,11 +19,13 @@ class nebula::profile::www_lib::vhosts::fulcrum (
   String $servername = 'www.fulcrum.org'
 ) {
   $authz_base_requires = {
-    enforce  => 'all',
-    requires => [
+    auth_type       => 'shibboleth',
+    custom_fragment => 'ShibRequestSetting requireSession 0',
+    enforce         => 'all',
+    requires        => [
       'not env badrobot',
       'not env loadbalancer',
-      'all granted'
+      'shibboleth'
     ]
   }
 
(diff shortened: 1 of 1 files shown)


manifests/profile/www_lib/vhosts/fulcrum.pp
```diff
@@ -154,12 +154,6 @@
       XSendFile on
       XSendFilePath ${derivatives_path}
       XSendFilePath ${alt_derivatives_path}
-      # Configure Shibboleth for authentication via InCommon partner login
-      <Location "/">
-        AuthType shibboleth
-        ShibRequestSetting requireSession 0
```
skorner, line 160: These two lines need to be present in the rendered `Location /` definition (see the diff I posted in slack). You can do this with an auth_require and custom fragment line, as is done [here](https://github.com/mlibrary/nebula/blob/cdc18854b8ab966eeed0f4b360b58474d86be032/manifests/profile/www_lib/vhosts/staff_lib.pp#L56-L60) (ignore the auth_require in that example as that is already covered with the require all definition in this manifest). You can probably make this addition to authz_base_requires.

skorner (changes requested):
See the comment I added that highlights one more change we have to address for this to work.

manifests/profile/www_lib/vhosts/fulcrum.pp
umjnega: auth_require or auth_type...  I may have pushed my last commit the opposite.  But i'll let you correct me if i'm reading this wrong.

skorner (approved):

manifests/profile/www_lib/vhosts/fulcrum.pp
skorner: Looks good to me! Go ahead and squash to 1 commit and merge after the tests pass from the squash.

PR #660, opened 2024-02-15
Move global shib settings out of the require block

skorner:
The first pass at restoring "badrobots" functionality inadvertently had the global settings to enable shib passively for the Fuclrum vhost within the  config for a require block. This resulted in these settings being ignored and not applied. The changes in this PR address that issue.


skorner:
That's correct that whether those two lines are before or after the RequireAll block doesn't affect the outcome.

PR #662, opened 2024-02-26
update sali1 to sali

umjnega:
I think this is what albert is asking for.  in the TDX ticket.

### DIFF
--- a/manifests/profile/www_lib/vhosts/apps_lib.pp
+++ b/manifests/profile/www_lib/vhosts/apps_lib.pp
@@ -141,27 +141,6 @@ class nebula::profile::www_lib::vhosts::apps_lib (
         require         => 'valid-user',
         custom_fragment => 'OIDCUnAuthAction auth true'
       },
-      {
-        provider        => 'locationmatch',
-        path            => '^/instruction/request',
-        custom_fragment => @(EOT)
-          # Set remote user header to allow app to use http header auth.
-          RequestHeader set X-Remote-User     "expr=%{REMOTE_USER}"
-          RequestHeader set X-Authzd-Coll     %{AUTHZD_COLL}e
-          RequestHeader set X-Public-Coll     %{PUBLIC_COLL}e
-          RequestHeader set X-Forwarded-Proto 'https'
-          RequestHeader unset X-Forwarded-For
-          Header set "Strict-Transport-Security" "max-age=3600"
-        | EOT
-      },
-      # This must be declared after the above block or it will be superseded. 
-      {
-        provider        => 'location',
-        path            => '/instruction/request/login',
-        auth_type       => 'openid-connect',
-        require         => 'valid-user',
-        custom_fragment => 'OIDCUnAuthAction auth true'
-      },
     ],
 
     # TODO: hopefully these can all be removed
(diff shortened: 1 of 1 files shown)


manifests/profile/www_lib/vhosts/apps_lib.pp
```diff
@@ -231,7 +231,7 @@
     ssl_proxy_machine_cert        => $client_cert,
 
     custom_fragment               => @(EOT)
-      ProxyPassReverse / https://sali1.lib.umich.edu:8443/
+      ProxyPassReverse / https://sali.lib.umich.edu:8443/
     | EOT
```
skorner, line 235: We shouldn't need any of lines 228-235, given we are not proxying to an app anymore.

manifests/profile/www_lib/vhosts/apps_lib.pp
```diff
@@ -192,7 +192,7 @@
       },
       {
         rewrite_cond => '%{REQUEST_URI} !^/openid-connect',
```
skorner, line 194: This condition can be removed now. It's probably also best to add a comment about what we are redirecting to (a standalone drupal-based SALI.

manifests/profile/www_lib/vhosts/apps_lib.pp
skorner, line 1: Lines 144-164 are also no longer needed.

skorner (changes requested):
I left some notes about more we can/should cleanup. The changes should probably be tested in a puppet env on macc-www-lib-testing first with Albert verifying the behavior is as expected (he'll need a local hosts alias of apps.lib to 141.213.128.250 to test).

manifests/profile/www_lib/vhosts/apps_lib.pp
umjnega: Ah, ok I misunderstood.  I Thought Albert was asking that we proxy back to sali.lib now that its hosted in AWS.

manifests/profile/www_lib/vhosts/apps_lib.pp
umjnega: Removed and comment added.

manifests/profile/www_lib/vhosts/apps_lib.pp
umjnega: removed this section as well.

umjnega:
If this passed,  I can apply it to macc-www-lib-testing and see if albert can test it.

manifests/profile/www_lib/vhosts/apps_lib.pp
```diff
@@ -190,10 +169,8 @@
         # 2016-08-29 skorner per nancymou
         rewrite_rule => '^/islamic	http://guides.lib.umich.edu/islamicmss/find 	[redirect=permanent,last]'
       },
-      {
-        rewrite_cond => '%{REQUEST_URI} !^/openid-connect',
-        rewrite_rule => '^(/instruction/request.*)$ https://sali1.lib.umich.edu:8443$1 [P]',
-      },
+
+        # No More SALI redirect in favor of stand-alone drupal app
```
skorner, line 173: We still need to include a basic redirect to the new site at the former URL root:
`rewrite_rule => '^(/instruction/request.*)$ https://sali.lib.umich.edu/ [redirect=permanent,last]'`

skorner (approved):
That's great news. This all looks good to me. Feel free to merge next week after squashing to 1 commit and for good measure changing the PR title from "update sali1 to sali" to something like "replace apache config for locally hosted sali with a redirect".

PR #663, opened 2024-02-26
test all roles include nebula::role::minimum

rrotter:
Test that all roles include nebula::role::minimum, and that all HT roles include nebula::role::hathitrust. In my testing this only adds a few seconds to the test run.


botimer (approved):
This looks fine to me, but I am curious: what prompted this additional test after all this time? Did you stumble on an omission error that you want to guard against?

rrotter:
It was an omission in a new role that isn't merged yet. Didn't want to step on that rake twice.

PR #666, opened 2024-02-28
update yard

rrotter:
fixes https://github.com/mlibrary/nebula/security/dependabot/5
edit: and https://github.com/mlibrary/nebula/security/dependabot/6


umjnega:
Looks ok to me.   Patching is always good.

PR #668, opened 2024-03-01
allow blocking path via haproxy

rrotter:


rrotter:
Not needed atm. We can revive this branch later.

PR #669, opened 2024-03-01
Add MysqlTooManyConnections alert

daaang:
I'm starting with `for: 5m` because I'm also setting the severity to ticket, so we don't mind if it's a bit noisy at first. Later we can tweak as needed and consider upgrading its severity to page.


skorner (approved):
This looks like a great way to start and see how it goes for us. Thanks!

PR #676, opened 2024-03-20
update domain to drop the www

umjnega:


skorner (approved):
Looks good and we get one more standardized! 

It's probably beneficial to the historical record if you include the name of the domain you are changing in the squashed commit message and PR title.

PR #677, opened 2024-04-03
grafana alloy profile

rrotter:
- install and configure grafana-agent
- add grafana-agent profile to solr6::catalog profile
- planning to also add this to the solr6::lss profile before squashing and merging

### DIFF
--- a/manifests/log.pp
+++ b/manifests/log.pp
@@ -1,0 +1,30 @@
+# nebula::log
+#
+# Send logs from a service to log server.
+# Currently implimented with Alloy and Loki
+#
+# @example
+#
+#   nebula::log { 'solr':
+#     files => ["/var/log/solr.log"],
+#   }
+#
+#   nebula::log { 'apache':
+#       "files" => ["/var/log/apache.log", "/var/log/apache.err"],
+#   }
+#
+define nebula::log (
+  String $service = $title,
+  Array[String] $files,
+){
+  include nebula::profile::loki
+
+  file { "/etc/alloy/${service}.alloy":
+    owner   => 'root',
+    group   => 'root',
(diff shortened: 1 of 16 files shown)


manifests/profile/hathitrust/solr6/classic.pp
```diff
@@ -0,0 +1,27 @@
...
+#
+# Profile for the classic HathiTrust solr lss/catalog servers
+#
+# @example
+#   include nebula::profile::hathitrust::solr6::classic
+class nebula::profile::hathitrust::solr6::classic (
+  String $loki_endpoint_url = lookup('nebula::profile::hathitrust::loki_endpoint_url'),
+  Boolean $catalog_solr = false,
+){
+  if($catalog_solr) {
```
aelkiss: I'm wondering if rather than having the conditionality here, it might make sense to provide a way to provide additional log files to ship via hiera? That would also make it easier to experiment with shipping additional log files on an individual host in the future without needing to add additional complexity/conditionality/more environments

aelkiss (approved):
This all makes sense to me. The only thing I'm wondering about is the ability to add additional log files for loki to collect via lensoftruth.

manifests/log.pp
```diff
@@ -0,0 +1,30 @@
...
+#     files => ["/var/log/solr.log"],
+#   }
+#
+#   nebula::log { 'apache':
+#       "files" => ["/var/log/apache.log", "/var/log/apache.err"],
+#   }
+#
+define nebula::log (
+  String $service = $title,
+  Array[String] $files,
```
daaang, line 18: Nice!

daaang (approved):

PR #678, opened 2024-04-05
finish transition to supporting new kubernetes apt repos

rrotter:
- parameterize kubernetes apt repo so we can set it in heira
- remove support for old style of kubernetes version strings


estelendur (approved):
shipit (tomorrow)

PR #679, opened 2024-04-15
HathiTrust DEV-1076: support for monitoring for imgsrv

aelkiss:
* Prometheus::Tiny::Shared via CPAN for collecting prometheus metrics in imgsrv
* Script to be run under sudo to allow a process to collect its childrens' final IO metrics

In particular, imgsrv spends much of its time in subprocesses that do a significant amount of I/O. We cannot directly instrument those processes (primarily `unzip` and `grk_decompress`), but would like to monitor some things about them. I have implemented a package in perl (https://github.com/hathitrust/mdp-lib/blob/66a5044a357387289fc60acdf679025b32392553/Utils/MonitorRun.pm#L55) that extends `IPC::Run` to collect metrics from subprocesses after they exit but before issuing a `wait()` system call (that is, the processes are in a "zombie" state). However, after a subprocess exits but before it is reaped, information about its owning user ID is no longer present, and the system makes some of its statistics readable only by root. This sudo script enables a process to read statistics information for its children.

### DIFF
--- a/files/imgsrv/catprocio
+++ b/files/imgsrv/catprocio
@@ -1,0 +1,19 @@
+#!/bin/bash
+
+pid=$1
+
+# ensure given a numeric process ID
+[[ "$pid" =~ ^[0-9]+$ ]] || exit 1
+
+# ensure running via sudo
+[[ "$(readlink /proc/$PPID/exe)" == "/usr/bin/sudo" ]] || exit 2
+
+# the grandparent of this process (i.e. the parent of sudo) must be the same as
+# the parent of the process we're checking
+
+parent=$(cut -f 4 -d ' ' "/proc/$pid/stat")
+this_grandpid=$(cut -f 4 -d ' ' "/proc/$PPID/stat")
+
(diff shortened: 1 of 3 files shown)


aelkiss:
@botimer @rrotter Your reviews here would be helpful to see if:
* there are other options you can think of for collecting the information in `/proc/PID/io` for subprocesses that I haven't thought about
* you can think of any security implications for running `catprocio` via sudo

aelkiss:
Right, there are some facilities for tracing subprocesses, but they seem more complicated to set up and would likely have a performance penalty. Thanks for considering. @rrotter unless you see any concerns here let's go ahead and merge?

aelkiss:
See also: https://github.com/hathitrust/mdp-lib/pull/35, https://github.com/hathitrust/imgsrv/pull/21

aelkiss:
@rrotter Just checking again if you have any concerns here or if we can go ahead and merge?

rrotter (approved):
lgtm, feel free to merge any time

PR #684, opened 2024-05-07
module updates

rrotter:
disable puppetdb management and upgrade modules to max version without updating to stdlib 9.

~This needs to be rebased on top of #682 and #683.~ done


spec/classes/resolv_conf_spec.rb
```diff
@@ -0,0 +1,60 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+describe 'nebula::resolv_conf' do
+  on_supported_os.each do |os, os_facts|
+    context "on #{os}" do
```
daaang: Ah, nice, tests

manifests/profile/haproxy.pp
```diff
@@ -43,7 +43,7 @@
   }
 
   $services.filter |$service, $params| {
-    $params.has_key('floating_ip')
+    'floating_ip' in $params
```
daaang: I forget whether this is something missed from a while ago that was like, no idea how this is still working—or if it's deprecated, and time to update the syntax, but either way, looks familiar. Updated stdlib and updated puppet language.

daaang (approved):
I'm looking at this one before I look at others. I'm assuming that I don't see any `s/action/jump/` or `s/provider/protocol/` because those are covered by the step-by-step upgrades at this point and are no longer relevant to this PR.

manifests/profile/haproxy.pp
rrotter: This was merged in #682, so it'll won't be in this PR once I move it out of draft mode. Anyway, neither syntax makes a ton of sense to me, but this passes the tests w/ either version of stdlib, the old syntax only passes on the outdated stdlib, and I checked just to be sure the code isn't a noop: replacing the line w/ `true` or `false` fails CI.

spec/classes/resolv_conf_spec.rb
rrotter: Yep, just trying to get `firewall` upgraded one step. Intentionally holding back the jump/protocol change for a later pr. I'm not sure which version of `firewall` adds support for the current semantics of `jump`, but I know it's not in `3.6.0`. Just breaking this down into the smallest logical steps.

PR #687, opened 2024-05-28
move all apt repos to keyrings

rrotter:
All non-OS keys belong in `/etc/apt/keyrings` and each repo should be individually pinned to a key. The `puppet/apt` module detects that you are using a keyring based on the use of the `'key'=>'name'` field, which isn't needed or allowed for classic keys since they all get cat'ed into the same file.

`/etc/apt/keyrings` removes the need to convert between binary and "ascii armored" keys: the format is specified by the file extension (.gpg/.asc).

The file names are arbitrary (they only need to be unique, and `puppet/apt` deals with both writing and linking them), but I tried to be somewhat consistent with the naming:
* Use the domain basename if it's clear what they are hosting:
  * adoptium.com => adoptium.gpg
* If the domain isn't com/net/org/edu, include it:
  * elastic.co => elastic.co.asc
* Prepend the name of the source if the domain is opaque (and keep the tld because: what's a "sury"??):
  * php-community + https://packages.sury.org/php/ => php-community-sury.org.gpg
* Binary files are .gpg, ascii files are .asc, this is an OS feature.

Note the `local` repo is managed in lens, not here, the corresponding environment has been updated there and does not depend on these changes.


rrotter:
Also, repo for mono is broken. Fixing it is out of scope here, but I added a note about it in the PR (see: manifests/profile/apt/mono.pp), because atm it's not even clear anyone is aware of this.

rrotter:
Had to add cleanup of sources.list.d for Ubuntu because we ended up w/ multiple conflicting puppet apt configs. We should probably manage all the apt files for Ubuntu like we do for Debian, but that's out of scope.

skorner (approved):
This all makes sense and looks good to me. I appreciate the thought given to the naming conventions and listing them here, as well as noting issues you came across that are out of scope but could use attention at some point.

PR #688, opened 2024-05-28
Add geoip support to the quod general profile

skorner:
This includes perl packages, geoip executables, and a symlink to where the country db is located on shared quod storage.

Roger has confirmed these changes are sufficient for his needs to add country identification for a new collection. We'll need to implement a scheduled refresh process for the database separately. Currently HT owns the service and the refresh after many years of LIT owning it. HT is fine sharing with us for this reviced purpose, as long as we document the arrangement.


rrotter (approved):
lgtm

PR #690, opened 2024-06-04
fix yaz and mono apt repo

rrotter:
added keys, default to a repo dist that actually exists for mono


rrotter:
This might not last long if we have a newer way to support marcedit, but this will also be relevant when we go to upgrade Debian and find that there are other repos that still don't support bookworm.

PR #691, opened 2024-06-04
add Ubuntu 22.04 to supported operating systems

rrotter:
WIP PR so we can get integration running against 22.04.


rrotter:
Merging this. Three's not that slow. We can remove 20.04 before we add another version of Debian or Ubuntu.

PR #692, opened 2024-06-05
update rewrite rule to remove https for http for Deepblue Data in suppor of routing the app on the private network

umjnega:


skorner (approved):
Looks good. Maybe update the title and description details to say that it is a change for Deep Blue Data in support of routing the app reverse proxy on the local private network.

PR #693, opened 2024-06-10
explicitly set HSTS header on 400 errors

rrotter:
If haproxy doesn't understand the apache response it throws it away and sends a 400. Since `set-header` directive only modifies backend responces it does not apply to a 400 that is produced by haproxy itself.

This explicitly sets Strict-Transport-Security on 400 errors on https frontends to fix this issue.

### DIFF
--- a/manifests/profile/haproxy.pp
+++ b/manifests/profile/haproxy.pp
@@ -19,20 +19,19 @@ class nebula::profile::haproxy(
     master => $master
   }
 
-  file { '/etc/haproxy/haproxy.cfg':
-    ensure  => 'present',
-    mode    => '0644',
-    content => template('nebula/profile/haproxy/haproxy.cfg.erb'),
-    require => Package['haproxy'],
-    notify  => Service['haproxy'],
-  }
-
-  file { '/etc/default/haproxy':
-    ensure  => 'present',
-    mode    => '0644',
-    content => template('nebula/profile/haproxy/default.erb'),
-    require => Package['haproxy'],
-    notify  => Service['haproxy'],
+  file {
+    default:
+      ensure  => 'present',
+      mode    => '0644',
+      require => Package['haproxy'],
+      notify  => Service['haproxy'],
+    ;
+    '/etc/haproxy/haproxy.cfg':
+      content => template('nebula/profile/haproxy/haproxy.cfg.erb');
+    '/etc/default/haproxy':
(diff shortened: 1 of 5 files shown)


manifests/profile/haproxy.pp
```diff
@@ -19,20 +19,19 @@
...
-    mode    => '0644',
-    content => template('nebula/profile/haproxy/default.erb'),
-    require => Package['haproxy'],
-    notify  => Service['haproxy'],
+  file {
+    default:
+      ensure  => 'present',
+      mode    => '0644',
+      require => Package['haproxy'],
+      notify  => Service['haproxy'],
```
daaang, line 27: Ok this all looks unchanged, just adding a file

manifests/profile/haproxy.pp
```diff
@@ -19,20 +19,19 @@
...
+    ;
+    '/etc/haproxy/haproxy.cfg':
+      content => template('nebula/profile/haproxy/haproxy.cfg.erb');
+    '/etc/default/haproxy':
+      content => template('nebula/profile/haproxy/default.erb');
+    '/etc/haproxy/errors/hsts400.http':
+      source  => 'puppet:///modules/nebula/haproxy/errors/hsts400.http';
```
daaang, line 34: Was confused for a second about `puppet:///` instead of a template, but I see, it's in `files` instead of `templates`.

spec/classes/profile/haproxy_spec.rb
```diff
@@ -183,6 +183,10 @@
         end
       end
 
+      describe 'haproxy errors' do
+        it { is_expected.to contain_file('/etc/haproxy/errors/hsts400.http').with_source('puppet:///modules/nebula/haproxy/errors/hsts400.http') }
```
daaang, line 187: :+1:

daaang (approved):
I take it the contents of the file in /etc/haproxy/services.d override all the `errorfile` settings in haproxy.cfg? Ah yeah, at a glance, those are all under `defaults` so there you are. Looks sane to me.

manifests/profile/haproxy.pp
rrotter: Yeah, it's full of "^M" line endings, and I'm pretty sure they are important. Not letting erb touch that.
(see: https://github.com/haproxy/haproxy/blob/master/examples/errorfiles/400.http)

PR #695, opened 2024-06-13
ubuntu apt cleanup

rrotter:
* Manage Ubuntu apt repos, use common apt config w/ Debian.
* Purge Ubuntu recommended packages that never should have been installed.
* Install aptitude, because I'm tired of installing it manually on every server.
* rm Ubuntu motd ads.


rrotter:
Just pushed again with some better commit messages. Now I'm done editing.

umjnega (approved):
Looks good to me.

PR #696, opened 2024-07-12
HathiTrust DEV-1058: scrape additional logs for apache

aelkiss:
I didn't see a way to inject additional logs via lensoftruth, so just adding them here.

### DIFF
--- a/manifests/profile/hathitrust/apache.pp
+++ b/manifests/profile/hathitrust/apache.pp
@@ -211,6 +211,13 @@ class nebula::profile::hathitrust::apache (
   }
 
   nebula::log { 'apache':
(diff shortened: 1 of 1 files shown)


botimer:
I think this approach is good -- similar to what we have done with Fulcrum, where more of the details are explicit in the Puppet files. It's not a generic library, so binding them here establishes a stronger and more discoverable convention.

I think it's also fine to have this explicit list, though it may be helpful in some cases to know that there is [globbing support](https://grafana.com/docs/alloy/latest/reference/components/loki/loki.source.file/#file-globbing) with the `local.file_match` component. Using it would require changes to the nebula::log define and template, though.

PR #698, opened 2024-07-17
HathiTrust DEV-1238: redirect for data api retirement

aelkiss:


aelkiss:
Note this will also redirect anything else starting with /cgi/htd, /cgi/htdc, or /cgi/kgs. There are no other valid things currently with those prefixes, and we are not intending to deploy any new CGI apps. That over-broadness to me is worthwhile in preference to potentially missing variations in potential suffixes/ways these might be called.

aelkiss:
@rrotter Could you take a quick look and approve (assuming no issues) but wait for me to merge? I just need to double-check with Ange before we deploy. Thanks!

rrotter (approved):
I guess we either don't use shcgi anymore or it's not relevant for this particular service?

Other than that, LGTM.

aelkiss:
Somehow we do still get hits to `/shcgi/ls` and `/shcgi/pt`, but not to the data api -- and in any case worst case they would get a 404, which I'm OK with (`/shcgi` hasn't really been a thing since we stopped using Cosign for Michigan HathiTrust users)

PR #700, opened 2024-07-19
Add php-bcmath to default production PHP

botimer:
The campus readership map is planned for an update soon, and will depend on php-bcmath to support Google Analytics 4. This follows the addition in development. At present, this materializes as php8.1-bcmath.


skorner (approved):
Looks good for adding the new php-bcmath requirement!

PR #701, opened 2024-07-29
support debian 12

rrotter:
Update puppet modules to add support for Debian 12, add Debian 12 to metadata.json so we run tests on it.


rrotter:
rev'ed puppet/php. Our version didn't support Debian 11, but the current version does.

rrotter:
Bumped a few more puppet modules. I consider this ready to merge.

daaang (approved):

PR #709, opened 2024-08-15
Fix path for startup_imgsrv

aelkiss:
Trivial change to watch the correct path for restarting imgsrv (see https://hathitrust.atlassian.net/browse/DEV-1317, https://github.com/hathitrust/babel/blob/main/bin/rsync.app.timestamp)

I don't love that these things are coupled, and maybe we should be deploying startup_imgsrv as part of babel, but right now this needs to be fixed here.


aelkiss:
FYI @moseshll

aelkiss:
(Long term, imgsrv should run in kubernetes, and then it would be using its own version of startup_imgsrv that's already in the repo -- https://github.com/hathitrust/babel/blob/main/imgsrv/bin/startup_imgsrv)

aelkiss:
@rrotter if you could get this merged when you have a chance I'd appreciate it -- thanks!

rrotter (approved):

PR #712, opened 2024-09-06
HathiTrust DEV-1303: Add USE_CATPROCIO env var

aelkiss:
We want to avoid calling 'sudo catprocio' when it isn't going to work; with a corresponding change in babel, this checks that the USE_CATPROCIO env var is set and only does sudo catprocio if so.


aelkiss:
Also see https://github.com/hathitrust/babel/pull/48

aelkiss:
I also removed some unused cron jobs that are likely silently failing at the moment.

rrotter (approved):
lgtm. Feel free to merge on Monday, or let me know if you'd like me to.

PR #714, opened 2024-09-13
Add a ticket level alert for prep volume disk pressure

skorner:
Our prep filesystems can have large and rapid fluctuations in utilization. They are also often close to full without this being a problem. This makes alerts challenging. However, it seems reasonable to receive a ticket level alert for 98% full on any prep filesystem, regardless of fstype. This alert as written does rely on prep filesystems having the word prep in their mount point, but that is almost always the case.


daaang (approved):
Makes sense

PR #715, opened 2024-09-18
Open /etc/ssh/ssh_config to be able to split into concat fragments

daaang:


botimer (approved):
Does what it says, and for good reason -- ship it!

PR #717, opened 2024-09-26
HathiTrust DEV-1318: limit access to metrics endpoint

aelkiss:
The metrics endpoint doesn't need to be public, and won't get consistent data coming from it anyway. This limits it to a provided set of IP addresses. While we could try to collect IPs for the prometheus role and the kubernetes gateway, it wasn't immediately clear how to collect the floating IP that the kubernetes gateway uses, so this defers to a set of allowed IP addresses listed in lensoftruth.


aelkiss:
Currently attempting to test on preview.babel.hathitrust.org

aelkiss:
@daaang Do you have any better recommended way to collect IP addresses of the Kubernetes gateways?

rrotter (approved):

PR #718, opened 2024-10-02
Filebeat update

antmoth:
Replicates manual changes made on brandy to upgrade to filebeat 7


antmoth:
there ARE still syntax changes needed for the small templates, which have only been moved so far [edited]

skorner (approved):
This PR now covers all the changes needed to update from Filebeat 5.x to 7.x, as well as removing configs for streams we are no longer processing. It's also been verified on two VMs.

I recommend squashing down to two commits, one for the filebeat update and one for the cleanup. 

We'll have to coordinate the merge with an apt update for the packages from elastic.

spec/classes/profile/elastic/filebeat/configs/ulib_spec.rb
```diff
@@ -19,7 +19,7 @@
         it do
           is_expected.to contain_file(file)
             .that_notifies('Service[filebeat]')
-            .with_content(%r{^\s+document_type: ulib$})
+            .with_content(%r{^\s+ulib_type: 90_day$})
```
skorner, line 22: An explanatory note for posterity: naming the type 90_day is a setup to allow for writing logs to different locations on disk based on potentially multiple retention policies for different types of logs or services (e.g. if Fulcrum and HT want longer than default retention. We likely won't be going any further with this plan though, as we are moving towards grafana alloy and loki.

PR #719, opened 2024-10-04
HathiTrust DEV-1350: /var/log/babel for web apps

aelkiss:
Currently, the HathiTrust web applications log to NFS, which is suboptimal for many reasons -- processes have to write to individual files rather than being able to share a log file, and logs can't easily be collected from there and sent to retention or query systems (filebeat, loki).

This:

* creates /var/log/babel and makes it writable by the web server
* adds a logrotate config to ensure logs are rotated (similar to apache logs)
* adds an alloy config to send these logs to loki

I have been testing this configuration on dev-1.babel.hathitrust.org and the alloy config appears to work; the logrotate config may still have some errors (will check next Monday.)

Once the logrotate config is proven to work as expected, this should be safe to merge at any time; once it is merged, we can deploy the corresponding change for the web applications.

### DIFF
--- a/manifests/profile/hathitrust/babel_logs.pp
+++ b/manifests/profile/hathitrust/babel_logs.pp
@@ -1,0 +1,34 @@
+# Copyright (c) 2024 The Regents of the University of Michigan.
+# All Rights Reserved. Licensed according to the terms of the Revised
+# BSD License. See LICENSE.txt for details.
+
+# nebula::profile::hathitrust::babel_logs
+#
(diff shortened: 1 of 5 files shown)


aelkiss:
The filebeat config does not need any changes; `/ulib-logs/archive/macc-ht-web-dev-161.umdl.umich.edu/var/log/babel` looks as expected.

rrotter (commented):
lgtm

rrotter (approved):

aelkiss:
The logrotate config appears to be working. I will go ahead and merge.

PR #720, opened 2024-10-08
add misc interactive packages on bastion hosts

rrotter:
Packages I'd like to have on the bastion hosts. May make sense to add this class to dev hosts as well, but I'm starting with the hosts I login to every day.

Didn't add them to minimum role because we really don't need any of this on most servers (since on the majority of servers no one ever logs in).


botimer (approved):
I like this set -- they're common enough, small enough, and don't get in the way.

daaang (approved):

PR #721, opened 2024-10-08
Fix vim settings (mostly for root)

rrotter:
- don't set bg=dark, vim automatically sets bg based on termcap
- create empty /root/.vimrc
  - w/o .vimrc vim loads defaults from /usr/share/vim/vim82/defaults.vim
  - defaults.vim sets mouse=a, overriding /etc/vim/vimrc

### DIFF
--- a/manifests/profile/vim.pp
+++ b/manifests/profile/vim.pp
@@ -15,4 +15,11 @@ class nebula::profile::vim {
     content => template('nebula/profile/vim/vimrc.vim.erb'),
     require => Package['vim'],
   }
+
+  # Write an empty /root/.vimrc to prevent vim from automatically
+  # loading /usr/share/vim/vim*/defaults.vim
+  file { '/root/.vimrc':
+    ensure => 'file',
+    mode   => '0644',
+  }
 }
--- a/templates/profile/vim/vimrc.vim.erb
+++ b/templates/profile/vim/vimrc.vim.erb
@@ -23,7 +23,7 @@ syntax on
 
 " If using a dark background within the editing area and syntax highlighting
 " turn on this option as well
-set background=dark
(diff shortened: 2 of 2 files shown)


manifests/profile/vim.pp
```diff
@@ -15,4 +15,11 @@
     content => template('nebula/profile/vim/vimrc.vim.erb'),
     require => Package['vim'],
   }
+
+  # Write an empty /root/.vimrc to prevent vim from automatically
+  # loading /usr/share/vim/vim*/defaults.vim
+  file { '/root/.vimrc':
+    ensure => 'file',
+    mode   => '0644',
+  }
```
rrotter, line 24: Hoping this is valid code. I just want to touch the file if it's missing. I don't care what's in the file.

rrotter:
I tested the changes to /etc/vim/vimrc with both a white and black background. In each case vim sets `bg` logically to light/dark.

Discovered what was going on w/ `set mouse=` getting quashed from this old thread:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=864074

manifests/profile/vim.pp
daaang: If it doesn't, you can add `content => ''`

daaang: Alternatively,

```puppet
exec { '/usr/bin/touch /root/.vimrc':
  creates => '/root/.vimrc',
}
```

When there's a `creates` on an `exec`, the `exec` only runs when the file doesn't exist.

rrotter: Tested this on two different hosts, does exactly what I wanted: creates the file, but does not change/delete the content if it exists.

PR #723, opened 2024-10-09
Open SSH to private LAN

daaang:


skorner (approved):
This makes sense and will be valuable sometimes. I've also been surprised/annoyed a few times in the past when I could not ssh via the private network from server to server.

PR #726, opened 2024-10-14
remove i40e-dkms package

rrotter:
This package was added to work around a buggy ethernet driver in 2018:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=892105

This has surely been patched in the last 6 years. Also, the package no longer works w/ current kernel/debian.

Reverts #37, #38
See also: https://mlit.atlassian.net/browse/AEIM-1502


skorner (approved):
Approved from my perspective given:

- this package no longer works with newer kernels (and is blocking puppet from applying the catalog on hosts calling for it where it can't be run)
- confirmation (by Ryan in slack) that the package is not currently active on any of the hosts it was intended for and we have not had network problems on those hosts (i.e. we won't be making an updated package)

daaang (approved):

aelkiss:
Was there anything you wanted from me on this PR? I do remember the i40e-dkms driver, but I'd defer to @rrotter and @skorner's assessment of the current state.

aelkiss (commented):
I do remember the i40e-dkms driver, but I'd defer to @rrotter and @skorner's assessment of the current state, and don't have anything to add -- not sure if I was tagged for review on purpose or not. Let me know if there's anything specific you'd like me to check.

skorner:
I think Ryan tagged you due to the historical connection in case you wanted to weigh in. I think Ryan will merge this today. Thanks!

PR #729, opened 2024-10-17
Run filebeat on new bastion hosts

daaang:
Alternative approaches which could better reflect what we want:

1.  We could base the bastion role on umich/hathitrust instead of minimum.
2.  We could add filebeat to the minimum role.


skorner (approved):
This covers the essential need. Of your two optional approaches, including filebeat in the minimal role feels more appealing than tying the bastion role to one or both of umich/hathitrust roles, unless that opens the door for some other umich and hathitrust specific benefits to bastion host setup.

daaang:
Not doing this, gonna add it to the minimum role instead.

PR #731, opened 2024-10-22
add nginx and open firewall port

umjnega:


manifests/profile/prometheus.pp
```diff
@@ -146,6 +146,28 @@
       port  => 443,
       block => 'umich::networks::all_trusted_machines',
     }
+  } else {
+    include nginx
```
daaang:
```diff
-    include nginx
+    class { 'nginx':
+      server_tokens: 'off',
+    }
```

to make it so the server doesn't display its version in 400 responses, in case we do ever open it to the world.

umjnega:
I'm not finding the failure in the details here... help...

PR #734, opened 2024-10-29
update rewrite to point at puma path

umjnega:
I updated the rewrite rule path...   I think its correct based on whats in that path on julep-1.   Thats where puma is running, from my understanding.  Is that correct.  @skorner

### DIFF
--- a/manifests/profile/www_lib/vhosts/mgetit.pp
+++ b/manifests/profile/www_lib/vhosts/mgetit.pp
@@ -54,11 +54,6 @@ class nebula::profile::www_lib::vhosts::mgetit (
 
     rewrites                    => [
       {
-        comment      => 'Serve static assets through apache',
-        rewrite_cond => ['/www/mgetit/current/public/$1/index.html -f [OR]', '/www/mgetit/current/public/$1 -f'],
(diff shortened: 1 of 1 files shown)


skorner (changes requested):
Your change would support apache serving static assets from a path local to the web server, but the asstes are not being deployed there. In order to allow the app's puma to serve the static assets, you'll want to remove [the conditional rewrite](https://github.com/mlibrary/nebula/blob/9823beaf4d7fbb17ba57bc688a9fe4c86141bb3b/manifests/profile/www_lib/vhosts/mgetit.pp#L56-L60) entirely. That will allow the subsequent rewrite to proxy everything to puma.

Once that change is in place, please apply it to macc-www-lib-testing in a puppet env and ask Albert to verify the behavior with a local hosts alias of 141.213.128.250 for mgetit.lib.umich.edu. Thanks!

rrotter:
Did you keep the wrong commit message in the squash? Looks like this is something more like "mgetit: remove static assets rewrite rule".

skorner (approved):

skorner:
+1 to changing the commit message to reflect that the change is about removing the static assets rewrite rule for mgetit rather than an update to it. Thanks!

PR #736, opened 2024-11-04
update puppet/letsencrypt module

rrotter:
We're only using letsencrypt directly in manifests/cert.pp. There, `$title` is automatically used as the first domain. IOW, the "breaking" change to [puppet/letsencrypt](https://github.com/mlibrary/nebula/blob/production/manifests/cert.pp) only enforces a choice that @daaang made when initially writing our use of in nebula.


umjnega (approved):
I think that comment was added under @rrotter's direction the last time we went over nebula dependancies.   At that time It was believed that the breaking change of:
"certonly: Use the first domain for $cert_name instead of the $title[#220]".  Would require changes to be made to lensoftruth.  I do not recall why that was or how the code was written at the time, that changes to lens would have been necessary.  But it appears upon further investigation that is not the case anymore.   I'm all for moving foward with 11 the newer version.

PR #739, opened 2024-11-04
upgrade firewall module to 8.0.0

rrotter:
To support firewall >= 7.0.0
s/jump/action/, s/provider/protocol/
see https://github.com/puppetlabs/puppetlabs-firewall?tab=readme-ov-file#migration-path-to-v700


rrotter:
This branch is diverged too far, starting over w/ a new PR.

PR #743, opened 2024-11-06
upgrade to puppetlabs/firewall 8.1.1

rrotter:


daaang (approved):
I double checked, and I'm seeing `-action` and `-provider` on the left and `+jump` and `+protocol` on the right. I cannot believe they went with `protocol` and `proto` at the same time, where `proto` is the one that means protocol. Just... embarrassing.

PR #747, opened 2024-11-13
commit additional apt keys

rrotter:
To avoid log spam and breakage caused by downloading these files from web servers we don't control, add every key that has either updated needlessly (in some cases on EVERY puppet run!) or not been updated when it should have been. 

Both issues are suspected to be the result of incorrect HTTP headers on the servers we download these keys from.


files/apt/keyrings/README
```diff
@@ -0,0 +1,5 @@
+key_name        source_url
+adoptium.asc    https://packages.adoptium.net/artifactory/api/gpg/key/public
+docker.asc      https://download.docker.com/linux/debian/gpg
+k8s.io.asc      https://pkgs.k8s.io/core:/stable:/v1.31/deb/Release.key
+puppetlabs.asc  https://apt.puppetlabs.com/pubkey.gpg
```
daaang, line 5: Nice documentation of where these are sourced—we definitely won't regret having this.

daaang (approved):
I guess my only question is whether changes to nebula will be picked up. We know that different https servers outside our control behave differently—or in other words, it is conceivable that changes here will be ignored in cases where the file is already present on a target machine.

But it's fine to wait until it matters to find out for certain, and it doesn't matter today.

PR #748, opened 2024-11-13
Remove password keeper ("pk") support

skorner:
The application has been sunset and all useful content migrated to Alma Digital.


daaang (approved):
_sigh of relief_

PR #749, opened 2024-11-14
disable HashKnownHosts

rrotter:
Unclear why we ever set this. Default is `no`. Provides no security benefit and makes removing outdated fingerprints a pain.

### DIFF
--- a/templates/profile/networking/ssh_config.erb
+++ b/templates/profile/networking/ssh_config.erb
@@ -1,7 +1,6 @@
 # Managed by puppet (nebula/profile/base/ssh_config.erb)
 Host *
     SendEnv LANG LC_*
-    HashKnownHosts yes
     GSSAPIAuthentication yes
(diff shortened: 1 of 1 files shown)


skorner (approved):
The default in Debian 9 was "yes" and is what likely led to it being copied into our managed config. I don't believe we ever made an intentional choice to enable it. I agree that it adds no real protective value.

daaang (approved):
It looks like the entire "pros" list for HashKnownHosts is that it obfuscates the list of servers you connect to, in the case of an attacker who secures access to your workstation and is attempting to access servers you connect to (<https://serverfault.com/questions/233855/why-should-i-use-hashknownhosts-yes-in-ssh-config>).

We secure our servers in multiple ways to prevent ssh from bad actors, so I'm not worried about this attack vector from any of our workstations. Server to server, we actually publish `/etc/ssh/ssh_known_hosts`, so this wolf is very much already in the barn.

PR #753, opened 2024-12-02
node 18

rrotter:
On Debian 12 node 18 is the default and node 12 is the oldest version supported by the nodesource apt repo.

We should set node version in the config repo if any version older than 18 is needed.

It's worth noting that currently setting a version here lower than the debian default doesn't actually work (i.e. even though we were setting 10 here, all debian 11 hosts have node 12 or newer)

### DIFF
--- a/manifests/profile/nodejs.pp
+++ b/manifests/profile/nodejs.pp
@@ -17,7 +17,7 @@
 #     version => '14',
 #   }
 class nebula::profile::nodejs (
-  $version = '10',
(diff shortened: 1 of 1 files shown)


rrotter:
Before merging, confirm where we actually need node 10, 12, 14 or 16 and set those in the config repo.

daaang (approved):
Good news! Node 18 has security support for 4 more months! https://endoflife.date/nodejs

PR #757, opened 2024-12-03
update nodejs to 22 for ht test vm

rrotter:
Node 18 is supported through April. Would like to bump this version before then if possible. If this is complicated I'm happy to wait or add support for upgrading one host at a time.


aelkiss:
@carylwyatt tried node 22 and this all worked as expected, so we should update to node 22 here and push it out.

PR #758, opened 2024-12-04
rm legacy facts

rrotter:
- remove legacy facts to support puppet 8
- fix tests that now need the `networking` fact mocked correctly

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -82,3 +82,23 @@ end
 def allow_call(dbl)
   allow(dbl).to receive(:call)
 end
+
+class Hash
+  # deep_merge copied with minor adaptation from:
+  # https://github.com/rails/rails/blob/main/activesupport/lib/active_support/deep_mergeable.rb
+  def deep_merge(other, &block)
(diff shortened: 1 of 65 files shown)


rrotter:
The facts that have their new home under `$facts['networking']` (like `$facts['networking']['ip']`) are all broken by our current tests.

rrotter:
Reviews should probably focus on `/spec`, as this is where nearly all the complexity of this PR lies.

There are more changes under `/manifests`, but they are a simple search and replace.

spec/spec_helper.rb
```diff
@@ -82,3 +82,23 @@ def stub(name)
 def allow_call(dbl)
   allow(dbl).to receive(:call)
 end
+
+class Hash
+  # deep_merge copied with minor adaptation from:
+  # https://github.com/rails/rails/blob/main/activesupport/lib/active_support/deep_mergeable.rb
+  def deep_merge(other, &block)
```
daaang, line 89: Nice

daaang (approved):

PR #761, opened 2024-12-05
Add filebeat to absolutely everything

daaang:


skorner (approved):
Thumbs-up to the expansion of our default log aggregation config to everything.

PR #763, opened 2024-12-05
Revert "Revert "Add logrotate for /var/log/babel on ingest machines""

rrotter:
Reverts mlibrary/nebula#760

Puppet postgres module isn't working, and I have roughly zero interest in figuring out why.


rrotter:
just ignore this mess. I was in the wrong window.

PR #767, opened 2024-12-06
update legacy facts under /templates

rrotter:
See:
* https://www.puppet.com/docs/puppet/8/upgrading-from-puppet7-to-puppet8.html#legacy-facts-deprecation
* #758

### DIFF
--- a/templates/profile/kubernetes/dns/hosts_04_ipv4_hostname.erb
+++ b/templates/profile/kubernetes/dns/hosts_04_ipv4_hostname.erb
@@ -1,1 +1,1 @@
-<%= @facts['networking']['ip'] %> <%= @facts['networking']['fqdn'] %> <%= @facts['networking']['hostname'] %>
+<%= @networking['ip'] %> <%= @networking['fqdn'] %> <%= @networking['hostname'] %>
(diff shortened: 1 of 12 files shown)


templates/profile/kubernetes/dns/hosts_04_ipv4_hostname.erb
rrotter, line 1: This is one that I fixed in the previous pass. Using `@networking` for consistancy, `@networking`  and `@facts['networking']` should be equivalent.

daaang (approved):
This all looks sane. Are you certain that `@networking` is accessible from all these templates? I don't have any certainty around whether each corresponding puppet code will need

```puppet
$networking = $::networking
```

This is probably at least *somewhat* covered by the tests...

rrotter:
> This is probably at least _somewhat_ covered by the tests...

In puppet code `$::facts['networking'] == $::networking` (where I've used the latter universally). I think this maps to erb templates such that `@facts['networking'] == @networking`. If this is wrong I'd just use `@facts['networking']` in the templates. I'd expect to see a lot of failing tests in that case, but I'm also going to test this out in an environment with a few hosts before merging.

rrotter:
Converted to draft as this needs to be tested further. Not to be merged before Monday.

rrotter:
There were a few places where `$ipaddress`, _et al._ were local variables so `@ipaddress` in the template was actually correct as we need the local variable and not the fact. I've reverted the cases of this issue that I found.

PR #768, opened 2024-12-09
Support rolling deployment of Cloudflare WAF

botimer:
By separating the connecting IP handling from the access restriction, we allow a period where direct and proxied clients can access a service. After we are satisfied that all DNS propagation is complete, we can then restrict to proxied clients only.

The strategy here involves unconditionally checking the source IP and whether the CF-Connecting-IP header is supplied. We unconditionally set the X-Client-IP to the source. Then, we conditionally overwrite it with the supplied header value, if and only if its is supplied and the source is a Cloudflare proxy server. The access restriction is only enforced when the service's cloudflare_protected property is true.

### DIFF
--- a/spec/defines/haproxy_service_spec.rb
+++ b/spec/defines/haproxy_service_spec.rb
@@ -74,9 +74,12 @@ describe 'nebula::haproxy::service' do
               errorfile 400 /etc/haproxy/errors/hsts400.http
               default_backend svc1-dc1-https-back
               acl blocked-ip src -f /etc/haproxy/global_badrobots.txt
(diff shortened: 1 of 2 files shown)


rrotter (approved):

daaang (approved):
I had questions but the commit message answered them. :+1:

botimer:
> I had questions but the commit message answered them. 👍

I suppose I could have made it slightly more clear by saying "unconditionally set the X-Client-IP to the source (to handle all directly-accessed services)". Thanks for reviewing!

PR #769, opened 2024-12-10
update unattended_upgrades

rrotter:
- disable AC power check: useless, only generates log spam
- add Puppetlabs source

Does not add `profile::unattended_upgrades` to any roles that don't already have it. Anything roles that should obviously get it? This can also be an additional PR.

edit: Also suppress metered connection check. See `/var/log/unattended-upgrades/unattended-upgrades.log` if you're interested in the output we're eliminating here.


umjnega (approved):
Makes sense.   Nothing we run unattended upgrades on would be battery powered.

PR #770, opened 2024-12-10
fully remove puppetdb management from nebula

rrotter:
Previously puppetdb was commented out, and tests were disabled with xit.


manifests/profile/puppet/db.pp
```diff
@@ -7,8 +7,5 @@
 # @example
 #   include nebula::profile::puppet::db
 class nebula::profile::puppet::db {
-  #class { 'puppetdb':
-  #  disable_cleartext => true,
-  #  manage_firewall   => false,
-  #}
+  # puppetdb is no longer managed by nebula
```
daaang, line 10: :+1: good comment

daaang (approved):
This is all good, but I would also support deleting the `master_with_db` profile (it's only mentioned in `role::puppet::master`) and both spec files.

PR #771, opened 2024-12-11
upgrade pdk [spike]

rrotter:
PDK expects that you won't modify certain files yourself. This PR relocates our CI customizations so this and future PDK upgrades will require a minimum of manual remediation after running `pdk update`.

Add `pdk` gem to Gemfile so we can easily run `bundle exec pdk`.


rrotter:
closing this as a spike and starting with a fresh branch

PR #772, opened 2024-12-11
Handle Cloudflare worker request info

botimer:
Our strategy here is to identify proxied requests that also carry the CF-Worker header. These requests include a special source address in IPv6 form, so we approximate it as an IPv4 address in the reserved "Class E" block, similar to their psuedo-IPv4 approach. We also augment the User-Agent to identify the specific worker. A separate header may ultimately be better, but this approach is transparently logged, for some convenience.

The magic address Cloudflare uses is 2a06:98c0:3600::103, and we map it statically to 240.36.0.103 to have vague similarity.


skorner (approved):
This is a very reasonable way for us to manage these requests while accommodating Cloudflare proxying in the most general way and without imposing new requirements on our tooling that assumes IPv4 only. I also appreciate the detailed problem statement and solution rationale in the comment.

PR #775, opened 2024-12-16
move lint to separate github workflow

rrotter:


daaang (commented):
Hm I'm gonna offer an alternative that's a little more GHA magic, give me a minute...

daaang:
See #776

PR #776, opened 2024-12-17
GitHub Actions - run linting and specs separately

daaang:


.github/workflows/ci.yml
```diff
@@ -8,10 +8,15 @@ on:
 jobs:
   test:
     runs-on: ${{ contains( github.event.pull_request.labels.*.name, 'use_fast_ci') && 'ubuntu-latest-8core' || 'ubuntu-latest' }}
+    strategy:
+      fail-fast: false
```
daaang, line 12: This is `true` by default and would e.g. stop running tests in the event that linting fails. Setting it to `false` will make tests run to completion, even if linting fails.

.github/workflows/ci.yml
```diff
@@ -8,10 +8,15 @@ on:
...
+      matrix:
+        target:
```
daaang, line 14: The word `target` is arbitrary here, just something to reference later as `${{ matrix.target }}`

.github/workflows/ci.yml
```diff
@@ -8,10 +8,15 @@ on:
 jobs:
   test:
     runs-on: ${{ contains( github.event.pull_request.labels.*.name, 'use_fast_ci') && 'ubuntu-latest-8core' || 'ubuntu-latest' }}
```
rrotter: We never want to run lint on ubuntu-latest-8core. We've not used this in a while, and I almost took it out in my PR. We should either remove this now, or modify it so it only applies to "specs" and not "lint".

rrotter (commented):
see comment on runner types

.github/workflows/ci.yml
rrotter:
```suggestion
    runs-on: ubuntu-latest
```

PR #778, opened 2024-12-17
upgrade to pdk 3, ruby 3

rrotter:
Upgrade to:
- ruby 3.2: used in current puppet 8
- pdk 3.3.0 (latest that doesn't cause gem conflicts)
- pdk template 3.0.1 (later versions change `rake spec` tasks significantly introducing a ton of warnings)
- ~add standardrb~ this will be another PR


rrotter:
Removed the last commit in this PR (which added standardrb), it doesn't belong here.
[edit: standardrb moved to #779]

daaang (approved):

PR #781, opened 2024-12-24
Restart every VM after a crash

daaang:
The bit of code that configures VMs to autostart was inside the "make sure this VM isn't already installed" condition, which was either a mistake or a shortest path to get the tests to pass. I've fixed the tests, so this commit will also mark every VM it manages as autostart.

Otherwise, this will only set a VM to restart after a crash if it is currently set to destroy after a crash.

All the options for what to do after a crash are:

1.  destroy
2.  restart
3.  rename-restart
4.  preserve
5.  coredump-destroy (only available on crash, not e.g. poweroff)
6.  coredump-restart (only available on crash, not e.g. poweroff)

I haven't looked into what any of these mean, but coredump is an intriguing word. This commit simply changes anything that's option 1 to option 2.

### DIFF
--- a/manifests/virtual_machine.pp
+++ b/manifests/virtual_machine.pp
@@ -93,7 +93,16 @@ define nebula::virtual_machine (
     ensure => 'directory',
   }
 
-  unless $facts['vm_guests'].member($title) {
+  if $facts['vm_guests'].member($title) {
+    # If the VM already exists, ensure that it's marked to autostart
+    # when the host boots. (This exec is in this conditional because it
+    # isn't constructed to be safe to run before the VM exists.)
+    exec { "${prefix}::autostart":
+      creates => "${autostart_path}/${title}.xml",
+      command => "/usr/bin/virsh autostart ${title}",
+    }
+  } else {
+    # If the VM does not already exist, create it.
     if $build == 'bullseye' {
       file { "${tmpdir}/preseed.cfg":
         content => template("nebula/virtual_machine/${build}.cfg.erb"),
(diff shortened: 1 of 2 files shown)


manifests/virtual_machine.pp
```diff
@@ -93,7 +93,12 @@
     ensure => 'directory',
   }
 
-  unless $facts['vm_guests'].member($title) {
+  if $facts['vm_guests'].member($title) {
+    exec { "${prefix}::autostart":
+      creates => "${autostart_path}/${title}.xml",
+      command => "/usr/bin/virsh autostart ${title}",
+    }
+  } else {
```
daaang, line 104: In case this is tricky to parse, here's my attempt at translating to English. It was:

1. If the VM doesn't exist, put an `exec` that creates the VM in the host catalog.
2. If the VM doesn't exist, put an `exec` in the catalog that ensures the VM is autostart, but `require` the install `exec`.

This was a problem because often, creating a VM times out and succeeds outside of the puppet run. So the autostart `exec` never ran on the initial run, and on subsequent runs it was in the wrong condition branch, since the VM exists now.

Now it's

1. If the VM doesn't exist, put an `exec` that creates the VM in the host catalog.
2. If the VM *does* exist, put an `exec` in the catalog that ensures the VM is autostart (with no `require`).

rrotter: And we can't run this outside of the condition statement (rather than moving from the `unless` to `if`) because we'd need a `require` in one case and not the other?

This is a bit odd, but I think it's fine. I'd suggest adding a comment like:
```
# ensure autostart settings on vm, after creation on a previous puppet run
if
...
# create vm, only if it doesn't exist
else
```

manifests/virtual_machine.pp
```diff
@@ -132,11 +137,14 @@
...
-      command => "/usr/bin/virsh autostart ${title}",
-    }
+  exec { "/usr/bin/virsh set-lifecycle-action ${title} crash restart --config":
+    onlyif => "/usr/bin/virsh dumpxml --inactive ${title} | grep on_crash | grep -q destroy",
+  }
+
+  exec { "/usr/bin/virsh set-lifecycle-action ${title} crash restart --live":
+    onlyif  => "/usr/bin/virsh dumpxml ${title} | grep on_crash | grep -q destroy",
+    require => Exec["/usr/bin/virsh set-lifecycle-action ${title} crash restart --config"],
   }
```
rrotter: 1. Not sure what exactly this does, but from the `--inactive` in one and not the other it looks like these are the both do the same thing but one is for stopped guests and the other is for running guests?
2. Shouldn't these be in the `if` above, so they don't run before or during vm creation?
3. What's the purpose of the require? Unless I was mistaken in (1) above, the order of these can't matter because they are mutually exclusive.

rrotter (changes requested):

manifests/virtual_machine.pp
daaang: 1. I'll add a comment that hopefully clarifies, but the first one is "if the config xml file says `<on_crash>destroy</on_crash>`, change the file to `restart`." The second one behaves differently depending on whether the VM is running. If it's not running, `dumpxml` and `dumpxml --inactive` do the same thing, so the second exec won't run. If the VM is running, this checks the live xml and changes that if necessary.
2. In this case it doesn't matter, because the `dumpxml` command will fail if the VM doesn't exist yet, preventing the `exec` from running. The autostart bit is different because it runs when a symlink hasn't been created, which will lead to it executing before the VM is created if it's neither requiring the vm creation nor in a different condition branch.
3. The require is probably not required, but if it weren't there, it could result in `set-lifecycle-action --live` getting run on a VM that isn't running, and I'm not sure what that would do. Could result in a failing puppet resource alert? Didn't want to find out.

manifests/virtual_machine.pp
daaang: If this runs before the VM is created, I'm not sure what its behavior would look like. I doubt it would cause lasting damage, but it also wouldn't do anything valid.

rrotter (approved):

PR #782, opened 2024-12-24
add vimrc.local

rrotter:
- revert customized /etc/vim/vimrc to os defaults
- move our customizations to `/etc/vim/vimrc.local` which won't conflict with packaged config

Default configs sources:
* http://ftp.us.debian.org/debian/pool/main/v/vim/vim-common_8.2.2434-3+deb11u1_all.deb
* http://ftp.us.debian.org/debian/pool/main/v/vim/vim-common_9.0.1378-2_all.deb
* http://security.ubuntu.com/ubuntu/pool/main/v/vim/vim-common_8.2.3995-1ubuntu2.21_all.deb


daaang (approved):
Am I right in thinking that this is part of a two-step process?

1. Merge this to get system vimrc to match what apt would put there.
2. Stop managing those files with puppet; let apt take over.

And so this is step 1?

rrotter:
Yes, that's why I added the comment concerning removing /etc/vim/vimrc later. I thought about just using scp for reverting the changes and, ugh…

PR #785, opened 2025-01-03
root dotfiles

rrotter:


files/root/.bashrc
```diff
@@ -0,0 +1,29 @@
...
+[ -z "$PS1" ] && return
+
+# color, abbreviation
+COLOR_OPTIONS='--color=auto'
+eval "$(dircolors)"
+alias l='ls'
+alias ls='ls $COLOR_OPTIONS'
+alias ll='ls $COLOR_OPTIONS -l'
+alias la='ls $COLOR_OPTIONS -A'
+alias lla='ls $COLOR_OPTIONS -lA'
```
daaang: I know you said this was largely copied from debian/ubuntu defaults, so I'm not inclined to do much bikeshedding, but since these are meant to be convenient, I'll suggest `s/-l/-lh/`, since `-h` is almost always preferable for taking information in at a glance.

botimer:
I do not care about a host of variant aliases. I care about colors by default, since that has been in place for login shells across distros for so long as to be de facto ls behavior.

I don't expect I will collide with these relatively common others; I would only prefer that we don't add or invent a bunch. The flags/combinations are already available (in any order) with one dash, so I find the aliases dubious.

rrotter:
- Removed `lla`, that was the only item in here that wasn't lifted from the actual default `/etc/skel/.bashrc` or `/root/.bashrc` on Debian or Ubuntu.
- Added `-h` to  `ll`. I agree with @daaang. There's another -F in there for similar (imo) ergonomic reasons, but I won't go too far defending either.
- Removed most of my editorial.

I will not edit this branch further. If there objections to what's here please feel free to add commits to this branch.

If there are no objections, then this is just waiting on confirming on Monday whether we need the odd local edits currently present on one or two servers that this commit will run over.

daaang (approved):
I guess I've got used to the lack of colors, but their addition won't hurt me either, plus I'll know at a glance which things are directories. Seems good.

I have been stubbing my toe on the lack of `ll` probably about once weekly since 2017 (never complained since I figured it was my fault), so that alias will be a welcome addition for me. I doubt I'll use any of the others, but they won't get in my way or harm me, and I think I've seen them all before.

`recurse => 'remote'` and `purge => false` sound right to me for this.

rrotter:
imo the most important things here were:
1. Eliminating random variation, while making people more comfortable.
2. Longer history so ^R works reliably at the cost of like 4k of disk space.

rrotter:
Thanks for the feedback, everyone. Merging.

PR #788, opened 2025-01-08
allow linux kernel args to be set in hieradata

rrotter:
These args are wrong for many hosts. We should be generating the correct ones once we have logic to accurately generate them. For now, this makes it possible to fix one-host-at-a-time via host data.


skorner (approved):
This is a really useful and practical step forward. Thanks!

PR #789, opened 2025-01-16
don't require two-factor for sudo

rrotter:
- remove check for historical duo-unix package
- don't modify /etc/pam.d/sudo
- revert /etc/pam.d/sudo to default state on each supported distro
  - this should be removed after all hosts are reverted

Because:
- we confirmed this isn't needed
- reverting to the os package supplied config files makes upgrades safer and easier


skorner (approved):
Thumbs-up to removing this redundant two-factor requirement.

PR #792, opened 2025-01-21
use drop-in config to manage ssh*_config

rrotter:
ssh_config and sshd_config are modified in place, which causes conflicts with dpkg. This PR will revert those changes and move them to the respective `*.d` drop-in config dirs.


rrotter:
This is too broad. Restarting this from a new pr targeting just sshd_config.

rrotter:
superceded by #879

PR #795, opened 2025-02-04
HathiTrust TTO-288: send shib headers to Dex even when blank

aelkiss:
Currently, HathiTrust uses a forked version of Dex (https://github.com/hathitrust/dex-shib-proxy) to present itself as an OIDC identity provider, serving as a proxy to the variety of SAML identity providers that HathiTrust supports. This forked version supports using additional headers to populate various claims in Dex.

As of Dex 2.38, Dex supports additional claims via the auth proxy. However, if one of the headers it is using is missing, Dex will report an error. This changes the behavior to always send the headers, even when empty.

Additionally, this version of Dex expects the delimiter for the 'groups' claim to be "," rather than ";".


aelkiss:
@rrotter I would like to be ready to merge this by the afternoon of Thursday, February 6th, assuming there aren't any concerns. I'd like to be the one to actually press merge, because there will be a corresponding configuration update for the HathiTrust Dex that needs to be aligned.

rrotter (approved):
Feel free to merge on your schedule.

PR #797, opened 2025-02-19
move grub config to drop-in

rrotter:
- move grub under `/etc/default/grub.d/`
- revert all line-in-file changes

### DIFF
--- a/manifests/profile/grub.pp
+++ b/manifests/profile/grub.pp
@@ -6,69 +6,59 @@
 #
 # Manage grub.
 #
-# @param kernel_args Optionally specify kernel args for physical host. Ignored for virtual machines.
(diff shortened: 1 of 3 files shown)


rrotter:
Last push was a `rebase -i` to reorder and squash commits to make the history less confusing. I should be done now.

rrotter:
Manually tested this on a couple of diverse servers. Critically, it causes `/boot/grub/grub.cfg` to get rewritten, but with the exact same content. That's all we wanted.

(`/etc/default/grub*` are bash scripts that affect how `/boot/grub/grub.cfg` when it's generated, which happens on every new kernel, as well as when `update-grub` is run manually)

PR #799, opened 2025-02-20
release ht solr catalog at 6, not 6:30

rrotter:
Solr release may cause a breif period where servers are unavailable to users. There seems to be no good reason to have this happen twice per day, so let's have catalog release at the same time as lss.


aelkiss:
Looks fine to me. If there's somehow an issue with this tomorrow morning, we can revert.

aelkiss (approved):

PR #800, opened 2025-02-20
switch from ntpd to chrony

rrotter:
- purge ntp, supporting packages
- install and configure chrony
- add nebula::profile::ntp::pools key to differentiate pools and servers
  - this will require a corresponding update to heira data
- remove ntp module
- intentionally not using chrony module:
  - our usage is too simple to justify a new dependency
  - ~because module set iburst by default, this update would actually be more complex with the module, not less~
- add ntp profile to mimimum role


skorner (approved):
Thumbs-up to the policy change of managing time everywhere and the change in approach of how we do it.

rrotter:
Added iburst. It's somewhat rude on public NTP servers, but everyone (i.e. Debian and Ubuntu default configs) does it anyway. And, we're not using public NTP servers.

PR #801, opened 2025-02-21
grub: don't manage /etc/default/grub

rrotter:
Management no longer needed, since #797 reverted this file to distro defaults.


rrotter:
Don't merge until 24 Feb.

PR #803, opened 2025-02-27
resolv_conf: disable systemd-resolved

rrotter:
On Ubuntu 22.04 jammy, the `systemd-resolved` service is running, and it shouldn't be.

This is part of the resolv_conf class because `systemd-resolved` presumes to overwrite /etc/resolv.conf.

I have confirmed:
* `systemd-resolved` service exists but is stopped+disabled on all bullseye (Debian 11) hosts
* `systemd-resolved` package is purged on all bookworm (Debian 12) hosts

…so on Debian this will be a no-op, just confirming the status quo.


umjnega (approved):
Looks good to me.

PR #804, opened 2025-02-27
grub: remove ixgbe flags

rrotter:
We're setting this in hiera for each of the (few) hosts that still use the ixgbe kernel module.


daaang (approved):
My read on this is that either this is being overridden for every host in lensoftruth, or there are a couple vmhosts that will be unable to come back after a reboot. I haven't verified that it's the former, so I'll trust you to say it's not the latter.

rrotter:
It's overridden in LoT for each host that uses the ixgbe kernel module.

PR #805, opened 2025-02-27
Update grub default serial config to ttyS0

rrotter:
console=ttyS0 seems to work on the most hosts, and critically, all of our most recently purchased hosts.


daaang (commented):
Seems like this conflicts or does the same thing as the other grub PR? I'm confused.

rrotter:
idk if there's a better way to do this in GitHub: Commit 1 is the same and shouldn't be in this PR. I think it will disappear from the diff when I merge that PR. Commit 2 is all that should be on this PR.

PR #806, opened 2025-02-28
apt: don't anger dpkg when removing 20apt-esm-hook

rrotter:
Instead of `rm`, use `dpkg-divert` to relocate this file so dpkg doesn't think it's missing.

### DIFF
--- a/manifests/profile/apt.pp
+++ b/manifests/profile/apt.pp
@@ -157,8 +157,17 @@ class nebula::profile::apt (
       'security' : release => "${facts['os']['distro']['codename']}-security";
     }
 
+    # remove unwanted recommendeds from `ubuntu-server` package
     package { 'landscape-common': ensure => purged }
     package { 'open-vm-tools': ensure => purged }
-    file { '/etc/apt/apt.conf.d/20apt-esm-hook.conf': ensure => absent }
+
+    # disable ubuntu subscription advertisements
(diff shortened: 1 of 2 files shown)


manifests/profile/apt.pp
```diff
@@ -157,8 +157,17 @@
...
     package { 'open-vm-tools': ensure => purged }
-    file { '/etc/apt/apt.conf.d/20apt-esm-hook.conf': ensure => absent }
+
+    # disable ubuntu subscription advertisements
+    exec { 'disable 20apt-esm-hook.conf':
+      creates => '/etc/apt/apt.conf.d/20apt-esm-hook.conf.disabled',
+      timeout => 30,
+      command => "/usr/bin/dpkg-divert --rename \
+--divert /etc/apt/apt.conf.d/20apt-esm-hook.conf.disabled \
+--add /etc/apt/apt.conf.d/20apt-esm-hook.conf"
```
daaang, line 170: It looks to me like this could be accomplished with `rm 20apt-esm-hook.conf; touch 20apt-esm-hook.conf.disabled` but that this is the "right" way to do it.

daaang (approved):

rrotter:
This should make `dpkg` believe the file still exists, so it doesn't complain (via a forced interactive prompt!) when we run `apt-get upgrade`.

@umjnega, by all means if this doesn't fix it, please let me know and I will take another look.

PR #807, opened 2025-02-28
create apt::nodejs w/ updated nodesource repo

rrotter:
- commit nodesource gpg keys
- use our copy of gpg key for both new and existing profile
- new profile for configuring nodejs repo
  - refuse to configure nodesource repo unless it's a newer version than the os ships
  - this change is not compatible w/ how the existing profile works, thus I envision existing roles transitioning to this profile one at a time
- updates nodejs for hathitrust only, but other roles can use this profile too


antmoth (changes requested):
the key files in the commit are named
```
nodesource.asc
nodesource_legacy.asc
```
but the filenames in the code are named
```
nodesource.asc
nodesource0.asc
```
which seems incorrect

antmoth (approved):
looks good to me

rrotter:
Rebased. Deleted several useless lines that were already commented out anyway.

PR #809, opened 2025-03-10
fix puppet lint

rrotter:
`s/ +=>/ =>/`

Not clear when this changed, but it's breaking CI. This should fix all the CI errors that are holding up #807.

Bizzarely, `bundle exec rake lint` is fine with both the before and after version on my local machine.


antmoth (approved):
??? ok sure. lint why?

PR #811, opened 2025-03-26
entire messy fulcrum-demo branch squashed for easy merging

antmoth:
there may still be things that I've changed that will affect other machines - in particular want to check that the fulcrum www_and_app_host role isn't negatively affected


antmoth:
confirmed no changes on fulcrum-staging-136 or macc-www-lib-testing yay

PR #814, opened 2025-03-31
remove puppetlabs apt source

rrotter:
This apt source is no longer maintained.


daaang (approved):
Looks right, could also just delete the test so there's less to delete in an hour, but it doesn't really matter.

PR #815, opened 2025-04-21
update haproxy tls settings

rrotter:
rm `/templates/profile/legacy/kubernetes/*`

update `haproxy.cfg`:
- disable weak ciphers (notably, disable DHE and RSA key exchange)
- require TLS 1.2 or 1.3, using current haproxy config syntax
- add backend cipher settings
- update some settings to debian 11 defaults
- allow client to choose cipher
- comments explaining provenance of config


templates/profile/kubernetes/haproxy/haproxy.cfg.erb
```diff
@@ -1,31 +1,49 @@
...
+  # Debian 11 defaults
   log  global
   mode  http
   option  httplog
   option  dontlognull
   timeout connect 5000
+  # 2018-08-10; prevent observed 504 errors
   timeout client  180000
   timeout server  360000
+
```
rrotter: I added this useless whitespace to make the files align. It would be better if we could add the `errorfile` directives and make them match. Then we could even use the one template in both places if we want. Thoughts?

PR #816, opened 2025-04-22
use TLSv1.3 for apache

rrotter:
Since apache only serves internal requests, we should have no need to support TLSv1.2.

- remove TLS cipher settings from all vhosts
- add global config for mod_ssl
  - only allow TLSv1.3
  - set curve preferences
  - don't set preferred cipher, as all TLSv1.3 ciphers are secure
  - accept client cipher preference


rrotter:
Closing. We should still do this, but this PR is a dead end.

PR #817, opened 2025-04-25
add read only mountpoint alert for /htapps

umjnega:


templates/profile/prometheus/rules.yml.erb
```diff
@@ -180,6 +180,13 @@ groups:
     for: 30m
     labels:
       severity: ticket
+  - alert: HTMountReadOnly
+    annotations:
+      summary: '{{$labels.hostname}}:{{$labels.mountpoint}} is READ ONLY.'
+    expr: 'node_filesystem_readonly{role=~"nebula::role::webhost::htvm.*",mountpoint="/htapps"} > 0'
+    for: 30m
```
aelkiss: I would recommend alerting much sooner than this, even with no `for`, or if you definitely need one `for: 1m`. This is not an expected condition at all, and if a web server goes in the pool in this state it will result in users getting 500 errors.

umjnega: I pulled those params from a different alert.   a quicker alert makes sense.   I also added it as a `page` rather than just a `ticket`, due to feeling like this is more of an urgent matter, rather than something that can wait until monday. Do you agree with that?

aelkiss: Agreed, this is a pants-on-fire, fix ASAP or support tickets start flooding in about errors.

skorner (approved):
I'd prefer to remediate this problem in an automated way, but in the absence of that and also as a backstop to that, this is useful.

daaang:
Since we are implementing an automated fix, I'm `+0` and not `-1` on this `for: 1m`. I do feel the need to state that the `for` should almost always be at least 5 minutes if only because we are talking about human intervention, and I'll quote our [on-call expectations](https://mlit.atlassian.net/wiki/spaces/AE/pages/52527377/On-call+expectations) directly:

> within 30 minutes between the hours of 08:00 and 00:00

So if you truly need response within `1m` you truly need this to be an automated fix.

aelkiss:
I agree an automated fix would be preferable. That said, my expectation (and so far the case has been) that this would only happen during the reboot process. During the reboot process, I'm assuming that someone is keeping an eye on things at all times and that response could be more or less immediate -- or at least before the machines go back in the pool. I could see `for: 5m` if the reasoning is that this could be something someone catches and fixes before the alert fires.

aelkiss:
Another thing to consider could be getting the node metrics from the A&E prometheus also into the HathiTrust prometheus, so that we could deliver alerts to our own #alerts channel in whatever way makes sense for our own service needs (admittedly this closed PR isn't the place to discuss that, but I'll add it to my agenda to discuss with @rrotter at some point)

PR #818, opened 2025-04-29
don't use puppet apt repo for vm creation

rrotter:
puppet apt repo is gone. install our apt repo instead.


daaang (approved):
Couple things:

1. We want to create new VMs a new way, not with this preseed. This can lead to thinking of what we have as "legacy" which in turn leads to a disdain and refusal to maintain it, which in turn leads to rotting the code we run on. So I want to be careful.
2. This is hard-coding a detail (our new apt repo url) on two lines of a file for bullseye that I fully expect to be copied into a similar file for bookworm, which smells like an unnamed parameter.
3. Premature abstractions are the worst thing to do.

I want to acknowledge thing 1 on the list, when I say that thing 3 is the reason I'm giving this a LGTM.

rrotter:
This can be merged after we image a VM with it.

PR #821, opened 2025-05-02
Add etcd db size alerts

daaang:
EtcdDatabaseFillingUpFast should probably be `severity: page`, but I don't want to do that without letting this live for a while.

EtcdDatabaseTooBig exists mostly so that alerts don't stop when the db finishes running out of space. It could be set closer to 100% full, but for years, "normal" etcd db size has been strictly less than 25% full, so 75% seems reasonably high to say there is a problem, even if it's no longer increasing.

This week's incident would have fired EtcdDatabaseFillingUpFast but would not have fired EtcdDatabaseTooBig, as it never got above 1.2Gi, so I'd hear an argument for decreasing the threshold to e.g. 1Gi.

### DIFF
--- a/templates/profile/prometheus/rules.yml.erb
+++ b/templates/profile/prometheus/rules.yml.erb
@@ -353,6 +353,24 @@ groups:
     annotations:
       summary: 'MySQL Connection Count is approaching MAX on {{$labels.hostname}}'
       description: 'MySQL Connection Count is approaching MAX'
+  - alert: "EtcdDatabaseFillingUpFast"
+    expr: >
+      max by (datacenter) (delta(etcd_debugging_mvcc_db_total_size_in_bytes[5m])) > 10000
+    for: "30m"
+    labels:
+      severity: "ticket"
+    annotations:
+      summary: "{{$labels.datacenter}} etcd db filling up fast"
+      dashboard: "https://grafana.lib.umich.edu/d/aeihrf5qwmccge/all-etcd-clusters"
+  - alert: "BigEtcdDatabase"
+    expr: >
+      max by (datacenter) (etcd_debugging_mvcc_db_total_size_in_bytes) > 500 * 1024 * 1024
+    for: "30m"
+    labels:
+      severity: "ticket"
+    annotations:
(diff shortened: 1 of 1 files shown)


templates/profile/prometheus/rules.yml.erb
```diff
@@ -353,6 +353,24 @@ groups:
     annotations:
       summary: 'MySQL Connection Count is approaching MAX on {{$labels.hostname}}'
       description: 'MySQL Connection Count is approaching MAX'
+  - alert: "EtcdDatabaseFillingUpFast"
+    expr: >
+      max by (datacenter) (delta(etcd_debugging_mvcc_db_total_size_in_bytes[5m])) > 10000
```
botimer, line 358: This means consistently +10KB every 5mins for 30 minutes, right? So, I think, as written with delta:

- Bursts would not trigger this threshold (I think delta 500MB over 20 minutes would fall to 0 within the window)
- The alert slope may be too low / noisy (it seems like 60KB would be quite easy to hit with CRDs and CI/CD)

I'm not sure, but this seems like a case for a subquery (or `deriv`):

https://prometheus.io/blog/2019/01/28/subquery-support/#motivation
https://prometheus.io/docs/prometheus/latest/querying/examples/#subquery
https://prometheus.io/docs/prometheus/latest/querying/functions/#deriv
https://grafana.com/blog/2019/04/01/how-were-using-prometheus-subqueries-at-grafana-labs./

What do you think?

daaang: Screenshot from 2025-05-02 11-19-53

All I did was find a threshold noisy enough to pick up clusters that weren't having trouble but quiet enough that `for: 30m` addresses that.

daaang: Which is to say, when this was happening to us in real time, the delta was greater than 1Mi, constantly, for a period of 15 hours, so a threshold of 10Ki for 0.5 hours should start firing alerts if this happens again.

botimer: Huh, can't argue with that graph. Ship it!

botimer (approved):

PR #822, opened 2025-05-03
root bashrc: colorize `ip`, `rgrep`

rrotter:
- `ip -color` make ip approximately 100x more readable
- `rgrep --color` added for consistancy

### DIFF
--- a/files/root/.bashrc
+++ b/files/root/.bashrc
@@ -19,6 +19,8 @@ alias la='ls -AF'
 alias grep='grep --color=auto'
 alias fgrep='fgrep --color=auto'
 alias egrep='egrep --color=auto'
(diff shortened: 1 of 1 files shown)


daaang:
I'm `+0` on this, but I will point out that `ls --color=auto | hexdump -C` and `grep --color=auto | hexdump -C` both show no ANSI colors, while `ip -c a | hexdump -C` retains the ANSI colors. In other words `ip -color` is more like `--color=always` than `--color=auto`.

But I can't remember the last time I actually wanted to pipe an `ip` command into anything, so I'm `+0` on this.

rrotter:
Turns out there is a `-color=auto` also. Updated to use that.

PR #824, opened 2025-05-05
pdk update

daaang:


rrotter:
I'm surprised the diff is as small as it is on 4c71d9b. The first thing I would look at to get this to work is to see if there's anything we can/should override in `.sync.yml`.

PR #825, opened 2025-05-06
Increase haproxy maxconn

aelkiss:
See description in commit.


skorner (approved):
Recent analysis indicates we could use the increased headroom globally, and the behavior when hitting either the global or front/back end limits is not helpful (silently queuing requests in the haproxy server's kernel socket queue).

PR #826, opened 2025-05-07
Add test.search.lib ServerAlias

botimer:
To test out weighted round-robin DNS, add a test alias hostname.

### DIFF
--- a/manifests/profile/www_lib/vhosts/search.pp
+++ b/manifests/profile/www_lib/vhosts/search.pp
@@ -14,7 +14,8 @@ class nebula::profile::www_lib::vhosts::search (
   String $ssl_cn = 'search.lib.umich.edu',
   String $docroot = '/www/search/current/public'
 ) {
-  $servername = "${prefix}search.${domain}"
+  $servername    = "${prefix}search.${domain}"
+  $serveraliases = ["test.search.${domain}"]
 
   file { "${apache::params::logroot}/search":
     ensure => 'directory',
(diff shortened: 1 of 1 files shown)


daaang (approved):
This is placing itself in "so simple it's incomprehensible" territory for me. I think what I'm seeing is

1. The `$servername` line is not actually changing, there are just some added spaces to make the puppet linter happy.
2. I assume `serveraliases` defaults to `[]` and it's just some extra domains that the webserver can know point to the same place, so `test.search` can use the same vhost.

Not sure what's up with `${prefix}search` and why that exists yet does not cover `prefix = test.` but I'm pretty sure I don't have to be sure about that. As long as I'm right about 1 and 2, lgtm.

(It's cute how we're so careful to parameterize `$domain` and yet `search.lib.umich.edu` is, like, right there, hardcoded into `ssl_cn`)

botimer:
Yes, that's all about right. This profile is included exactly once with an empty "prefix".

The `$serveraliases` parameter is passed along from [nebula::apache::www_lib_vhost](https://github.com/mlibrary/nebula/blob/production/manifests/apache/www_lib_vhost.pp#L8), unmodified, to the base [apache::vhost](https://github.com/mlibrary/nebula/blob/production/manifests/apache/www_lib_vhost.pp#L141).

I'd have considered using `prefix` to _add_ a test vhost instead of modify this one, but we include it in [a loop](https://github.com/mlibrary/nebula/blob/9cf995d3b128bfa9239bf055901cf88440b554c0/manifests/profile/www_lib/apache/misc.pp#L53-L58) where the [empty prefix](https://github.com/mlibrary/nebula/blob/9cf995d3b128bfa9239bf055901cf88440b554c0/manifests/role/webhost/www_lib_vm.pp#L40) is shared across a bunch of "misc" sites. Not open to this extension, so I chose modification.

In all, this change results in the search http and https conf files each receiving a single new line:

```
ServerAlias test.search.lib.umich.edu
```

PR #829, opened 2025-05-27
Prometheus certificate keys should be 600 not 644

daaang:
Noticed this while renewing. These aren't servers anyone should be able to connect to directly, but security is about layers, so let's tighten this layer.

### DIFF
--- a/manifests/profile/prometheus.pp
+++ b/manifests/profile/prometheus.pp
@@ -123,14 +123,26 @@ class nebula::profile::prometheus (
 
   file { '/etc/prometheus/tls/ca.crt':
     source => 'puppet:///ssl-certs/prometheus-pki/ca.crt',
+    mode   => '0644',
+    owner  => 'nobody',
+    group  => 'nogroup',
+    notify => Docker::Run['prometheus'],
   }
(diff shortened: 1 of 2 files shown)


daaang:
I converted this to a draft the moment I realized that prometheus currently runs as `nobody:nogroup` which obviously won't be able to read a `600` file owned by `root:root`. Ways forward:

1. Backward instead of forward: close this PR
2. Add `owner => 'nobody'` and `group => 'nogroup'` to the tls files—probably harmless, easy to check with a dev environment.
3. Set up `prometheus` and `pushgateway` users and possibly a `prometheus` group for them to be members of, and make the key `640` instead of `600`.

I'm fine with 1; I think 2 is probably the way to go; and I think 3 is the right thing to do but also more trouble than this problem is worth.

botimer:
Agreed. The `nobody` answer isn't great, but it is better than 644. A user for the daemon would be nice... worth it now? I don't have a strong opinion.

daaang:
I've tested this on macc-prometheus and on prometheus-3.macc.test, and it works as expected.

PR #836, opened 2025-06-25
rm sudo group

rrotter:
Don't install sudo group by default.

If we want to do this, this is almost certainly not the way. Just floating the idea.


rrotter:
nah, this needs some cleanup, but this is the wrong first step.

PR #837, opened 2025-06-25
Don't add useless users on ht prod

rrotter:
- don't install afs on htprod
- don't install sudo group on htprod


rrotter:
Probably should do this eventually, but it's not a current priority.

PR #846, opened 2025-08-07
start puppet if it's stopped

rrotter:


daaang (changes requested):
Since git blame is how I found the cause of the problem, I think it's worth noting in the commit (or with a comment) that, because we can disable all puppet agent activity with `puppet agent --disable`, there should not exist a situation where both you want to run `puppet agent --test` **and** you want the systemd service to be stopped.

Assuming that's right, ensuring the systemd service is running makes installation and agent upgrades easier, so we prefer it.

daaang (approved):
Nice!

PR #849, opened 2025-08-12
openjdk-8-jre has been replaced with nvidia-openjdk-8-jre

umjnega:


rrotter (changes requested):
Let's not do this. We have other ways to install openjdk, but let's figure out why it's missing.

PR #850, opened 2025-08-12
remove openjdk-8-jre

umjnega:
Debian has decided that this package is now obsolete and removed it from the Debian 11 repos.   Was caught when doing a dist-upgrade that uninstalled it.


rrotter:
Would it make more sense to change this line instead? Replace it with temurin 8 and/or openjdk 11?

rrotter (approved):
Since we confirmed that we're not installing java on these anymore: lgtm!

PR #851, opened 2025-08-20
Don't manage firewall on k8s hosts

rrotter:
- role::minimum: make firewall management optional
- controller/worker/etcd: disable firewall management

These nodes are all behind NAT. Puppet management of host firewall is redundant and conflicts with calico.

- firewall profile: rm support for internal_routing = kubernetes_calico, and reject any unrecognised options

### DIFF
--- a/manifests/profile/networking/firewall.pp
+++ b/manifests/profile/networking/firewall.pp
@@ -74,37 +74,20 @@ class nebula::profile::networking::firewall (
         }
       }
 
-      'kubernetes_calico': {
-        $input_ignore = [
-          '-j cali-INPUT',
-          '-j KUBE-FIREWALL',
-          '-j KUBE-SERVICES',
-          '-j KUBE-EXTERNAL-SERVICES',
-        ]
-
-        $output_ignore = [
-          '-j cali-OUTPUT',
-          '-j KUBE-FIREWALL',
-          '-j KUBE-SERVICES',
-        ]
-
-        $forward_ignore = [
-          '-j cali-FORWARD',
-          '-j KUBE-FORWARD',
-          '-j KUBE-SERVICES',
-        ]
-      }
-
       'fwknop': {
         $input_ignore = ['-j FWKNOP_INPUT']
         $output_ignore = []
         $forward_ignore = []
       }
 
-      default: {
+      '': {
         $input_ignore = []
         $output_ignore = []
         $forward_ignore = []
+      }
+
+      default: {
+        fail('internal_routing set to an unsupported option')
       }
     }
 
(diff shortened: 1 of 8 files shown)


manifests/profile/networking/firewall.pp
```diff
@@ -74,38 +74,21 @@
...
-      }
-
       'fwknop': {
         $input_ignore = ['-j FWKNOP_INPUT']
         $output_ignore = []
         $forward_ignore = []
       }
 
-      default: {
+      '': {
```
daaang, line 83: Tilting my head and raising an eyebrow at this. A conversation I missed? Looks like we're switching from having a default behavior to requiring `''` or erroring? Seems a little astonishing, but I don't know what I missed.

manifests/role/minimum.pp
```diff
@@ -17,8 +18,13 @@
...
 
-    class { 'nebula::profile::networking::firewall':
-      internal_routing => $internal_routing,
+    if ($manage_firewall) {
+      class { 'nebula::profile::networking::firewall':
+        internal_routing => $internal_routing,
+      }
+    } else {
+      package { 'netfilter-persistent': ensure => purged }
+      package { 'iptables-persistent': ensure => purged }
```
daaang, line 27: I like the look of this.

spec/classes/profile/networking/firewall_spec.rb
```diff
@@ -107,21 +107,7 @@
...
-              .with_purge(false)
-          end
-
-          it do
-            expect(subject).to contain_firewallchain("#{chain}:filter:IPv6")
-              .with_ensure("present")
-              .with_purge(false)
-          end
-        end
+        it { is_expected.not_to compile }
```
daaang, line 119: lol amazing, no notes

spec/classes/role/kubernetes_spec.rb
```diff
@@ -61,37 +61,14 @@
...
-            ensure: "present",
-            purge: false,
-            ignore: ["-j cali-FORWARD",
-              "-j KUBE-FORWARD",
-              "-j KUBE-SERVICES"]
-          )
+        it "does not configure firewallchains" do
+          is_expected.not_to contain_firewallchain("INPUT:filter:IPv4")
+          is_expected.not_to contain_firewallchain("OUTPUT:filter:IPv4")
+          is_expected.not_to contain_firewallchain("FORWARD:filter:IPv4")
```
daaang, line 71: ...hm. I'm just realizing that I have zero experience managing linux hosts without any firewall management, so I have no idea what to expect in real life. Calico is certainly just blitzing its own firewall chains, and linux hosts definitely continue having networking when iptables isn't installed.

I guess this is a neutral comment, sort of like replying with a popcorn emoji in slack I guess. I'm so used to thinking about INPUT, OUTPUT, and FORWARD as so basic that they're like `int main`, and expecting not to have them feels like dividing by zero.

daaang (approved):
It's a big enough change to warrant checking on the test cluster before merging, but at least in theory, a big :+1: from me.

PR #852, opened 2025-09-04
update deepblue.lib vhost config to support DSpace 7 running in Kubernetes instead of DSpace 6 on bulleit-2

umjnega:
Apache redirect for the new dspace in kubernetes.

### DIFF
--- a/manifests/profile/www_lib/vhosts/deepblue.pp
+++ b/manifests/profile/www_lib/vhosts/deepblue.pp
@@ -9,8 +9,8 @@
 # @example
 #   include nebula::profile::www_lib::vhosts::deepblue
 class nebula::profile::www_lib::vhosts::deepblue (
-  String $prefix,
   String $domain,
+  String $prefix,
   String $ssl_cn = 'deepblue.lib.umich.edu',
   String $docroot = '/www/deepblue/web'
 ) {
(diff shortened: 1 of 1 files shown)


manifests/profile/www_lib/vhosts/deepblue.pp
```diff
@@ -43,13 +43,19 @@
...
       {
         comment      => 'Deep Blue Repositories home page is on www.lib now',
         rewrite_cond => '%{REQUEST_URI} ^((\/?|/index.html)$|/splash/)',
         rewrite_rule => '^(.*)$	https://www.lib.umich.edu/collections/deep-blue-repositories [redirect=permanent,last]'
       },
+      #{
+      #  comment      => 'Temporarily disable login for go live',
+      #  rewrite_cond => '%{REQUEST_URI} !^(/data|/openid-connect)',
+      #  rewrite_rule => '^(.*)$ https://lib.umich.edu/outages [L]',
+      #},
```
skorner: We should remove this from the final commit.

skorner (changes requested):
Looks good with two changes:
- removing the commented out outages redirect
- editing the PR title to reflect the final purpose instead of the original purpose; perhaps something like "update deepblue.lib vhost config to support DSpace 7 running in Kubernetes instead of DSpace 6 on bulleit-2"

skorner (approved):
Looks good!

PR #853, opened 2025-09-11
Allow ping

daaang:
This puppet code translates to the following iptables rule getting added

    -A INPUT -p icmp -m icmp --icmp-type 8 -m limit --limit 5/sec \
        -m comment --comment "500 allow ping" -j ACCEPT

### DIFF
--- a/manifests/profile/networking/firewall.pp
+++ b/manifests/profile/networking/firewall.pp
@@ -162,6 +162,13 @@ class nebula::profile::networking::firewall (
     jump    => 'accept',
   }
 
+  firewall { '500 allow ping':
+    proto => 'icmp',
+    icmp  => 8,
+    limit => '5/second',
+    jump  => 'accept',
+  }
+
   firewall { '999 drop all':
     proto  => 'all',
     jump   => 'drop',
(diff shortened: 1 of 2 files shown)


daaang:
To be clear this is a proposed change, not based on any actual complaints or frustrations, in service of no principle other than "maybe our servers should behave a little more like normal servers." It's a pretty weak principle, so really any reason to reject this is probably a good one.

I'm putting this here so that "maybe our servers should behave a little more like normal servers" can outweigh "I don't know how to make puppet do this" as a basic principle.

daaang:
PS if we're specifically worried about smurf attacks, this should generate the couple other lines that appear to be standard lore on that

```diff
     jump    => 'accept',
   }

+  firewall { '498 guard against smurf attacks 1/2':
+    proto => 'icmp',
+    icmp  => 'address-mask-request',
+    jump  => 'drop',
+  }
+
+  firewall { '499 guard against smurf attacks 2/2':
+    proto => 'icmp',
+    icmp  => 'timestamp-request',
+    jump  => 'drop',
+  }
+
   firewall { '500 allow ping':
     proto => 'icmp',
     icmp  => 8,
```

botimer (approved):

PR #854, opened 2025-09-17
add kubernetes dns entries via hiera data

rrotter:
Adds new `static_dns_entries` key under kubernetes cluster hiera


manifests/profile/kubernetes/dns_server.pp
```diff
@@ -43,14 +44,19 @@
...
+    '/etc/hosts static entries':
+      content => template('nebula/profile/kubernetes/dns/hosts_20_static_entries.erb'),
+      order   => '20',
+    ;
+
     '/etc/hosts ipv6 localhost':
-      content => template('nebula/profile/kubernetes/dns/hosts_05_ipv6_localhost.erb'),
-      order   => '05',
+      content => template('nebula/profile/kubernetes/dns/hosts_50_ipv6_localhost.erb'),
+      order   => '50',
```
daaang, line 54: Any reason for the ordering changes?

spec/classes/profile/kubernetes/dns_server_spec.rb
```diff
@@ -66,10 +66,15 @@
         it do
           expect(subject).to contain_concat_fragment("/etc/hosts ipv6 localhost")
             .with_target("/etc/hosts")
-            .with_order("05")
```
daaang, line 69: This at least I can get behind, yes

daaang (approved):
I'm curious about the reordering, but not enough to slow anything down. I'm sure there's a reason.

I was thinking of using the datacenter hieradata, but your using the cluster stuff in common.yaml keeps everything together :+1:

manifests/profile/kubernetes/dns_server.pp
rrotter: Purely an asthetic choice. I thought /etc/hosts looked more tidy maintaining the ipv6 entries at the bottom and adding the new entries just above.

PR #855, opened 2025-09-25
Revert "Add cf.* serveralias for hathitrust web"

rrotter:
Reverts mlibrary/nebula#812

This alias is no longer needed.


aelkiss (approved):
Can't think of any reason why we would still need this.

PR #856, opened 2025-09-26
Allow hiera-managed /etc/hosts

daaang:


botimer (approved):
Looks good! Let's talk about it if we find we're adding a bunch of stuff here... A local, caching DNS server/dnsmasq may make sense in that case.

PR #860, opened 2025-10-16
Retain 5 days of kubernetes gateway haproxy logs

daaang:
Instead of 52 days.


skorner (approved):
Enthusiastic thumbs-up, especially considering we rely on centralized log aggregation for retention compliance.

PR #862, opened 2025-10-22
ETT-49: align HT catalog pear modules with dockerfile

aelkiss:
* Prep for migration to PHP 8
* Removes deprecated/unused modules
* Adds some that are explicitly specified in the Dockerfile - may not be needed (or are installed automatically) but worth keeping things in line


aelkiss:
See recent changes in the catalog Dockerfile: https://github.com/hathitrust/catalog/compare/5.0.28..main#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557

We should wait to merge this until after we've released & deployed the catalog.

moseshll (approved):
Looks good to me, APPROVE

aelkiss:
@rrotter I also don't think even after this is merged it will actively _remove_ anything but still better safe than sorry. We should consider going through and doing that manually (should be straightforward enough).

For the end state this list of modules should be sufficient to have installed (this is what ends up installed in the catalog docker image):

```
root@catalog-dev:/app# pear list
Installed packages, channel pear.php.net:
=========================================
Package               Version State
Archive_Tar           1.4.12  stable
Console_Getopt        1.4.3   stable
DB                    1.12.2  stable
DB_DataObject         1.11.5  stable
Date                  1.4.7   stable
File                  1.4.1   stable
File_CSV              1.0.0   stable
File_MARC             1.4.1   stable
File_Util             1.0.0   stable
HTTP_Request2         2.7.0   stable
HTTP_Session2         0.7.3   beta
Log                   1.14.6  stable
MDB2                  2.5.0b5 beta
MDB2_Driver_mysql     1.5.0b4 beta
Net_URL2              2.2.3   stable
PEAR                  1.10.12 stable
PEAR_Manpages         1.10.0  stable
PHP_Compat            1.5.0   stable
Pager                 2.5.1   stable
Structures_Graph      1.1.1   stable
Structures_LinkedList 0.2.2   beta
XML_Util              1.4.5   stable
```

rrotter:
Looks like `ensure => "absent"` is supported for extensions if you want them gone. Can always follow up with a second PR that clears the redundant code. 

https://github.com/voxpupuli/puppet-php/blob/master/manifests/extension.pp

aelkiss:
I'll think more about what we want to do here -- I probably will want to clean this up on e.g. preview and then set ensure absent.

rrotter (changes requested):
I'd like this to ensure absent on the unwanted extensions. Once this is in place just let me know whether/when you'd like it merged or if we should run it in an environment on a test host first.

aelkiss:
@rrotter Assuming tests pass let's give this a try as an environment on ht-web-preview. No tremendous rush.

rrotter (approved):
This completes successfully now. It took a couple of runs before completing w/o errors, but I have no problem w/ eventual consistancy.

aelkiss:
Verified preview.catalog.hathitrust.org is working as expected. I think you can go ahead and merge this whenever is convenient.

PR #866, opened 2025-10-28
Add metric label and alerts for HAProxy sessions and maxconn

daaang:


botimer (approved):

botimer:
We had a pretty lengthy discussion on the naming here, and the topology doesn't quite line up for anything more precise. It's not quite the load balancer/load balancing pair, nor a service, cluster, backend, or anything that isn't already overloaded. So, `load_balancer_for` is at least descriptive of the somewhat amorphous _unnamed conceptual thing that is load-balanced_ without being misleading. 🚢

PR #871, opened 2025-11-05
Add a preseed for debian 13 trixie

daaang:
what if it just works

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -36,7 +36,7 @@
   "operatingsystem_support": [
     {
       "operatingsystem": "Debian",
-      "operatingsystemrelease": ["11","12"]
+      "operatingsystemrelease": ["11","12","13"]
     },
     {
       "operatingsystem": "Ubuntu",
--- a/templates/virtual_machine/trixie.cfg.erb
+++ b/templates/virtual_machine/trixie.cfg.erb
@@ -1,5 +1,5 @@
-# Managed by puppet (nebula/virtual_machine/bullseye.cfg.erb)
-#### Contents of the preconfiguration file (for bullseye)
+# Managed by puppet (nebula/virtual_machine/trixie.cfg.erb)
+#### Contents of the preconfiguration file (for trixie)
 ### Localization
 # Preseeding only locale sets language, country and locale.
 d-i debian-installer/locale string en_US.UTF-8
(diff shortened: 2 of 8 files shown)


templates/virtual_machine/trixie.cfg.erb
```diff
@@ -0,0 +1,493 @@
...
+#d-i apt-setup/local0/source boolean true
+# URL to the public key of the local repository; you must provide a key or
+# apt will complain about the unauthenticated repository and so the
+# sources.list line will be left commented out.
+#d-i apt-setup/local0/key string http://local.server/key
+# If the provided key file ends in ".asc" the key file needs to be an
+# ASCII-armoured PGP key, if it ends in ".gpg" it needs to use the
+# "GPG key public keyring" format, the "keybox database" format is
+# currently not supported.
+d-i apt-setup/local0/repository string https://apt.lib.umich.edu bullseye main
```
rrotter: We have bookworm on this apt server. I'll try to add trixie today.

templates/virtual_machine/trixie.cfg.erb
```diff
@@ -0,0 +1,493 @@
...
+# This command is run just before the install finishes, but when there is
+# still a usable /target directory. You can chroot to /target and use it
+# directly, or use the apt-install and in-target commands to easily install
+# packages and run commands in the target system.
+#d-i preseed/late_command string apt-install zsh; in-target chsh -s /bin/zsh
+
+# install puppet.conf, firewall
+d-i preseed/late_command string                       \
+    in-target apt-get update -qq;                     \
+    in-target apt-get install -qq -y puppet-agent;    \
```
rrotter: Will install this: https://packages.debian.org/trixie/puppet-agent

I guess I/we/someone should partially revert #818

metadata.json
```diff
@@ -36,7 +36,7 @@
   "operatingsystem_support": [
     {
       "operatingsystem": "Debian",
-      "operatingsystemrelease": ["11","12"]
+      "operatingsystemrelease": ["11","12","13"]
```
rrotter, line 39: 😁

templates/virtual_machine/trixie.cfg.erb
rrotter: Oops, I think this in only here to install puppet. So what we actually need is to switch to the vox repo instead.

templates/virtual_machine/trixie.cfg.erb
daaang: Oh good point. I guess there's little reason not to get going with vox out the gate, long as setting that up isn't too weird.

metadata.json
daaang: Here's to

```diff
-11,12,13
+13
```

(even if `12,13` is more likely...)

templates/virtual_machine/trixie.cfg.erb
rrotter: #872 will fix this

PR #872, opened 2025-11-05
preseed: install openvox-agent, qemu-guest-agent; support bookworm

rrotter:
Remove our repo from preseed because it was only there installing puppet. Replace w/ openvox, and use updated package name for install.


daaang (approved):

rrotter:
added qemu-guest-agent to preseed, created initial bookworm preseed, rebased

If this all works fine I'll merge it this morning. If there are issues w/ bookworm I'll probably just remove that final commit and merge w/ bullseye only.

PR #874, opened 2025-11-10
support mounting all ht nfs mounts on truenas

rrotter:
use_truenas flag moves all "smartconnect" mounts to truenas

This is an interim solution: "smartconnect" is an isilon feature, but for now I'm using it as "other mounts that need to be mounted from whatever nfs server we're using for HT".

### DIFF
--- a/manifests/profile/hathitrust/mounts.pp
+++ b/manifests/profile/hathitrust/mounts.pp
@@ -36,12 +36,23 @@ class nebula::profile::hathitrust::mounts (
     options => "size=${ramdisk_size}"
   }
 
+  # TODO: rename smartconnect_mounts, remove support for smartconnect
   $smartconnect_mounts.each |$mount| {
-    nebula::nfs_mount { $mount:
-      remote_target   => "nas-${facts['datacenter']}.sc:/ifs${mount}",
-      tag             => 'smartconnect',
-      private_network => true,
-      monitored       => true
+    if($use_truenas) {
+      nebula::nfs_mount { $mount:
+        options         => 'auto,hard',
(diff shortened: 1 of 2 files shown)


manifests/profile/hathitrust/mounts.pp
```diff
@@ -36,12 +36,23 @@
...
+        monitored       => true
+      }
+    }
+    else {
+      nebula::nfs_mount { $mount:
+        remote_target   => "nas-${facts['datacenter']}.sc:/ifs${mount}",
+        tag             => 'smartconnect',
+        private_network => true,
+        monitored       => true
+      }
```
daaang, line 55: It looks to me like we used to only mount smartconnect mounts here, and we now have a second kind of mount, and the old smartconnect mount has been moved to an else branch for when `$use_truenas` is `false`.

It also looks like the intention is to turn this back into a nonbranching code block, with truenas mounts replacing smartconnect mounts. Seems fine. Do we also intend to get rid of `$use_truenas`? I only ask because boolean parameters are a smell (where we might prefer separate named profiles or defined types—but if it's temporary it's not an issue).

daaang (approved):

manifests/profile/hathitrust/mounts.pp
rrotter: Yes, $use_truenas will go away, because soon everything will [use truenas].

rrotter: We'll still need the `$smartconnect_mounts` variable, but the name should change to something else. Right now I'm using it for both truenas and smartconnect mounts.

PR #875, opened 2025-11-11
ht php: rm ensure absent

rrotter:
These will no longer be needed once puppet has run on all affected hosts. I'll wait a few days to merge this, at least until there are no outstanding environments running on ht apache hosts.


aelkiss (approved):
Makes sense to me. Just noting right now the only machines with modules still left to remove (so far as I can tell) are macc-ht-web-139 and ictc-ht-web-208, which are on the `ht_mnt_htapps` branch.

PR #877, opened 2025-11-17
move vhosts ssl cert from www.lib to lib.umich.edu

umjnega:


manifests/profile/www_lib/vhosts/www_lib.pp
umjnega, line 1: I think this needed to be added for the ssl cert to actually get updated in the vhost file.

manifests/profile/www_lib/apache/misc.pp
```diff
@@ -44,6 +44,7 @@
       'search.lib.umich.edu',
       'theater-historiography.org',
       'www.lib.umich.edu',
```
skorner: Might as well remove this final www.lib reference as part of this first round.

skorner (approved):
Looks good. I left one comment. And before merging you'll want to squash and also update the cert per my Jira comment. Thanks!

PR #878, opened 2025-11-18
update apt source to use DEB822 source format

aayuraje:


manifests/profile/apt.pp
```diff
@@ -72,24 +72,28 @@
...
-      release  => "${facts['os']['name'].downcase()}${facts['os']['release']['major']}",
-      repos    => $puppet_repo,
-      key      => {
-        name   => 'openvox.asc',
-        source => 'puppet:///modules/nebula/apt/keyrings/openvox.asc',
-      },
+      source_format => 'sources',
+      location      => ['https://apt.voxpupuli.org'],
+      release       => "${facts['os']['name'].downcase()}${facts['os']['release']['major']}",
+      repos         => $puppet_repo,
```
rrotter:
```suggestion
      repos         => [$puppet_repo],
```
Looks like you turned this parameter into an array above. We'll never need two repos here, so maybe change it back and just make it into an array here.

rrotter (changes requested):

manifests/profile/apt.pp
aayuraje: Got it, I've made this change.

spec/classes/profile/kubernetes/kubelet_spec.rb
```diff
@@ -76,7 +76,7 @@
 
         it do
           expect(subject).to contain_apt__source("kubernetes").with(
-            location: "https://pkgs.k8s.io/core:/stable:/v1.29/deb/",
+            location: ["https://pkgs.k8s.io/core:/stable:/v1.29/deb/"],
```
rrotter, line 79: Test is failing because we're still testing for apt::source:key which is a field we're not using anymore. In addition to removing this, we should be testing for the keyring that replaces it.


```
        it "contains k8s.io apt source" do
          is_expected.to contain_apt__source("kubernetes")
            .with_keyring("/etc/apt/keyrings/k8s.io.asc")
            .with_location(["https://pkgs.k8s.io/core:/stable:/v1.29/deb/"])
            .with_release("/")
          is_expected.to contain_apt__keyring("k8s.io.asc")
            .with_source("puppet:///modules/nebula/apt/keyrings/k8s.io.asc")
        end
```

spec/fixtures/hiera/default.yaml
```diff
@@ -168,4 +168,4 @@ nebula::jdk_version: '11'
 
 nebula::profile::kubelet::kubelet_version: default.invalid
 
-nebula::profile::kubernetes::apt::location: 'https://pkgs.k8s.io/core:/stable:/vX.YZ/deb/'
+nebula::profile::kubernetes::apt::location: ['https://pkgs.k8s.io/core:/stable:/vX.YZ/deb/']
```
rrotter: We should avoid changes that will break existing keys in heiradata unless there is a strong reason to do so.

I'm fairly sure we don't need multiple values here.

spec/fixtures/hiera/kubernetes/default.yaml
```diff
@@ -83,4 +83,4 @@ nebula::profile::kubernetes::clusters:
 nebula::cifs::credentials::users:
 - kubernetes
 
-nebula::profile::kubernetes::apt::location: 'https://pkgs.k8s.io/core:/stable:/v1.29/deb/'
+nebula::profile::kubernetes::apt::location: ['https://pkgs.k8s.io/core:/stable:/v1.29/deb/']
```
rrotter: Same comment as on `‎spec/fixtures/hiera/default.yaml‎`. This should be reverted.

manifests/profile/apt.pp
```diff
@@ -72,24 +72,28 @@
...
-          'source' => "puppet:///modules/nebula/apt/keyrings/${hpe_key}",
-        },
+        source_format => 'sources',
+        location      => ['https://downloads.linux.hpe.com/SDR/repo/mcp'],
+        release       => "${facts['os']['distro']['codename']}/current",
+        repos         => ['non-free'],
+        keyring       => "/etc/apt/keyrings/${hpe_key}",
+      }
+
+      apt::keyring { 'hpe.asc':
```
rrotter: By convention `.gpg` keys are binary and `.asc` is the extension for ascii armored keys. I'm not 100% sure whether the name is actually used by apt for parsing keys, but I'd like to stick with the convention.
```suggestion
      apt::keyring { "${hpe_key}":
```

manifests/profile/yarn.pp
```diff
@@ -8,16 +8,17 @@
 class nebula::profile::yarn (
 ) {
   apt::source { 'yarn':
```
rrotter, line 10: missing keyring field
```suggestion
  apt::source { 'yarn':
    keyring       => '/etc/apt/keyrings/yarnpkg.asc'
```

manifests/profile/apt/mono.pp
```diff
@@ -14,12 +14,13 @@
...
-    location => 'https://download.mono-project.com/repo/debian',
-    release  => "stable-${apt_release}",
-    repos    => 'main',
-    key      => {
-      'name'   => 'mono-project.asc',
-      'source' => 'https://download.mono-project.com/repo/xamarin.gpg',
-    },
+    source_format => 'sources',
+    location      => ['https://download.mono-project.com/repo/debian'],
+    release       => "stable-${apt_release}",
```
rrotter, line 19: keyring?

rrotter (changes requested):
Looks like there are a few `apt::source` resources missing the `keyring` field. I'd also like to see the new `apt::keyring` resources checked in the spec files.

spec/classes/profile/kubernetes/kubelet_spec.rb
aayuraje: yes, i noticed that. fixing them and rebasing the changes

manifests/profile/apt/mono.pp
aayuraje: Added keyring

manifests/profile/yarn.pp
aayuraje: Added keyring

manifests/profile/apt.pp
aayuraje: You're correct. Fixed now

spec/fixtures/hiera/default.yaml
aayuraje: fixed

spec/fixtures/hiera/kubernetes/default.yaml
aayuraje: fixed

manifests/profile/apt/yaz.pp
```diff
@@ -8,15 +8,17 @@
...
       'deb' => true,
     },
-    key      => {
-      'name'   => 'yaz-indexdata.dk.gpg',
-      'source' => 'https://download.indexdata.com/debian/indexdata.gpg',
-    },
+  }
+
+  apt::keyring { 'yaz-indexdata.dk.gpg':
+    source => 'https://download.indexdata.com/debian/indexdata.gpg',
```
rrotter: We need to remove the remote source here. Most of these have been moved to files/apt/keyrings. This should be moved there too.

manifests/profile/apt.pp
rrotter: Can remove this TODO!

manifests/profile/apt.pp
```diff
@@ -108,20 +112,23 @@
     # TODO: remove non-free where we're not using it
     # TODO: remove branch when we're off bullseye
     $debian_repos = $facts['os']['distro']['codename'] ? {
-      'bullseye' => 'main contrib non-free',
-      default    => 'main contrib non-free non-free-firmware',
+      'bullseye' => ['main', 'contrib', 'non-free'],
+      default    => ['main', 'contrib', 'non-free', 'non-free-firmware'],
     }
+
     apt::source { 'main':
```
rrotter, line 119: Probably fine that there is no keyring on the official Debian/Ubuntu repos for now.

Ultimately we want the generated version of these files to be exactly the same as what they are on a fresh Debian install. Looking into this can wait for future development, certainly not a blocker here.

manifests/profile/elastic.pp
```diff
@@ -20,19 +20,20 @@
...
+    release       => 'stable',
+    repos         => ['main'],
+    keyring       => '/etc/apt/keyrings/elastic.co.asc',
+    include       => {
       'src' => false,
       'deb' => true,
     },
   }
+  apt::keyring { 'elastic.co.asc':
+    source => 'https://artifacts.elastic.co/GPG-KEY-elasticsearch',
```
rrotter: Another key to move into nebula.

manifests/profile/loki.pp
```diff
@@ -15,13 +15,15 @@
...
-    },
+    source_format => 'sources',
+    location      => ['https://apt.grafana.com/'],
+    release       => '',
+    repos         => ['stable', 'main'],
+    keyring       => '/etc/apt/keyrings/grafana.asc',
+  }
+
+  apt::keyring { 'grafana.asc':
+    source => 'https://apt.grafana.com/gpg.key',
```
rrotter: One more…

rrotter (commented):
Looks good. I'd prefer to get rid of all the external key sources, but there are quite a few of them.

I'll leave it up to you. I'm fine with remediating some/all of those in this PR, or leaving them for later.

The only thing I definitly want changed is removing that `# TODO: port to DEB822` comment, since you just did.

Thanks!

manifests/profile/apt.pp
aayuraje: :white_check_mark:

manifests/profile/apt.pp
aayuraje: I added a TODO for this.

manifests/profile/loki.pp
aayuraje: moved

rrotter (approved):
I hope you don't mind I pushed an update to `/files/apt/keyrings/README` to include the keyrings you added.

Looks good. As we discussed, let's just install this on a few hosts w/ a puppet environment and make sure there are no surprises before merging.

PR #879, opened 2025-11-20
Move sshd_config to sshd_config.d

rrotter:
### `sshd_config` / `ssh_config`
- add default distro versions of sshd_config, ssh_config to revert previous changes
- move our customizations to drop-ins in sshd_config.d, ssh_config.d
  - remove non-customizations (don't need to set settings that are already set the distro default config, unless they are very important)

### `pam.d/sshd`
- drop-in not supported here, relocate w/ `dpkg-divert`

Also see: #882, which removes the temporary code added here.

### DIFF
--- a/files/default/bookworm/etc/ssh/ssh_config
+++ b/files/default/bookworm/etc/ssh/ssh_config
@@ -1,0 +1,53 @@
+
+# This is the ssh client system-wide configuration file.  See
+# ssh_config(5) for more information.  This file provides defaults for
+# users, and the values can be changed in per-user configuration files
+# or on the command line.
+
+# Configuration data is parsed as follows:
+#  1. command line options
+#  2. user-specific file
+#  3. system-wide file
+# Any configuration value is only changed the first time it is set.
+# Thus, host-specific definitions should be at the beginning of the
+# configuration file, and defaults at the end.
+
+# Site-wide defaults for some commonly used options.  For a comprehensive
+# list of available options, their meanings and defaults, please see the
+# ssh_config(5) man page.
+
+Include /etc/ssh/ssh_config.d/*.conf
+
+Host *
+#   ForwardAgent no
+#   ForwardX11 no
+#   ForwardX11Trusted yes
+#   PasswordAuthentication yes
+#   HostbasedAuthentication no
+#   GSSAPIAuthentication no
+#   GSSAPIDelegateCredentials no
+#   GSSAPIKeyExchange no
+#   GSSAPITrustDNS no
+#   BatchMode no
+#   CheckHostIP yes
+#   AddressFamily any
+#   ConnectTimeout 0
+#   StrictHostKeyChecking ask
+#   IdentityFile ~/.ssh/id_rsa
+#   IdentityFile ~/.ssh/id_dsa
+#   IdentityFile ~/.ssh/id_ecdsa
+#   IdentityFile ~/.ssh/id_ed25519
+#   Port 22
+#   Ciphers aes128-ctr,aes192-ctr,aes256-ctr,aes128-cbc,3des-cbc
+#   MACs hmac-md5,hmac-sha1,umac-64@openssh.com
+#   EscapeChar ~
+#   Tunnel no
+#   TunnelDevice any:any
+#   PermitLocalCommand no
+#   VisualHostKey no
+#   ProxyCommand ssh -q -W %h:%p gateway.example.com
+#   RekeyLimit 1G 1h
+#   UserKnownHostsFile ~/.ssh/known_hosts.d/%k
+    SendEnv LANG LC_*
+    HashKnownHosts yes
+    GSSAPIAuthentication yes
--- a/files/default/trixie/etc/ssh/sshd_config
+++ b/files/default/trixie/etc/ssh/sshd_config
@@ -1,0 +1,124 @@
+
+# This is the sshd server system-wide configuration file.  See
+# sshd_config(5) for more information.
+
+# This sshd was compiled with PATH=/usr/local/bin:/usr/bin:/bin:/usr/games
+
+# The strategy used for options in the default sshd_config shipped with
+# OpenSSH is to specify options with their default value where
+# possible, but leave them commented.  Uncommented options override the
+# default value.
+
+Include /etc/ssh/sshd_config.d/*.conf
+
+#Port 22
+#AddressFamily any
+#ListenAddress 0.0.0.0
+#ListenAddress ::
+
+#HostKey /etc/ssh/ssh_host_rsa_key
+#HostKey /etc/ssh/ssh_host_ecdsa_key
+#HostKey /etc/ssh/ssh_host_ed25519_key
+
+# Ciphers and keying
+#RekeyLimit default none
+
+# Logging
+#SyslogFacility AUTH
+#LogLevel INFO
+
+# Authentication:
+
+#LoginGraceTime 2m
+#PermitRootLogin prohibit-password
+#StrictModes yes
+#MaxAuthTries 6
+#MaxSessions 10
+
+#PubkeyAuthentication yes
+
+# Expect .ssh/authorized_keys2 to be disregarded by default in future.
+#AuthorizedKeysFile	.ssh/authorized_keys .ssh/authorized_keys2
+
+#AuthorizedPrincipalsFile none
+
+#AuthorizedKeysCommand none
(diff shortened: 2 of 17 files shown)


files/default/bookworm/etc/ssh/ssh_config
```diff
@@ -0,0 +1,53 @@
...
+#   EscapeChar ~
+#   Tunnel no
+#   TunnelDevice any:any
+#   PermitLocalCommand no
+#   VisualHostKey no
+#   ProxyCommand ssh -q -W %h:%p gateway.example.com
+#   RekeyLimit 1G 1h
+#   UserKnownHostsFile ~/.ssh/known_hosts.d/%k
+    SendEnv LANG LC_*
+    HashKnownHosts yes
```
daaang, line 52: HashKnownHosts isn't my favorite, and it seems especially silly for our servers, but I'm guessing this is just copying in debian defaults (and I don't dislike anywhere near enough to say "let's don't do defaults").

files/default/trixie/etc/ssh/sshd_config
```diff
@@ -0,0 +1,124 @@
...
+
+# override default of no subsystems
+Subsystem	sftp	/usr/lib/openssh/sftp-server
+
+# Example of overriding settings on a per-user basis
+#Match User anoncvs
+#	X11Forwarding no
+#	AllowTcpForwarding no
+#	PermitTTY no
+#	ForceCommand cvs server
```
daaang, line 124: Yeah I'm not running diffs on these. I'm just assuming you copied in three default config files bundled with each of the four distro/versions we use.

manifests/profile/networking/sshd.pp
```diff
@@ -26,26 +26,49 @@
     hasrestart => true,
   }
 
+  # Reset sshd_config to original distro config. May be deleted once this has been applied to existing hosts.
   file { '/etc/ssh/sshd_config':
+    source => "puppet:///modules/nebula/default/${facts['os']['distro']['codename']}/etc/ssh/sshd_config",
+    notify => Service['sshd'],
+  }
+
+  file { '/etc/ssh/sshd_config.d/50-lit.conf':
```
daaang, line 35: Makes sense.

manifests/profile/networking/sshd.pp
```diff
@@ -26,26 +26,49 @@
...
     content => template('nebula/profile/networking/sshd_config.erb'),
     notify  => Service['sshd'],
   }
 
-  concat { '/etc/ssh/ssh_config': }
```
daaang, line 34: Ah I was wondering where this was coming from.

manifests/profile/bolt.pp
```diff
@@ -60,4 +60,6 @@
         | SSH_CONFIG
     }
   }
+
+  concat { '/etc/ssh/ssh_config.d/50-bolt.conf': }
```
daaang, line 64: Any reason for the concat with a single fragment instead of just a file? I'm assuming it was a concat previously in order to span profiles.

manifests/profile/networking/sshd.pp
```diff
@@ -26,26 +26,49 @@
...
+  }
+  # Diverts file without moving. This is required if `sshd-defaults` already exists, but isn't safe unless
+  # we're also writing `sshd-defaults` afterwords.
+  exec { 'divert pam.d/sshd 2':
+    # TODO: remove this exec once all existing hosts have been remediated.
+    unless  => "/usr/bin/test `/usr/bin/dpkg-divert --truename /etc/pam.d/sshd` = '/etc/pam.d/sshd-defaults'",
+    timeout => 30,
+    command => '/usr/bin/dpkg-divert --no-rename --divert /etc/pam.d/sshd-defaults --add /etc/pam.d/sshd',
+    before  => File['/etc/pam.d/sshd-defaults'],
+    require => Exec['divert pam.d/sshd'],
```
daaang, line 59: Ok.

1. If `/etc/pam.d/sshd-defaults` does not exist,
2. We run `dpkg-divert --rename --divert /etc/pam.d/sshd-defaults --add /etc/pam.d/sshd` and then
3. If `dpkg-divert --truename /etc/pam.d/sshd` is not `/etc/pam.d/sshd-defaults`
4. We run `dpkg-divert --no-rename --divert /etc/pam.d/sshd-defaults --add /etc/pam.d/sshd`

I read the man page, but I don't get what a divert is, or why we need to do it, or why the above sequence is important.

daaang (commented):
Overall, I see what this is doing. But I do not understand what the divert does, and I do not understand what you want it to do, so I cannot say whether this will do what you want it to do.

manifests/profile/networking/sshd.pp
daaang: Wait no, this is incomplete, and I'm guessing the above step 3 will prevent step 4 from ever happening, which is the point.

1. If `/etc/pam.d/sshd-defaults` does exist,
2. And if `dpkg-divert --truename /etc/pam.d/sshd` is not `/etc/pam.d/sshd-defaults`
3. We run `dpkg-divert --no-rename --divert /etc/pam.d/sshd-defaults --add /etc/pam.d/sshd`

Either path, the point is that `dpkg-divert --truename /etc/pam.d/sshd` is `/etc/pam.d/sshd-defaults` and `sshd-defaults` exists, everywhere and in an hour, and then we're ready for phase two.

I still don't get what it means to have a truename diverted, but this should do it everywhere, either (I assume?) creating sshd-defaults with `--rename` or running `--no-rename` if ssh-defaults is already there.

daaang: Is this to catch dpkg up to what we're currently doing with puppet, so that we can have puppet let go of the files and undo this with dpkg so that it takes over when we finish?

files/default/bookworm/etc/ssh/ssh_config
rrotter: Good catch. I strongly agree with quashing this. Almost anything here can be overridden in a drop in. I'll add that.

files/default/trixie/etc/ssh/sshd_config
rrotter: Yep. Started each in containers, installed sshd, copied the respective files out.

manifests/profile/bolt.pp
rrotter: Look up to line 50, looks like a loop to me.

manifests/profile/networking/sshd.pp
rrotter: - `dpkg-divert --divert` tells dpkg that when it's managing a file that a package expects to be at `/etc/pam.d/sshd`, that it should be at `/etc/pam.d/sshd-defaults` instead.
- `--rename` is the option you'd normally want to use. This says: rename sshd to sshd-defaults. This is great as long as `/etc/pam.d/sshd-defaults` does not exist, if it does, this command will always fail.
- With the `--no-rename` flag, we make the same changes to the dpkg database, but the file doesn't get moved. Then we have to follow up and ensure the correct data is at `/etc/pam.d/sshd-defaults`.
- `dpkg-divert --truename /etc/pam.d/sshd` prints the path where dpkg expects to find `/etc/pam.d/sshd`. If it prints `/etc/pam.d/sshd-defaults` we've already successfully diverted the file, and we don't need to do it again.

rrotter: The goal is that dpkg should manage `sshd-defaults` in leiu of managing `sshd`, which we'll continue managing with puppet.

Once we remediate existing hosts we'll only need `dpkg-divert` to run on new hosts, where it will create `sshd-defaults` by relocating `sshd`, so puppet won't need to write this file anymore.

We have three starting states we need to handle correctly:
1. new host
2. existing host with puppet managed `sshd` and `sshd-defaults`
3. existing host where this code has already run

In each case the end state should be:
* `pam.d/sshd-defaults` contains the data that Debian normally writes to `pam.d/sshd`
* dpkg has "diverted" `pam.d/sshd` accordingly
* `pam.d/sshd` has the same contents we're already generating, but order is critical, because we can't have this data end up in sshd-defaults instead

files/default/bookworm/etc/ssh/ssh_config
rrotter: Fixed in last push.

daaang (approved):

rrotter:
rebased to production.

PR #880, opened 2025-11-20
k8s gateway sshd: stop blocking puppet bolt

rrotter:
Override default MaxStartups setting. Default randomly blocks some ssh sessions when there are already >= 10 pending connections. We routinely start ~30 connections simultaneously when using Puppet Bolt.


rrotter:
Note that the networking and sshd profiles are not present on these hosts, so this does not overlap at all with #879.

daaang (approved):
The test is a nice addition; underlines the importance of the config.

PR #884, opened 2025-11-22
break up ci spec jobs into multiple tasks

rrotter:
Edit: it was really useful to have all of these changes in the same CI run to see how much performance there is to be gained. But this is too much for one PR. I'll resubmit these changes over a few new PRs.

* Use 9 runners instead of 1 for rspec.
* Strategically sort jobs to limit starved cpu cores.
* Make a few individual spec files faster.
  * This is more important that it sounds. The less variation we have in run time per spec file, the more efficiently all the tests run because we can keep all cores busy longer.

In my testing:
* github actions CI completes in < 3 minutes vs. > 9 minutes
* locally, `bundle exec rake parallel_spec_standalone` is consistantly finishing in under 2:00, vs about 2:30 before (o/c this is very hardware dependent)

Future development: with these tasks down to 1-3 minutes, the 60 seconds it takes to run `bundle exec rake spec_prep` is really significant. Not sure what it would take to cache this data, but it could make difference.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,7 +23,7 @@ jobs:
         - "profiles_2"
         - "roles"
     steps:
-    - uses: actions/checkout@v5
+    - uses: actions/checkout@v6
     - name: Run tests
       run: |
         docker compose --file docker-compose.test.yml build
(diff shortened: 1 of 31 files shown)


rrotter:
Supersceded by #886 #887 #888 #889. That takes of everything here except for the docker and gh actions updates in 60c704f, which I'll take care of in another PR.

PR #886, opened 2025-11-24
rm `is_expected.to compile` from role specs

rrotter:
- Remove `is_expected.to compile` from role specs. It is surprisingly slow on some roles, and in any case we run it twice since it's also in `all_roles_spec`
- Also remove role spec files that have no other content besides this example.


rrotter:
Discussed on Slack. Merging.

PR #891, opened 2025-11-26
add hiera-eyaml gem to puppet master role

rrotter:


rrotter:
appears that this is already installed

PR #893, opened 2025-11-27
ci: use more gh actions runners in ci matrix

rrotter:
- run commands directly in gh actions runner, don't use docker
  - this enables faster action startup, caching, and automatic output parsing
- cache bundle
- cache fixtures (i.e. the result of `rake spec_prep`)
- split parallel_rspec job into parts that can run on separate runners
- serialize output for improved readability
- also remove docker from librarian workflow


antmoth:
beautiful runtimes here :eyes:

rrotter:
See sample run here: https://github.com/mlibrary/nebula/actions/runs/19877325080?pr=893

Note the annotations for `xit`s. Similar annotations are produced on failures, but in red. afaict, this is something that GH Actions gives us for "free", but it has been broken by running ci inside of docker and/or our interleaved output.

Output is no longer interleaved, so it shoud be way more readable if you need to read it, but with annotations there should be far less need to look at the raw log.

rrotter:
I should acknowledge here that the test matrix in ci.yml‎ does make the process of running tests more complex. If there is a bug in my logic this could miss tests or break entirely. Logic is intentionally structured to avoid this, e.g.:
* do this:
  * "run all tests starting w/ a-h" then "run ALL tests EXCEPT those that start w/ a-h"
* not this:
  * "run all tests starting w/ a-h" then "run all tests starting w/ i-z"

Previously it was up to the rake task to ensure all tests are run. This rake task wraps parallel_rspec and hides the features needed to balance threads or split the jobs up for more than one runner.

rrotter:
closing this to split it up

PR #898, opened 2025-12-05
HathiTrust ETT-1166: Add php-smarty3

aelkiss:
We are working towards PHP 8 / Debian 12/13 compatibility; smarty3 is provided by Debian including in Debian 12 and 13. It is also available in Debian 11 and works with PHP 7.4; the current version of the catalog also works with this package installed on dev-N.catalog.hathitrust.org -- that is, Smarty 2 via php-smarty (our package) does not conflict with Debian-packaged php-smarty3.

### DIFF
--- a/manifests/profile/hathitrust/php.pp
+++ b/manifests/profile/hathitrust/php.pp
@@ -22,6 +22,7 @@ class nebula::profile::hathitrust::php () {
       'php-mdb2',
       'php-mdb2-driver-mysql',
       'php-smarty',
+      'smarty3',
       'php-xml',
       'php-mbstring',
       'php-yaml',


aelkiss:
FYI @moseshll

@rrotter I think it should be fine to merge this Monday unless you have anything else you'd want us to test first.

Ultimately we may end up on Smarty 4 or 5, but I think our plan right now is to work on fixing other deprecations & warnings, make sure we can get things working fully under PHP 8 on Debian 12 and 13, and only after that look at updating to newer versions of Smarty -- assuming Smarty 3 in fact works with PHP 8.

rrotter (approved):

manifests/profile/hathitrust/php.pp
```diff
@@ -22,6 +22,7 @@
       'php-mdb2',
       'php-mdb2-driver-mysql',
       'php-smarty',
+      'php-smarty3',
```
rrotter:
```suggestion
      'smarty3',
```

rrotter: I think this is the package you meant.

rrotter (changes requested):

manifests/profile/hathitrust/php.pp
aelkiss: Yes indeedy.

rrotter (approved):
lgtm. Just let me know when you're done with the locally packaged php-smarty so we can clear it out.

PR #900, opened 2025-12-09
use truenas for ht solr6

rrotter:
- remove smartconnect
- change target of all nfs mounts
- remove nfs3 flag on all nfs mounts
- remove redundant mountpoints (mount /htsolr/lss, not sub-directories)
- rewrite `index-release` script

### DIFF
--- a/templates/profile/hathitrust/solr6/index-release.sh.erb
+++ b/templates/profile/hathitrust/solr6/index-release.sh.erb
@@ -1,117 +1,108 @@
-#! /bin/bash
+#!/bin/bash
 
-TODAY=`date +%Y-%m-%d`
-#TODAY="2019-08-31"
+onoe () {
+  echo "${1}"
+  exit 1
+}
 
-BASE=/htsolr/<%= @solr_name %>
-SYMLINKBASE=/htsolr/serve
-SNAP=".snapshot/<%= @snapshot_name %>_${TODAY}"
-CORES="<%= @solr_cores.join(' ') %>"
+base=/htsolr/<%= @solr_name %>
+serve=/htsolr/serve
+manual_snap="$1"
+if [[ -n "${manual_snap}" ]]; then
+  snap="${base}/.zfs/snapshot/${manual_snap}"
+  if ! [[ -d "${snap}" ]]; then
+    echo "can't find snapshot: '${manual_snap}'"
+    echo "valid snapshots are: "
+    \ls ${base}/.zfs/snapshot/
+    exit 1
+  fi
+else
+  snaps=( "${base}/.zfs/snapshot/htsolr-<%= @solr_name %>_"* )
+  snap="${snaps[-1]}"
+  pat="htsolr-<%= @solr_name %>_$(date +%Y-%m-%d)"
+  [[ "$snap" =~ "$pat" ]] || onoe "latest snap was not made today, refusing to run automatic release"
+fi
+snapname="$(basename "${snap}")"
 
 # check if the administrative release stop flag is set
-if [ -f ${BASE}/flags/<%= @solr_stop_flag %> ]; then
-  /bin/echo "<%= @solr_stop_flag %> flag present...skip release today"
-  exit 1
-fi
+[[ -f "${base}/flags/<%= @solr_stop_flag %>" ]] && onoe "<%= @solr_stop_flag %> flag present...skip release today"
 
 # check that today's snap exists
-for s in $CORES; do
-  if [ ! -d $BASE/cores/${s}/${SNAP} ]; then
-    /bin/echo "missing today's snapshot...aborting release"
-    exit 1
-  fi
-done
+[[ ! -d ${snap}/cores/ ]] && onoe "can't read snapshot dir...aborting release"
+[[ ! -d ${snap}/flags/ ]] && onoe "can't read snapshot dir...aborting release"
 
 # check today's snap for 'busy' flag
-if [ -f $BASE/flags/${SNAP}/busy ]; then
-  /bin/echo "indexing and/or optimization ocurring...skip release today"
-  exit 1
-fi
+[[ -f "${snap}/flags/busy" ]] && onoe "indexing or optimization ocurring...skip release today"
 
-<% if @is_primary_site %>
-# # check whether release happened at mirror site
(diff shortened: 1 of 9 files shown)


templates/profile/hathitrust/solr6/index-release.sh.erb
```diff
@@ -1,117 +1,85 @@
...
-    /bin/echo "at least one core served by `hostname -s` has an improper number of segments...aborting release"
-    exit 1
-  fi
-done
+<% end %>
+# check that cores appear valid
+<% @solr_cores.each do |core| %>
+[[ ! -d ${snap}/cores/<%= core %>/data/index ]] && onoe "core data missing...aborting release"
+segments=( "${snap}/cores/<%= core %>/data/index/"*.fdt )
+(( ${#segments[@]} > 0 )) || onoe "core '<%= core %>' served by $(hostname -s) has an improper number of segments...aborting release"
```
aelkiss, line 51: This appears to check that the core has at least 1 segment. That loses the check that full-text has either 1 or 2 segments. While for full text such an index might or might not work, it's probably an indication that the optimization process didn't complete, and we probably wouldn't want to do a release. Normally, the busy flag should be set in such a case, but I could imagine a case where something went wrong, the busy flag isn't set, optimize didn't finish, and the core is in an inconsistent state. That might be unlikely enough (> 2 segments, busy flag not set, index not usable) it isn't worth checking for here, though.

templates/profile/hathitrust/solr6/index-release.sh.erb
```diff
@@ -1,117 +1,85 @@
...
-systemctl stop solr
-if [ $? -ne 0 ]; then
-  /bin/echo "error stopping lss solr on `hostname -s`"
-  exit 1
-fi
+systemctl stop solr || onoe "error stopping solr on $(hostname -s)"
 
 # replace the snapshot symlinks
-<% if @is_lss %>
-rm -f ${SYMLINKBASE}/lss-shared && ln -s ${BASE}/shared/${SNAP} ${SYMLINKBASE}/lss-shared
```
aelkiss, line 57: Is something else handling the lss-shared symlink now or is that otherwise obviated?

templates/profile/hathitrust/solr6/index-release.sh.erb
```diff
@@ -1,117 +1,85 @@
...
+}
 
-BASE=/htsolr/<%= @solr_name %>
-SYMLINKBASE=/htsolr/serve
-SNAP=".snapshot/<%= @snapshot_name %>_${TODAY}"
-CORES="<%= @solr_cores.join(' ') %>"
+base=/htsolr/<%= @solr_name %>
+serve=/htsolr/serve
+snaps=( "${base}/.zfs/snapshot/htsolr-<%= @solr_name %>_"* )
+snap="${snaps[-1]}"
```
aelkiss: Will the glob always sort such that we're guaranteed to get the most recent snap? If so this should be fine, just want to make sure this isn't potentially unpredictable

aelkiss: It's also worth noting that before this would look for a snapshot for today and if it didn't exist it would complain; now we don't have that behavior. That's potentially risky insofar as that if the snapshots stop happening for some reason we 1) won't necessarily notice and 2) the old snapshot might expire before we notice.

aelkiss: We also might want to consider accepting a snapshot name as a parameter to the script such that we can more easily do a manual release or rollback, e.g.

```bash
snap="snap="${1:-${snaps[-1]}}"
```
that is, the value is the first positional parameter, or if that is not set, `${snaps[-1]}`

aelkiss (commented):
The error reporting function `onoe` helps make this more readable; I also like unrolling the loops per solr core with erb rather than in the bash script.

I do have a couple questions about lss-specific things specifically `lss-shared` and the segment count check.

templates/profile/hathitrust/solr6/index-release.sh.erb
rrotter: I was mostly referencing the realized script from catalog when writing this and missed this difference in lss. Fixed in last push.

aelkiss (approved):
It looks like this addresses the concerns I had & the additional checking for the manual specification of snap generally looks OK.

templates/profile/hathitrust/solr6/index-release.sh.erb
```diff
@@ -1,117 +1,108 @@
...
-SYMLINKBASE=/htsolr/serve
-SNAP=".snapshot/<%= @snapshot_name %>_${TODAY}"
-CORES="<%= @solr_cores.join(' ') %>"
+base=/htsolr/<%= @solr_name %>
+serve=/htsolr/serve
+snaps=( "${base}/.zfs/snapshot/htsolr-<%= @solr_name %>_"* )
+manual_snap="$1"
+if [[ -n "${manual_snap}" ]]; then
+  snap="${base}/.zfs/snapshot/${manual_snap}"
+  if ! [[ " ${array[@]} " =~ " ${snap} " ]]; then
```
aelkiss: Could just check if the directory exists rather than the more complicated array thing:

```bash
if ! [[ -d "${snap}" ]]
```

templates/profile/hathitrust/solr6/index-release.sh.erb
```diff
@@ -1,117 +1,108 @@
...
+    echo "can't find snapshot: '${manual_snap}'"
+    echo "valid snapshots are: "
+    \ls ${base}/.zfs/snapshot/
+    exit 1
+  fi
+else
+  snap="${snaps[-1]}"
```
aelkiss, line 21: Could avoid the bashism for the array & ensure sorted order with

templates/profile/hathitrust/solr6/index-release.sh.erb
rrotter: Agreed. That's much better. Done.

templates/profile/hathitrust/solr6/index-release.sh.erb
rrotter: Looks like there was more to this message? Not sure what your suggestion is.

I'm just trying to avoid shelling out to `ls`. Probably several ways to do this.

aelkiss: I was going to suggest something like `$(ls -tr | tail -1)` I think, but if you prefer to keep it all with bash built-ins that's fine.

PR #901, opened 2025-12-10
make ci faster, improve output

rrotter:
- add ci matrix to increase parallelism
- serialize output improve readability of logs
- cache gems and test fixtures
  - improve speed of most ci runs
  - improve reliability, in case rubygems.org or forge.puppet.com are unreachable
  - be good citizens by reducing load on hosting sites
- GitHub actions should be able to automatically parse and annotate output


daaang (approved):
This makes sense to me, and the results speak for themselves.

PR #902, opened 2025-12-10
add ci:spec_profiles, ci:spec_classes to ci matrix

rrotter:
Split up ci:spec into two groups:
- ci:spec_profiles (profile specs)
- ci:spec_classes (all other specs)


rrotter:
Closing, but keeping the branch. At time of writing this branch "gets the CI under 2 minutes instead of ~3 to 3:30". Maybe I'll bring this PR back if we ever creep over 5 min again, but at this point IMO it's not worth it.

PR #903, opened 2025-12-10
remove smartconnect dns server entry

rrotter:
- do nothing to configure smartconnect
- delete `nameserver 127.0.0.1` from resolv.conf
- does NOT remove bind9, we'll purge it manually
- after this has run everywhere the smartconnect class can be removed completely


rrotter:
let's not do this. didn't need intermediate steps for solr. can remediate any junk left behind manually.

PR #906, opened 2025-12-11
uninstall prometheus node exporter

rrotter:
todo:
- [x] remove unused templates
- [x] complete follow-up PR, #923
- [x] test on a server
- [x] immediately before merge, remove prometheus user/group from control repo


daaang (approved):
This looks right. I'm glad to see you aren't holding onto much "just in case" or whatever. If it turns out we need any of this old custom stuff after all for some reason, it's all here in the git history.

PR #908, opened 2025-12-12
update gems

rrotter:
- [x] update pdk template to 3.3.0
- [x] update pdk template to 3.4.0, then to 3.4.0.3
- [ ] update pdk template to 3.5.0
- [ ] move to OpenFact, remove pdk
- [x] get CI to test Debian 13
- [x] get CI to succeed on Debian 13
- [ ] update modules

edit: see https://github.com/puppetlabs/pdk-templates/blob/main/CHANGELOG.md if there is any question why we need to abandon PDK. They are moving support from the dead open source puppet to the shared source puppetcore. We need to move to openvox/openfact gems instead.


rrotter:
starting over. divered too far. it was a good spike.

PR #915, opened 2025-12-15
Support Debian 13 for HT apache servers

rrotter:
~This is what I had to change to get puppet to apply sucessfully on Debian 12.~
Changes to support Debian 13.


aelkiss:
I'll have some more edits based on what I suspect to actually be in use.

aelkiss (approved):
I think this should be good to go on preview but we'll want to wait on rolling this out more widely until probably mid-January so that everybody has sufficient time to test.

manifests/profile/hathitrust/perl.pp
```diff
@@ -139,18 +146,11 @@
     'File::Pairtree',
     'CGI::Application::Plugin::Routes',
     'Algorithm::LUHN',
-    'OAuth::Lite',
```
aelkiss, line 142: These weren't required by our Dockerfile and likely aren't needed any more at all:

* `OAuth::Lite` (@moseshll the CRMS Dockerfile depends on it but I don't see it in use anywhere in the code)
* `Sub::Uplevel`
* `Devel::Cycle`
* `Test::Memory::Cycle`
* `Mozilla::CA`

@moseshll Are you aware of any needs for these modules? (Some could be transitive dependencies of things we still need to install via CPAN?)

aelkiss: Also available via Debian:

* `libip-geolocation-mmdb-perl`
* `libmarc-perl`
* `libmarc-xml-perl` (includes `MARC::File::XML`)

Apparently not used AFAICT:

* `CGI::Application::Plugin::Routes`

moseshll: OAuth::Lite used to be used with Data API in CRMS but that code was removed. The dependency can be removed AFAICT. The rest were probably copypasta from `feed` or some other Perl repo.

manifests/profile/hathitrust/perl.pp
```diff
@@ -70,6 +70,7 @@
     'libio-html-perl',
     'libio-socket-ssl-perl',
     'libio-string-perl',
+    'libip-geolocation-mmdb-perl',
```
aelkiss, line 73: only available on debian 13, not debian 12. Are we expecting this needs to work specifically on debian 12 or are we aiming to go straight to debian 13? Do we need to add any conditionality? Or should we just keep some things installed via CPAN until everything is on debian 13?

manifests/profile/hathitrust/dependencies.pp
```diff
@@ -20,7 +20,7 @@
       'kakadu',
```
aelkiss: we don't have kakadu packaged for debian 13, and I think it's time to cut bait and see if grokj2k will work for production purposes there.

manifests/profile/hathitrust/perl.pp
rrotter: I want to be done w/ Debain 12.

manifests/profile/hathitrust/dependencies.pp
rrotter: I added grokj2k to the PR. So far it was just installed by hand on one test host.

manifests/profile/hathitrust/php.pp
```diff
@@ -39,16 +47,16 @@
...
       'Pager'                 => { package_prefix => '', provider => 'pear' },
-      'PHP_Compat'            => { package_prefix => '', provider => 'pear' },
-      'Structures_LinkedList' => { ensure => 'beta', package_prefix => '', provider => 'pear' },
+
+      'DB'                    => { ensure => 'absent', package_prefix => '', provider => 'pear' },
+      'DB_DataObject'         => { ensure => 'absent', package_prefix => '', provider => 'pear' },
+      'HTTP_Session2'         => { ensure => 'absent', package_prefix => '', provider => 'pear' },
+      'Log'                   => { ensure => 'absent', package_prefix => '', provider => 'pear' },
+      'PHP_Compat'            => { ensure => 'absent', package_prefix => '', provider => 'pear' },
+      'Structures_LinkedList' => { ensure => 'absent', package_prefix => '', provider => 'pear' },
```
rrotter: These are only installed if we installed them, right (i.e. not part of Debian)? Okay if we remove the `ensure => 'absent'` lines, since we'll be imaging new hosts w/ Debian 13?
```suggestion
```

aelkiss: That should be correct.

rrotter:
Rebased and squashed.

PR #916, opened 2025-12-17
Ensure the clamav-freshclam service is running on fulcrum servers

daaang:


umjnega (approved):
Thanks for grabbing this from the shared inbox.  Looks good to me.

PR #919, opened 2025-12-18
Add exechosts -P for running a plan instead of a command

daaang:
Bolt's inventory.yaml is perfectly fine for establishing a hierarchy of hosts and configuring how to connect to them. But it really just does not have an easy way to say "all the VMs not at ICTC" or "nonproduction mysql physical servers."

And that's fine because someone already built taghosts, which works great.

### DIFF
--- a/templates/profile/taghosts/exechosts.sh.erb
+++ b/templates/profile/taghosts/exechosts.sh.erb
@@ -1,10 +1,11 @@
 #!/bin/sh
 set -e
 PROG="$0"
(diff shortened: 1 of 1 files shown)


rrotter (commented):
Looks good! I like the reuse of the `-j` option. This workflow is what I envisioned when requesting we add that flag, but I think I actually used it once or twice to test it and never again. This looks vastly more usable and should be a big improvement in making regular use of bolt plans.

PR #921, opened 2025-12-22
update Gemfile, Rakefile; use voxpupuli gems

aayuraje:
moving from puppet to voxpupuli gems


spec/defines/haproxy_service_spec.rb
```diff
@@ -19,7 +19,8 @@
         os_facts.merge(
           datacenter: "dc1",
           networking: {
-            ip: "40.41.42.43"
+            ip: "40.41.42.43",
+            fqdn: "useless.mock"
```
daaang: Looks like `facts.networking` needs to always have both `ip` and `fqdn` now if we're overriding the default, even when we only care about one of them?

spec/classes/profile/known_host_public_keys_spec.rb
```diff
@@ -8,13 +8,15 @@
 describe "nebula::profile::known_host_public_keys" do
   on_supported_os(supported_os: Nebula.supported_os).each do |os, os_facts|
     context "on #{os}" do
+      let(:my_ip) { Faker::Internet.ip_v4_address }
```
daaang: Is `:my_ip` a pattern from something? Just in this PR, I'm only seeing it being assigned to `networking.ip`, so I'm left wondering what it's for. Like, if I saw us checking against `my_ip` later in the tests, that would make sense, but if not, why not just set `networking.ip` to something invalid like `0.1.2.3`?

Not requesting a change or anything, just wondering if I'm missing anything.

daaang (approved):
Cool, unless I'm misreading this, it looks like mostly name changes (necessary and expected) and a new testing requirement that we either leave `networking` alone or that we must set both `ip` and `fqdn` to something.

spec/classes/profile/known_host_public_keys_spec.rb
rrotter: Copy-pasted from another spec file. I like your suggestion. That mirrors what I was trying to do with the "useless.mock" domain: leave a clue that the mock is only there to silence a warning, and we aren't really using it in the test.

spec/defines/haproxy_service_spec.rb
rrotter: Yes. rspec-puppet specifically looks for these two facts and prints warnings if they're not present.

https://github.com/puppetlabs/rspec-puppet/blob/main/lib/rspec-puppet/support.rb?#L298-L304

PR #922, opened 2025-12-22
update compose file

rrotter:
- replace rubocop w/ standard
- rm deprecated `version` field
- rename to compose


rrotter:
We'll probably fold this into the gem updates

PR #924, opened 2026-01-05
ht solr release script: umount stale snapshots

rrotter:
umount stale snapshots, misc lint fixes

Currently, when we delete old snaps they may still be mounted solr servers. This `umount`s all snaps except the one we're releasing so these excess mountpoints don't stick around.

### DIFF
--- a/manifests/profile/hathitrust/solr6/catalog.pp
+++ b/manifests/profile/hathitrust/solr6/catalog.pp
@@ -40,6 +40,7 @@ class nebula::profile::hathitrust::solr6::catalog (
   $core_link_prefix = ''
   $is_catalog = true
   $is_primary_node = true # catalog solr is only one node per site
+  ensure_packages(['jq'])
   file { '/usr/local/bin/index-release':
     owner   => 'root',
     mode    => '0755',
--- a/manifests/profile/hathitrust/solr6/lss.pp
+++ b/manifests/profile/hathitrust/solr6/lss.pp
@@ -48,6 +48,7 @@ class nebula::profile::hathitrust::solr6::lss (
   $solr_stop_flag = 'STOPLSSRELEASE'
   $core_link_prefix = 'lss-'
   $is_lss = true
+  ensure_packages(['jq'])
   file { '/usr/local/bin/index-release':
     owner   => 'root',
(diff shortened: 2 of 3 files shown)


aelkiss (commented):
The general idea makes sense to me. It should be safe enough to unmount these since 1) solr is stopped at the point of unmounting and 2) anything we still need should get automatically mounted.

I tried the jq expression and it looked good, but I think it could be written more simply as:

```
.filesystems[].children[].target | select(. != "'"${snap}"'")
```

I'm not an expert at jq by any means, though, so if there are implications to this version I think it's fine to leave it as-is.

rrotter:
> The general idea makes sense to me. It should be safe enough to unmount these since 1) solr is stopped at the point of unmounting and 2) anything we still need should get automatically mounted.
> 
> I tried the jq expression and it looked good, but I think it could be written more simply as:
> 
> ```
> .filesystems[].children[].target | select(. != "'"${snap}"'")
> ```
> 
> I'm not an expert at jq by any means, though, so if there are implications to this version I think it's fine to leave it as-is.

`.filesystems[].children` may be null which causes this to error w/o the extra `select()`.

manifests/profile/hathitrust/solr6/catalog.pp
```diff
@@ -40,6 +40,7 @@
   $core_link_prefix = ''
   $is_catalog = true
   $is_primary_node = true # catalog solr is only one node per site
+  ensure_packages(['jq'])
```
daaang, line 43: I'm pretty sure this is installed on all linuxes for the pushgateway script, but ensuring it's installed is harmless, and it's always good to make dependencies clear.

daaang (approved):
Cool. I hope this is enough to fix whatever isn't connecting with nfs and zfs snapshots, but even if it isn't, cleaning up unused mounts is an end in itself, plus it should make it safer to prune old snapshots.

manifests/profile/hathitrust/solr6/catalog.pp
rrotter: Yep, I copy-pasted the ensure from the prometheus profile.

PR #926, opened 2026-01-09
remove skynet cookie and usertrack log

aayuraje:
ticket for this work: https://mlit.atlassian.net/browse/AESYS-327

### DIFF
--- a/manifests/apache/www_lib_vhost.pp
+++ b/manifests/apache/www_lib_vhost.pp
@@ -9,7 +9,6 @@ define nebula::apache::www_lib_vhost (
   Boolean $ssl = false,
   Optional[Integer] $port_override = undef,
   Boolean $auth_openidc = false,
-  Boolean $usertrack = false,
   Optional[String] $auth_openidc_redirect_uri = undef,
   String $ssl_cn = $servername,
   Array[Hash] $directories = [],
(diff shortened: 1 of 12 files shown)


templates/profile/kubernetes/kubeadm_config.yaml.erb
```diff
@@ -31,3 +31,4 @@ clusterName: kubernetes
 apiVersion: 'kubelet.config.k8s.io/v1beta1'
 kind: 'KubeletConfiguration'
 cgroupDriver: 'systemd'
+FailCgroupV1: false
```
botimer, line 34: This looks like an extraneous addition to this PR.

manifests/apache/www_lib_vhost.pp
```diff
@@ -9,7 +9,7 @@
   Boolean $ssl = false,
   Optional[Integer] $port_override = undef,
   Boolean $auth_openidc = false,
-  Boolean $usertrack = false,
+  Boolean $usertrack = false,  # TODO: remove usertrack after making sure it's not needed anymore
```
botimer: We have confirmed that we do not need this module for now. Safe to remove.

botimer (changes requested):
We can do the final few removals as well, but the shape of and details of this change look good to me.

templates/profile/kubernetes/kubeadm_config.yaml.erb
aayuraje: You're correct, I didn't notice that I was on the wrong branch. I'll get rid of this before merging.

manifests/apache/www_lib_vhost.pp
aayuraje: Great, I'll remove mod_usertrack.

botimer (approved):
I think this is ready to go. 👍 🚢

PR #929, opened 2026-01-19
don't ensure falcon running

rrotter:
since falcon fails to restart on upgrade, ignore service start errors


umjnega (approved):
Enabled is good enough.   Good to go.

PR #931, opened 2026-01-22
Kubeadm config

antmoth:
create a drop-in sysctl config file for kubelet setting the cgroups fail = false as a command line argument. this will create the file on etcd nodes even though it seems to be unnecessary there. since this change only needs to exist until all our nodes are on trixie, it seems ok?


rrotter:
No longer needed on Debian 13.

PR #933, opened 2026-01-26
HathiTrust ETT-634: PHP composer

aelkiss:
Add composer in development environments for vendoring PHP libraries

Not needed in production - we'll essentially do composer install (or whatever) in dev, then scp those files to production.


aelkiss:
Note that eventually, we should be able to do away with some of the Debian supplied PHP libraries in favor of using vendored versions installed via composer at deploy time.

aelkiss:
FYI @moseshll

rrotter (approved):

PR #936, opened 2026-01-28
remove libaprutil1-dbd-oracle from quod only

umjnega:
Removing libaprutil1-dbd-oracle to fix tang upgrade issue as well as confirm that puppet doesn't remove it from production quod machines.


umjnega:
Confirmed... this did not remove the package on mocha-2.   And also fixed the dependency on tang.   should be good to go.

skorner (approved):
Approved. We need this as a workaround until we can build a new deb or are done with legacy auth.

PR #940, opened 2026-02-02
correct spelling

rrotter:
correct spelling w/ [crate-ci/typos](https://github.com/crate-ci/typos)

### DIFF
--- a/templates/profile/prometheus/rules.yml.erb
+++ b/templates/profile/prometheus/rules.yml.erb
@@ -377,7 +377,7 @@ groups:
       summary: 'MySQL replication is not receiving data on {{$labels.hostname}}'
       description: 'MySQL replication is not receiving data'
       documentation: 'https://mlit.atlassian.net/wiki/spaces/AE/pages/9339612/MySQL'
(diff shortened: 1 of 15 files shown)


templates/profile/prometheus/rules.yml.erb
```diff
@@ -377,7 +377,7 @@ groups:
       summary: 'MySQL replication is not receiving data on {{$labels.hostname}}'
       description: 'MySQL replication is not receiving data'
       documentation: 'https://mlit.atlassian.net/wiki/spaces/AE/pages/9339612/MySQL'
-  - alert: MysqlMaxPreparedStatment
+  - alert: MysqlMaxPreparedStatement
```
rrotter, line 380: @daaang, can you comment as to whether this is an intentional misspelling to match an API, or if this string would need to be changed elsewhere (outside of this repo) to match this change?

daaang: Can confirm, the alert name only matters as a string that we read as humans. Changing an alert name won't have any surprising side effects.

PR #941, opened 2026-02-10
update puppetlabs/apt and other modules

rrotter:
- remove lock on apt module, because puppetlabs/docker update was finally released
- update all modules to current
- add documentation on using `rake forge:update` task
- bump nebula version to 3.1.0

### DIFF
--- a/.fixtures.yml
+++ b/.fixtures.yml
@@ -12,7 +12,7 @@ fixtures:
       ref: 5.0.0
     letsencrypt:
       repo: puppet/letsencrypt
-      ref: 13.1.0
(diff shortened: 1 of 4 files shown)


.fixtures.yml
```diff
@@ -45,16 +45,16 @@ fixtures:
       ref: 2.0.2
     docker:
       repo: puppetlabs/docker
-      ref: 10.3.0
+      ref: 10.4.0
```
daaang, line 48: 🎊

daaang (approved):

PR #946, opened 2026-02-19
update nodejs

rrotter:
- always use profile::apt::nodejs, so we don't have two places we're managing this source
- only use nodesource source if it's a newer version than os supplies
- remove support for deprecated nodesource node_14 repo
- use current keyring everywhere
- set default nodejs version to 22


umjnega (approved):
Looks good to me.

PR #953, opened 2026-02-27
taghosts tab completion

rrotter:


templates/profile/taghosts/taghosts.sh.erb
```diff
@@ -154,6 +168,26 @@ get_all_tags() {
...
+      lines="`echo "$lines" | happy_grep "\\<$tag\\>"`"
+    fi
+  done
+  lines=$(echo "$lines" | sort -u)
+  tags=$(echo ${lines[@]} | sed -e 's/ /\n/g' | sort -u)
+  for tag in ${tags[@]}; do
+    if echo "${lines[@]}" | grep -qv -- $tag; then
+      echo $tag
+    fi
+  done
```
rrotter: If completion feels laggy, you're probably waiting for this loop. The zsh completion is a bit faster because it's native zsh rather than using grep, but the logic is the same. I don't have any great ideas on making this faster w/o rewriting in a compiled language, which isn't worth the trouble.

rrotter: Rewrote alogorithm using only a pipeline, no slow bash/zsh looping. 3-5x speedup.

PR #954, opened 2026-03-03
Remove the now unused chipmunk and app_host::standalone roles

skorner:
We are no longer hosting a chipmunk instance and do not plan to again. This was also the last use of the app_host::standalone role, which we are also unlikely to use in the same form again.


botimer (approved):
Let us celebrate this retirement. Our chipmunk has settled in greener pastures.

PR #956, opened 2026-03-11
update gpg key to use hpe2 key, per text comment

umjnega:


umjnega:
Apt was failing on a server and I do not think the HPE repo is needed anymore now that its a VM, but i saw the comment below in the nebula code that is probably the better change to make.

The comment:

      #`hpe1` is deprecated, but as of this writing is still used to sign HPE's Debian repos.
      # At some point in 2025 this key should stop working, at which point we can move Debian
      # to hpe2 as well. Due to an apparent bug in apt, hpe2 does not work when armored, so
      # it has been included in this repo as a dearmored binary key.
      # See also: https://downloads.linux.hpe.com/SDR/keys.html

skorner:
We only considered a p->v for  that server, so it is still a physical server. But this change still makes sense in general for our remaining HP servers.

umjnega:
Thats right, i forgot it was still physical.   My bad.  Thanks for the feed back.

manifests/profile/apt.pp
```diff
@@ -67,7 +67,7 @@
       # it has been included in this repo as a dearmored binary key.
       # See also: https://downloads.linux.hpe.com/SDR/keys.html
       $hpe_key = $facts['os']['name'] ? {
-        'Debian' => 'hpe1.gpg',
+        'Debian' => 'hpe2.gpg',
         'Ubuntu' => 'hpe2.gpg',
       }
```
rrotter, line 72:
```suggestion
      $hpe_key = $facts['os']['distro']['codename'] ? {
        'bullseye' => 'hpe1.gpg',
        default => 'hpe2.gpg',
      }
```

rrotter (changes requested):
Looked into this comment on the HPE site:
> fingerprint: 26C2B797 ([extended through 2027](https://downloads.linux.hpe.com/SDR/hpePublicKey2048_key1.pub))

But it is indeed the same key we already have, so no action needed on that.

If this worked everywhere I'd remove the condition rather than just setting Debian to hpe2 (and also remove hpe1 from the files dir). But this does not work on bullseye, so for now my suggested change should do it.

rrotter (approved):

rrotter:
lgtm. One good last step before merging would be to re-test this on one Debian 12 host and one Debian 11 host.

PR #960, opened 2026-03-12
rm dead code

rrotter:
- don't install dkms on all hosts
- useless i40e test
- unused hathitrust::prod profile


rrotter:
Review q's:
- @skorner, any concern about the dkms change? It's a dependency of any dkms module anyway, so not clear to me why it was ever there, probably just copy pasta from an old runlist.
- @aelkiss: fyi on the removed profiles

skorner (approved):
No concern, and these deletes all look good to me.

PR #962, opened 2026-03-17
use upstream duo .deb

rrotter:
Debian duo packages are unmaintained. Install and configure Duo maintained `duo-unix` package and purge the Debian provided `libpam-duo`.


skorner (approved):
This all makes sense and looks good to me.

PR #964, opened 2026-03-19
ht solr: expose nfs mount opts via hiera

rrotter:


manifests/profile/hathitrust/solr6.pp
```diff
@@ -36,7 +37,7 @@
     '/htsolr/serve':;
   }
   nebula::nfs_mount { '/htapps':
-    options         => 'auto,hard',
+    options         => $nfs_options,
```
rrotter, line 40:
```suggestion
    options         => 'auto,hard',
```
Revert this. No need for custom opts here.

PR #965, opened 2026-03-23
add sha tags for all github actions

rrotter:


daaang (approved):
This syntax is exactly what I was planning on suggesting if the question came up.

PR #968, opened 2026-03-27
update www_lib for trixie and some timely cleanup

umjnega:


manifests/profile/apache/authz_umichlib.pp
umjnega, line 1: This is probably used somewhere outside of www-lib webservers, but i noticed it was being called twice for the www-lib-testing test server.   Removed temporarily since it was installed manually via CPAN for testing.

manifests/profile/www_lib/vhosts/publishing.pp
```diff
@@ -11,6 +11,7 @@
 # @example
 #   include nebula::profile::www_lib::vhosts::publishing
 class nebula::profile::www_lib::vhosts::publishing (
+  String $ssl_cn = 'maps.publishing.umich.edu',
```
rrotter: Shoud this be "zone" or "domain" or "servername" instead? Can we remove the default (or default to "publishing.invalid", etc)?

Where does this get used? Looks like this variable was meant to be used below in `ssl_cert` and `ssl_key` (and could also be used in `servername`) but then wasn't actually used?

- I'd like to see this variable used, but with either no default or a fake default.
- Alternatively, deleting the variable would be okay. I don't like the hardcoded domain, but it was already present in the apache::vhost config below before this PR.

manifests/profile/www_lib/dependencies.pp
```diff
@@ -22,7 +22,7 @@
       'git',
       'emacs',
       'imagemagick',
-      "${java_source}-${jdk_version}-jre",
+      'default-jre-headless',
```
rrotter, line 18: - removing explicit jre versions where we can get away w/ it: 👍👍
- please delete the extranious `$jdk_version`, `$java_source` assignments above

manifests/profile/www_lib/perl.pp
```diff
@@ -136,6 +135,13 @@
     'WebService::Solr::Tiny']:
   }
 
+  # ONLY WORKS ON TRIXIE - libaio is symlinked as the oracle expects
```
rrotter, line 138: worth adding a branch for the bullseye/bookworm version of libaio: `libaio1`? Or just guarding this and the  `libaio1t64` install with an `if (not bullseye)`?

(not sure if anything else is trixie-only though)
[this is a question/suggestion, not necessarily a request to change this]

manifests/profile/www_lib/vhosts/publishing.pp
daaang: I think `default.invalid` is our go-to for fake defaults. +1 to changing the variable name—it seems like let's encrypt is moving away from even having a CN (if you opt into their newer options, there's just SANs and no CN), so I think `server_name` will better stand the test of time.

umjnega: Yeah, I believe i put this there when i was trying to figure out how to specify the ssl_key and ssl_cert.   It was not here originally and I forgot to clean it up once i put the ssl_cert and ssl_key in place.  Its removed.

manifests/profile/www_lib/dependencies.pp
umjnega: I believe i understood this correctly.    removed.

PR #972, opened 2026-04-01
hpe apt repo: fix key errors

rrotter:
Use BOTH hpe1 and hpe2 for all distros


umjnega (approved):
Tested on two hosts.  Looks good to go.

PR #973, opened 2026-04-01
faster ci

rrotter:
- Switch to arm64 runners: consistantly 15%+ faster (~30s) on exactly the same tasks. We're not doing anything architecture specific, so I don't see any downside to taking the free speed bump.
- ~Split up specs into 2 jobs: profile specs, and everything else. This doubles the number of runners we're using.~ [would save another 20s]
- Commit Gemfile.lock. Saves ~10s on recalculating this file. Arguably makes tests more repeatable.
  - Counterpoint: it's a change, and we've had Gemfile.lock in .gitignore for a long time.
- Overall: run time goes from ~2m30s to ~1m40s.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,7 +9,7 @@ on:
 jobs:
   setup_tests:
     name: _setup
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04-arm
     outputs:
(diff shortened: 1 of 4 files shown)


rrotter:
Oops, didn't actually change the runner for all jobs. Fixed up. A cool 15% faster.

PR #974, opened 2026-04-03
update gems, move to non-bundled standardrb

rrotter:
Latest openvox gems require a rubocop that isn't supported by any version of standard. Attempting to fix this by using unbundled standard.


rrotter:
Giving a 👎 to my own PR. Lint takes nearly as long as rspec, because this breaks gem caching. It's also a regression wrt controlling our supply chain.

Probably the best course of action is asking the voxpupuli folks to loosen up their gemspecs (or even adopt Standard Ruby, but that's a bigger ask).

(ffr: specifically voxpupuli-rubocop and voxpupuli-test require rubocop `~> 1.85.1`, and standardrb requires `~> 1.84.0`)

PR #977, opened 2026-04-09
remove haproxyctl no longer supported in trixie

umjnega:


aelkiss:
FWIW we do use haproxyctl for quite a few things including "set_weights.rb" (https://github.com/mlibrary/nebula/blob/e2387bb96479444f59b67512a4de10a19e23a45f/manifests/profile/haproxy.pp#L81-L86 / https://github.com/mlibrary/ae-utils/blob/master/bins/set_weights.rb#L23; there's also quite a bit of documentation in the LIT wiki that references haproxyctl. I'm not quite sure what the recommended way to control haproxy is now (haproxyctl appears to be abandoned) -- maybe it's to access the control socket with socat as in https://docs.haproxy.org/3.0/management.html? I'm not entirely sure how to do the same things that we do with haproxyctl. Just hoping this is something you're looking at.

rrotter:
replaced by #1005

PR #989, opened 2026-04-21
update yarn keyring

rrotter:
Tested new keyring manually on a server, `apt update` works.

Note: various deprecation warnings at https://classic.yarnpkg.com/en/docs/install#debian-stable, is this something we need to push devs to upgrade to?


umjnega:
Looks good to me.  @rrotter   I forgot this was covered, thank you for the reminder.

PR #996, opened 2026-05-05
Remove nodejs from a couple roles

antmoth:
Remove from app_host prod with search metrics; leave on prod mysql metrics; remove from mgetit log


skorner (approved):
Looks good for the minimum change we need right now.

PR #999, opened 2026-05-08
Configure dnsmasq to ignore the presence of /sbin/resolvconf

daaang:
Installing netplan requires installing systemd-resolved which in turn installs /usr/bin/resolvectl which in turn adds a wrapper that is supposed to be compatible with /sbin/resolvconf, but which we've also never wanted, used, or relied on.

Dnsmasq looks for /sbin/resolvconf and tries to play nice with it, but in this case it's getting tricked into trying to play nice with resolvconf when it is in fact systemd's implementation of resolvconf, which dnsmasq can safely just ignore, and which does not behave the way dnsmasq expects.

So we take over management of /etc/default/dnsmasq in order to uncomment `IGNORE_RESOLVCONF=yes`, which will tell dnsmasq to ignore the presence of a systemd implementation of /sbin/resolvconf and not try to use it to actually do anything.

Dnsmasq will instead just assume /etc/resolv.conf is set up correctly.

### DIFF
--- a/manifests/profile/kubernetes/dns_server.pp
+++ b/manifests/profile/kubernetes/dns_server.pp
@@ -63,6 +63,19 @@ class nebula::profile::kubernetes::dns_server {
   concat { '/etc/ssh/ssh_known_hosts': }
   Concat_fragment <<| tag == "${cluster_name}_known_host_public_keys" |>>
 
+  exec { 'divert /etc/default/dnsmasq':
+    creates => '/etc/default/dnsmasq.dist',
+    command => '/usr/bin/dpkg-divert --rename --divert /etc/default/dnsmasq.dist --add /etc/default/dnsmasq',
+    notify  => Service['dnsmasq'],
+    require => Package['dnsmasq'],
+  }
+
+  file { '/etc/default/dnsmasq':
+    content => "CONFIG_DIR=/etc/dnsmasq.d,.dpkg-dist,.dpkg-old,.dpkg-new\nIGNORE_RESOLVCONF=yes\n",
+    notify  => Service['dnsmasq'],
+    require => Exec['divert /etc/default/dnsmasq'],
+  }
+
   file { '/etc/dnsmasq.d/local_domain':
     content => "local=/${private_domain}/\n",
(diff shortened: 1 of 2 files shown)


manifests/profile/kubernetes/dns_server.pp
```diff
@@ -63,6 +63,19 @@
   concat { '/etc/ssh/ssh_known_hosts': }
   Concat_fragment <<| tag == "${cluster_name}_known_host_public_keys" |>>
 
+  exec { 'divert /etc/default/dnsmasq':
+    creates => '/etc/default/dnsmasq.dist',
+    command => '/usr/bin/dpkg-divert --rename --divert /etc/default/dnsmasq.dist --add /etc/default/dnsmasq',
+    notify  => Service['dnsmasq'],
+    require => Package['dnsmasq'],
```
daaang, line 70: I did notice that other implementations add `timeout => 30` to this, but when I tried running it on a couple machines, it was instant. Am I missing a reason I should want to add that?

rrotter: 30 seconds is far longer than it should ever take. The default is 300, which is absurdly long for nearly every command.
https://help.puppet.com/core/current/Content/PuppetCore/Markdown/exec.htm#exec-attribute-timeout

manifests/profile/kubernetes/dns_server.pp
```diff
@@ -63,6 +63,19 @@
...
+  }
+
+  file { '/etc/default/dnsmasq':
+    content => "CONFIG_DIR=/etc/dnsmasq.d,.dpkg-dist,.dpkg-old,.dpkg-new\nIGNORE_RESOLVCONF=yes\n",
```
rrotter, line 74: I don't understand the CONFIG_DIR setting, but it looks like that's what's in the original file, so fine?

rrotter (approved):

manifests/profile/kubernetes/dns_server.pp
daaang: Yeah that's exactly it. These are the two noncomment nonempty lines.

PR #1001, opened 2026-05-18
add unattended_upgrades to minimum role

rrotter:
- add all origins
- blacklist apache and mariadb

considerations:
- we could add an option to turn this on/off in hiera if we want to roll it out gradually, but I don't want that to become at option to except a host forever; unattended_upgrades should not be optional
- @daaang, could you specifically weigh in on automatically updating containerd (which should be the only thing we're using from the Docker origin)?

### DIFF
--- a/manifests/profile/unattended_upgrades.pp
+++ b/manifests/profile/unattended_upgrades.pp
@@ -1,9 +1,26 @@
 class nebula::profile::unattended_upgrades {
   class { 'unattended_upgrades':
     extra_origins                      => [
-      'origin=Vox Pupuli',
+      'origin=mlibrary',
+      'origin=Docker',            # containerd
+      'site=pkg.duosecurity.com', # duo
+      'origin=HPE',               # hpe
+      'origin=elastic',           # filebeat, metricbeat
+      'site=apt.grafana.com',     # grafana alloy
+      'origin=XamarinBuster',     # mono
+      'origin=nginx',             # nginx
+      'site=deb.nodesource.com',  # nodejs
+      'origin=Artifactory',       # openjdk
+      'origin=Vox Pupuli',        # openvox
+      'origin=deb.sury.org',      # php
+      'origin=yarn',              # yarn
     ],
     only_on_ac_power                   => false,
     skip_updates_on_metered_connection => false,
+    blacklist                          => [
+      'falcon-sensor',
+      'apache2','apache2-bin','apache2-data','apache2-utils',
+      'mariadb-server','mariadb-server-.*',
+    ],
   }
 }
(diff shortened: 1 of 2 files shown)


manifests/role/minimum.pp
```diff
@@ -34,5 +34,6 @@
     include nebula::profile::ntp
   }
 
+  include nebula::profile::unattended_upgrades
```
skorner, line 37: Do we want to remove this line from the 5 roles that explicitly have unattended upgrades currently and will be superseded by this change?

skorner (approved):
This is great from my perspective. I left one question, but feel free to try it out when ready!

manifests/profile/unattended_upgrades.pp
```diff
@@ -1,9 +1,26 @@
...
+      'origin=HPE',               # hpe
+      'origin=elastic',           # filebeat, metricbeat
+      'site=apt.grafana.com',     # grafana alloy
+      'origin=XamarinBuster',     # mono
+      'origin=nginx',             # nginx
+      'site=deb.nodesource.com',  # nodejs
+      'origin=Artifactory',       # openjdk
+      'origin=Vox Pupuli',        # openvox
+      'origin=deb.sury.org',      # php
+      'origin=yarn',              # yarn
```
daaang, line 16: Excellent use of comments, lovely 👍

daaang (approved):
My understanding is that this should perform `apt-get dist-upgrade` for us, for "most" things (excepting anything either missed by `extra_origins` or blocked by `blacklist`), and our own experience of running `apt-get dist-upgrade` by hand should become an exercise in patching specifically those exceptions.

In the near future, I think we should consider adding an `apt-get --simulate dist-upgrade` cronjob that can dump a metric like `node_package_upgrades_available $count` in `/var/lib/prometheus`, so that we can check remaining attended upgrades via alerts and/or dashboards, maybe even with some basic categorization e.g. `{group="apache|mariadb|other"}`.

(Long as we're creating custom metrics, maybe we can also add something that detects staged kernels, since that was apparently an abuse of `/run/reboot-required`.)

daaang:
Oh—you wanted me to weigh in on unattended containerd upgrades. True things:

1. We have not tested this.
2. Containerd was never pinned, so most nodes have whatever version a human last gave them.
3. Ergo the test cluster (aka usually the best place to test things) has the least-testable hosts.
4. I've also never seen anyone on the internet complain about disruptive containerd software updates.

(1) speaks for itself, and (2) and (3) are "but that would be hard" counterpoints. (4) provides the "not only hard but almost certainly overcautious" point that sides me in favor of testing in production,

But also yeah we're technically talking about testing in production.

+1

PR #1007, opened 2026-06-01
ht rsync: explicitly allow out-of-module symlinks

rrotter:
Previously, we relied on the (apparently?) unintended behavior of `use chroot = false` to allow symlinks to reach out of the rsync module. This "feature" was recently "fixed". With chroot enabled, we can instead make it explicit that rsync should be able to see the entire filesystem by prepending `/./` to the module path. (see: [rsyncd.conf(5)](https://manpages.debian.org/bullseye/rsync/rsyncd.conf.5.en.html#use))


aelkiss (approved):
Seems worth trying to me.

aelkiss:
FFR the maintainer indicated this was a regression and that it has been fixed upstream: https://github.com/RsyncProject/rsync/issues/944

Not holding my breath that the fix will make it into Debian any time soon, though.

PR #1008, opened 2026-06-03
ETT-1533 plack-lib uses ancient Net::CIDR::Lite with multiple CVEs

moseshll:
- The HathiTrust `babel` repo contains two out of date vendored Perl modules. Neither appear to be available in the default config so we're requesting additional Perl support.
  - `Net::CIDR::Lite` has a number of CVEs that have been fixed recently
    - The Debian trixie version is 0.22 lagging the latest 0.24 which addresses additional vulnerabilities. Better install from CPAN.
  - `Plack::Builder::Conditionals` is simply out of date. No CVEs.
    - Debian provides `libplack-builder-conditionals-perl`

Note to reviewer: I'm in somewhat unfamiliar territory here. This is meant to apply to all of the hosts supporting HathiTrust's discovery apps (in particular imgsrv).

### DIFF
--- a/manifests/profile/hathitrust/perl.pp
+++ b/manifests/profile/hathitrust/perl.pp
@@ -100,6 +100,7 @@ class nebula::profile::hathitrust::perl () {
     'libpackage-stash-perl',
     'libparse-recdescent-perl',
     'libplack-perl',
+    'libplack-builder-conditionals-perl',
     'libpod-simple-perl',
     'libproc-processtable-perl',
     'libprometheus-tiny-shared-perl',
@@ -145,6 +146,7 @@ class nebula::profile::hathitrust::perl () {
     'Algorithm::LUHN',
     'EBook::EPUB',
     'MARC::File::MiJ',
+    'Net::CIDR::Lite',
     'Noid'
     ]:
   }
--- a/manifests/profile/hathitrust/perl.pp
+++ b/manifests/profile/hathitrust/perl.pp
@@ -91,6 +91,7 @@ class nebula::profile::hathitrust::perl () {
     'libmoose-perl',
     'libmouse-perl',
     'libmro-compat-perl',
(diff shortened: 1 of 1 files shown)


manifests/profile/hathitrust/perl.pp
```diff
@@ -145,6 +146,7 @@
     'Algorithm::LUHN',
     'EBook::EPUB',
     'MARC::File::MiJ',
+    'Net::CIDR::Lite',
```
rrotter: Let's use the [libnet-cidr-lite-perl](https://packages.debian.org/en/trixie/libnet-cidr-lite-perl) Debian package instead. Not as outdated as it looks. It says 0.22, but look at the change log you'll see they backported the security content of 0.23 about a month ago (although not 0.24 yet). .deb version numbers suffixed with '-' and a number greater than 1 is the give away that Debian has repackaged or patched something.

The problem with CPAN modules is that while we may get a more up to date version for now, we won't get future updates from Debian automatically, adding to maintenance cost.

rrotter (changes requested):

rrotter:
I'll add that I read the CVEs and IMO they are very marginal. Sounds like they are very dependent on how one actually uses the module. Particularly one would have to be trusting user supplied data and then trusting this module to sanitize the data for you in order for the bug to have any security implications.

manifests/profile/hathitrust/perl.pp
moseshll: Oh, that explains the numbering scheme. I wondered in passing what that was all about.

moseshll:
Frankly I'm more irritated that we're clogging up our repos with vendored cruft, than worried about CVE edge cases. Like the `File::Slurp` copy we had in imgsrv up until AE removed it. Anyway, updated to cut CPAN out of the picture.

PR #1009, opened 2026-06-04
keepalived conf updates

rrotter:
- rm unused smtp config
- replace track_script with track_process: more efficient and reliable
- rm enable_script_security directive: noop since I removed the vrrp_script block
- reduce config to a single .erb template
- add manual weight offset file (at `/etc/keepalived/weight`) to allow overrides for maintenance


umjnega (approved):
This all makes sense to me.  I agree to the change.

PR #1011, opened 2026-06-08
Manage quod apache with puppet

daaang:


rrotter (changes requested):
The rdist this is replacing runs `systemctl reload apache2`. Something like `notify => Service['apache2']` would do the job here I think.

Other than that, lgtm.

skorner (approved):
This is a fantastic solution for the immediate need at hand.

PR #1013, opened 2026-06-09
bin/standardrb: run bundle install automatically

rrotter:
So we don't have to run `bundle install --gemfile=standard.gemfile`, do it automatically in `bin/standardrb`.

Mutually exclusive with PR #1014


rrotter:
replaced by the far surperior #1033

PR #1014, opened 2026-06-09
add bin/bundle-install

rrotter:
So we don't have to run `bundle install --gemfile=standard.gemfile`, add `bin/bundle-install` script to do it.

Mutually exclusive with PR #1013


rrotter:
nm, this is just clunky.

PR #1016, opened 2026-06-18
Require authentication for pageturner/imgsrv

aelkiss:
FYI @rrotter. Staging this here to run CI and just in case we need it. **DO NOT MERGE** without confirming with HT leadership.


rrotter:
Modified this so we can toggle this config change from the control repo. That gives us the option of merging this branch and making a one-line change in hieradata if we ever want to turn this on. I'm also happy to remove my commit if you prefer, just presenting this as an option.

PR #1021, opened 2026-07-01
percona_toolkit profile

rrotter:
optionally install percona-toolkit from profile::mariadb::server


rrotter:
Doesn't appear to be a need for this right now. I'll keep the branch in case it comes up again.

PR #1027, opened 2026-07-20
role::minimum: add rsync package

rrotter:
After needing to manually install `rsync` on each of the last several servers I've imaged, I'm feeling this should be part of the standard base config.

If for some reason we really don't want it everywhere I will add it to the mariadb::server profile instead.

### DIFF
--- a/manifests/role/minimum.pp
+++ b/manifests/role/minimum.pp
@@ -27,6 +27,7 @@ class nebula::role::minimum (
       package { 'iptables-persistent': ensure => purged }
     }
 
+    stdlib::ensure_packages(['rsync'])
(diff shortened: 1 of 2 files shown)


manifests/role/minimum.pp
```diff
@@ -27,6 +27,7 @@
       package { 'iptables-persistent': ensure => purged }
     }
 
+    stdlib::ensure_packages(['rsync'])
```
rrotter, line 30: I can see an argument that this line should be wrapped in a profile.
I can also see an argument that "role::minimum" effectively _is a profile._

rrotter:
Actually, this belongs in profile::base not role::minimum. I will redo it some other time.

PR #1, opened 2018-04-02
Debian umich (AEIM-925)

daaang:
Pending successful completion of AEIM-1054.

PR #1034, opened 2026-08-03
profile::base: rm dbus

rrotter:
Not sure why we ever cared whether `dbus` was installed, but it's part of the default Debian install, so whatever the original reason, this line effectively a no-op.

Note: `dbus` package was added in aaaacd29, part of https://github.com/mlibrary/nebula/pull/1.

### DIFF
--- a/manifests/profile/base.pp
+++ b/manifests/profile/base.pp
@@ -30,7 +30,6 @@ class nebula::profile::base (
   }
 
   ensure_packages([
-    'dbus', # ??
     'zstd', # prevent warnings about fallback on `apt dist-upgrade`
   ])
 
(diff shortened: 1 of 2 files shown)


skorner (approved):
Given dbus is part of the default install, it seems quite helpful to remove this confusing line!

daaang (approved):
My best guess from what I remember doing at the time is that I was probably copying something I found in an abandoned ansible playbook. Certainly, if you find yourself wondering why something was added in a commit that starts with `aaaa`, there’s a very high chance the answer is “unthinking copy/paste from something Matt found on mimosa during his first few months here.”

PR #1036, opened 2026-08-03
Rethink etcd db alerts

daaang:
The previous alerts' 30 minute window was, in this case, way too long to be effective. It takes about 30 minutes in practice to go from "everything is perfectly fine" to "it's full and everything stopped responding to all requests."

Given the short window, it also seems silly to bother with deltas. It should never ever be more than 25% full let alone 50% full, so that should be good enough.

### DIFF
--- a/templates/profile/prometheus/rules.yml.erb
+++ b/templates/profile/prometheus/rules.yml.erb
@@ -395,24 +395,16 @@ groups:
     annotations:
       summary: 'MySQL Connection Count is approaching MAX on {{$labels.hostname}}'
       description: 'MySQL Connection Count is approaching MAX'
-  - alert: "EtcdDatabaseFillingUpFast"
+  - alert: "EtcdDatabaseTooBig"
     expr: >
-      max by (datacenter) (delta(etcd_debugging_mvcc_db_total_size_in_bytes[5m])) > 10000
-    for: "30m"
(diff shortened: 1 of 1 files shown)


templates/profile/prometheus/rules.yml.erb
```diff
@@ -395,24 +395,16 @@ groups:
     annotations:
       summary: 'MySQL Connection Count is approaching MAX on {{$labels.hostname}}'
       description: 'MySQL Connection Count is approaching MAX'
-  - alert: "EtcdDatabaseFillingUpFast"
+  - alert: "EtcdDatabaseTooBig"
     expr: >
-      max by (datacenter) (delta(etcd_debugging_mvcc_db_total_size_in_bytes[5m])) > 10000
-    for: "30m"
+      max by (datacenter) (etcd_mvcc_db_total_size_in_bytes) > 2 ^ 30
```
botimer: I like that the summary includes the more familiar value of 1Gi, though I don't find the 2 ^ 30 form particularly scannable. Should we consider `1 * 1024 * 1024 * 1024` or `1 * 1024^3`?

botimer (approved):
Agreed that, while we prefer the delta style in general, etcd db size is a bit different than general disk trend in terms of expectation and intervention.

templates/profile/prometheus/rules.yml.erb
daaang: Sure

